PR #2, opened 2019-11-11
Repo changes

ipanova:


CHANGES/5625.removal
```diff
@@ -0,0 +1 @@
+Sync is no longer available at the {remote_href}/sync/ repository={repo_href} endpoint.
```
dralley: I updated the changelog entry in the other PRs to be this:

```Sync is no longer available at the {remote_href}/sync/ repository={repo_href} endpoint. Instead, use POST {repo_href}/sync/ remote={remote_href}.

Creating / listing / editing / deleting Docker repositories is now performed on /pulp/api/v3/docker/docker/ instead of /pulp/api/v3/repositories/. Only Docker content can be present in a Docker repository, and only a Docker repository can hold Docker content.```

dralley:
@ipanova Merge at will, everything else has been merged.

CHANGES/5625.removal
ipanova: @dralley  thank you, i updated the entry

PR #11, opened 2019-11-20
As a user, I can remove all container content with ["*"]

ipanova:
closes #5756
https://pulp.plan.io/issues/5756


pulp_container/app/viewsets.py
```diff
@@ -271,6 +271,10 @@ def remove(self, request, pk):
 
         if 'content_units' in request.data:
             for url in request.data['content_units']:
+                if url == '*':
```
dkliban: It would be better if we required '*' to be the only URL and not just one of the URLs. What do you think?

ipanova: can we probably do this, on the other hand if we see [ 'href, 'href', '*', 'href'] i guess we can still treat this as 'remove all'
shall we raise a validation error in case '*' is not the only provided data in the list?

dkliban: yes, let's raise a validation error if '*' is not the only item in the list.

PR #24, opened 2019-12-03
Do not prepend a scheme to a token server in the Www-Authenticate header

lubosmj:
In this commit, there was also updated the documentation. Before this commit, there was not mentioned that pulp_container expects a FQDN of a token server together with a port number. The additional setting CONTENT_HOST was created because the token server should expect only service names wihthout a scheme.

closes #5779
https://pulp.plan.io/issues/5779

### DIFF
--- a/pulp_container/app/authorization.py
+++ b/pulp_container/app/authorization.py
@@ -1,3 +1,4 @@
+import re
 import base64
 import hashlib
 import random
@@ -16,7 +17,7 @@ from pulpcore.plugin.content import Handler
 
 TOKEN_EXPIRATION_TIME = 300
 
-KNOWN_SERVICES = [settings.CONTENT_ORIGIN]
+KNOWN_SERVICES = [re.sub(r'(http://|https://)', '', settings.CONTENT_ORIGIN, count=1)]
 ANONYMOUS_USER = ''
 EMPTY_ACCESS_SCOPE = '::'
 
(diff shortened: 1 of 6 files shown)


pulp_container/app/token_verification.py
```diff
@@ -95,10 +97,10 @@ def _build_authenticate_string(self, source_path):
         Build a formatted authenticate string.
 
         For example, A created string is the following format:
-        realm="https://token",service="docker.io",scope="repository:my-app:push".
+        realm="https://token:123",service="docker.io",scope="repository:my-app:push".
```
ipanova: is 123 a port? port usually comes after the hostname

pulp_container/app/authorization.py
```diff
@@ -16,7 +16,7 @@
 
 TOKEN_EXPIRATION_TIME = 300
 
-KNOWN_SERVICES = [settings.CONTENT_ORIGIN]
+KNOWN_SERVICES = [settings.CONTENT_ORIGIN.lstrip('http://')]
```
ipanova: we cannot assume http protocol was used, we need also to include https

pulp_container/app/token_verification.py
ipanova: what about instead of https://token:123 you write https://<hostname>:<port>/token,

PR #47, opened 2020-01-16
Prevent the schema conversion for manifests with foreign layers

lubosmj:
closes #5646
https://pulp.plan.io/issues/5646

### DIFF
--- a/pulp_container/app/schema_convert.py
+++ b/pulp_container/app/schema_convert.py
@@ -20,7 +20,7 @@ log = logging.getLogger(__name__)
 FS_Layer = namedtuple("FS_Layer", "layer_id uncompressed_digest history")
 
 
-class Schema1ConverterWrapper:
+class Schema2toSchema1ConverterWrapper:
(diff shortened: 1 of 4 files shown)


pulp_container/app/schema_convert.py
```diff
@@ -89,6 +92,13 @@ def __init__(self, manifest, config_layer, name, tag):
         self.fs_layers = []
         self.history = []
 
+        self._assert_foreign_layers()
+
+    def _assert_foreign_layers(self):
+        for layer in self.manifest['layers']:
+            if layer.get('mediaType') in (MEDIA_TYPE.FOREIGN_BLOB, MEDIA_TYPE.FOREIGN_BLOB_OCI):
```
ipanova: in theory we do not need to worry about MEDIA_TYPE.FOREIGN_BLOB_OCI because manifests with oci type will raise error here https://github.com/pulp/pulp_container/blob/master/pulp_container/app/registry.py#L168 and we will not arrive till here

PR #63, opened 2020-02-14
Convert functional tests to use bindings

lubosmj:
In addition, the test test_convert.py was moved to the directory tests/unit/.

closes #6069
https://pulp.plan.io/issues/6069


lubosmj:
The file test_crud_content_unit.py contains only exemplary changes since the tests are not executed due to non-implemented support for push or uploads.

PR #64, opened 2020-03-05
Removes sudo from the buildah command

dkliban:
This patch also removes the use of a volume mount. There seems to be a problem with using a
volume mount when running buildah inside a container.

fixes: #5895
https://pulp.plan.io/issues/5895

### DIFF
--- a/CHANGES/5895.bugfix
+++ b/CHANGES/5895.bugfix
@@ -1,0 +1,1 @@
+Building an image from a Containerfile no longer requires root access.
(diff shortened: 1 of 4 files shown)


ipanova:
@dkliban i had an impression that volume mount was needed because it is used by automation hub?

dkliban:
@ipanova It is not strictly required. When the build container starts up, it automatically makes the context directory available for operations such as COPY. That reminds me that I should update the example Containerfile with this PR.

PR #65, opened 2020-03-05
Adds webserver_snippets for nginx and apache

dkliban:
Adds webserver_snippets for nginx and apache

fixes: #6292
https://pulp.plan.io/issues/6292


dkliban:
@ipanova I tested this change with the latest ansible-pulp release. I was able to perform 'podman pull' on port 80. This only tested the nginx config, but I am confident that the apache one work also.

dkliban:
@bmbouter, can you please review?

PR #67, opened 2020-03-05
1.2.0 changelog version bumps

dkliban:
There are 3 commits here:

    Changelog for 1.2.0
    version bump to 1.2.0
    version bump to 1.3.0.dev

The plan is sto release the second commit as 1.2.0.


dkliban:
I closed the previous PR because I used a + in the name of the branch. That caused a problem with building of the container. This branch name uses only dashes.

ipanova:
@dkliban https://github.com/pulp/pulp_container/blob/master/unittest_requirements.txt we need to remove the upper requirement for pulpcore here to make travis pass

PR #69, opened 2020-03-30
Removing stale tests

fabricio-aguiar:
updating tests due to pulpcore change: https://github.com/pulp/pulpcore/pull/594
[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_recursive_add.py
+++ b/pulp_container/tests/functional/api/test_recursive_add.py
@@ -74,20 +74,6 @@ class TestManifestCopy(unittest.TestCase):
             self.repositories_api.copy_manifests(self.to_repo.pulp_href, {})
         self.assertEqual(context.exception.status, 400)
 
-    def test_empty_source_repository(self):
-        """Ensure exception is raised when source_repository does not have latest version."""
-        delete_response = self.repositories_api.delete(self.to_repo.latest_version_href)
-        monitor_task(delete_response.task)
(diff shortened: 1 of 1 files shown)


pulp_container/tests/functional/api/test_recursive_add.py
```diff
@@ -76,17 +76,11 @@ def test_missing_repository_argument(self):
...
-        monitor_task(delete_response.task)
         with self.assertRaises(ApiException) as context:
-            self.repositories_api.copy_manifests(
-                self.to_repo.pulp_href,
-                {
-                    # to_repo has no versions, use it as source
-                    "source_repository": self.to_repo.pulp_href,
-                }
-            )
+            self.repositories_api.delete(self.to_repo.latest_version_href)
```
dkliban, line 80: Where is the call to copy_manifests() ?

fabricio-aguiar: I wasn't sure if I had to make this call, as for now pulpcore does not even start the task:
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/repository.py#L208-L215

ipanova:
@fabricio-aguiar  @dkliban  i think we should just remove these 2 tests. Initially they were ensuring that an exception is raised when source_repository does not have latest version. Now in pulp any repo even newly created has repo version. In other words - we cannot have a repo in pulp that would not have a repo version.

PR #71, opened 2020-04-03
Assert sorted lists instead of using the functions all() and any()

lubosmj:
This change fixes the issue that caused the test to pass with a sucess if the fetched tags were matching only a subset of whitelisted tags (e.g. fetched_tags['tag_a', 'tag_a', 'tag_a'] == whitelist_tags['tag_a', 'tag_b', 'tag_c']).

In addition to that, a new test case, that verifies whether an invalid tag was included in a synced repository or not, was added.

[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_sync.py
+++ b/pulp_container/tests/functional/api/test_sync.py
@@ -200,6 +200,13 @@ class WhitelistedTagsSyncTestCase(unittest.TestCase):
 
         cls.repository = None
 
+    def test_sync_with_non_existing_whitelisted_tag(self):
+        """Check whether the sync machinery ignores a non-existing tag."""
+        whitelist_tags = ["manifest_a", "non_existing_manifest"]
+        self.sync_repository_with_whitelisted_tags(whitelist_tags)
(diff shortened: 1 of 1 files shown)


ipanova:
@lubosmj you can remove this variable since it is not used https://github.com/pulp/pulp_container/pull/71/files#diff-d59c6213e39251f834018812f59b6e3aR212

ipanova:
@lubosmj can you elaborate on how can you have in the fetched tags same 2 tags names? we have a repo_key where within a repo you can have only 1 tag with same name

lubosmj:
@ipanova, you are totally right, but the test case will not be working if a repository will return two tags with the same name one day. Let's say that this commit is just an improvement to the existing code.

ipanova:
If a repo will return 2 tags with same name one day that would mean that
our repo is malformed, so I think it would be actually better if the test
catches that and fails. wdyt?

lubosmj:
@ipanova, you are totally right, that is why I created this PR. It will catch that and fail, eventually.

PR #80, opened 2020-05-25
Enable s3

mdellweg:


pulp_container/app/registry.py
```diff
@@ -81,30 +82,48 @@ def _base_paths(path):
...
+            return file_response
+        elif settings.DEFAULT_FILE_STORAGE == 'storages.backends.s3boto3.S3Boto3Storage':
+            content_url = file.storage.url(
+                file.name,
+                parameters={
+                    "ResponseContentType": full_headers.pop("Content-Type"),
+                    "ResponseContentDisposition": full_headers.pop("Content-Disposition"),
+                },
+            )
+            raise HTTPFound(content_url, headers=full_headers)
```
ipanova, line 124: I think 302 should not contain headers from full_headers at all.

mdellweg: I agree, and i hope the tests do, too.

PR #86, opened 2020-05-29
Refactor push api

mdellweg:


pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -0,0 +1,67 @@
...
+        self.addCleanup(distributions_api.delete, self.distribution.pulp_href)
+
+    def test_push_using_podman(self):
+        """Test push with official registry client"""
+        # TODO better handling of the "http://"
+        local_url = urljoin(cfg.get_base_url(), self.distribution.base_path)[7:]
+        registry.pull("busybox:latest")
+        registry.tag("busybox:latest", local_url)
+        registry.push(local_url)
+        self.assertTrue(True)
```
ipanova, line 67: this assertion is not needed, each command from RegistryClient is checked upon completion of process on its returncode, if it will be !=0 an expection will be raised

PR #87, opened 2020-06-01
Create distribution and repository before initiation of 'docker push'

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -594,11 +594,15 @@ class BlobUploads(ViewSet):
         """
         This methods handles the creation of an upload.
         """
-        distribution = get_object_or_404(models.ContainerDistribution, base_path=path)
-        if distribution.repository:
-            repository = distribution.repository
-        else:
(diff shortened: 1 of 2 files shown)


pulp_container/app/viewsets.py
```diff
@@ -594,11 +594,14 @@ def create(self, request, path):
...
-        else:
-            raise Http404("Repository {} does not exist.".format(path))
+        # TODO add repo push type to distinguish from sync repo type
+        distribution, _ = models.ContainerDistribution.objects.get_or_create(name=path, base_path=path)
+        repository = distribution.repository
+        if not repository:
+            repo, _ = models.ContainerRepository.objects.get_or_create(name=path)
+            distribution.repository = repo
+            distribution.save()
+
```
dkliban, line 604: i think you need change the name 'repo' to 'repository' or assign one to the other.

ipanova: i updated the code, however it seems like it was working, that's why i did not notice this variable name typo.
it seems like 'docker push' retries when it pushes

PR #90, opened 2020-06-03
Adds a custom lookup value regex for the Manifest viewset.

dkliban:
This fixes a bug where manifest tags could not contain periods.

fixes: #6884
https://pulp.plan.io/issues/6884

### DIFF
--- a/pulp_container/app/content.py
+++ b/pulp_container/app/content.py
@@ -6,8 +6,6 @@ from pulp_container.app.registry import Registry
 
 registry = Registry()
 
-app.add_routes([web.get('/pulp/container/', registry.serve_v2)])
-app.add_routes([web.get('/pulp/container/_catalog', registry.list_repositories)])
 app.add_routes([web.get(r'/pulp/container/{path:.+}/blobs/sha256:{digest:.+}',
                         registry.get_by_digest)])
 app.add_routes([web.get(r'/pulp/container/{path:.+}/manifests/sha256:{digest:.+}',
--- a/pulp_container/app/urls.py
+++ b/pulp_container/app/urls.py
@@ -3,6 +3,7 @@ from rest_framework.routers import Route, SimpleRouter
 from pulp_container.app.viewsets import (
     Blobs,
     BlobUploads,
+    CatalogView,
     Manifests,
     VersionView
 )
@@ -25,5 +26,6 @@ router.register(r'^v2/(?P<path>.+)/manifests', Manifests, basename='manifests')
 
 urlpatterns = [
     url(r'^v2/$', VersionView.as_view()),
+    url(r'^v2/_catalog', CatalogView.as_view()),
     url(r'', include(router.urls))
 ]
--- a/pulp_container/app/urls.py
+++ b/pulp_container/app/urls.py
@@ -7,6 +7,7 @@ from pulp_container.app.viewsets import (
     VersionView
 )
 
+
 router = SimpleRouter(trailing_slash=False)
 
 head_route = Route(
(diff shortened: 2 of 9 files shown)


pulp_container/app/urls.py
```diff
@@ -7,7 +7,28 @@
     VersionView
 )
 
-router = SimpleRouter(trailing_slash=False)
+
+class RegistryRouter(SimpleRouter):
```
ipanova, line 12: @dkliban do i understand correctly that this change will be applied to the all the routes?
i was wondering if applying this regexp directly on the viewset won't be a more isolated change?
```
[ipanova@fluffy pulp_container]$ git diff
diff --git a/pulp_container/app/viewsets.py b/pulp_container/app/viewsets.py
index f78c0ef..cc0db59 100644
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -739,7 +739,7 @@ class Manifests(ViewSet):
     """
     ViewSet for intereacting with Manifests
     """
-
+    lookup_value_regex = '[^/^&*%!~@#+=?]+'
     # allow anyone to access
     authentication_classes = []
     permission_classes = []
```

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -36,32 +29,14 @@
...
         # TODO better handling of the "http://"
-        local_url = urljoin(cfg.get_base_url(), self.distribution.base_path)[7:]
+        local_url = urljoin(cfg.get_base_url(), 'foo/bar:1.0')[7:]
         registry.pull("busybox:latest")
         registry.tag("busybox:latest", local_url)
         registry.push(local_url)
-        self.assertTrue(True)
+        repository = repositories_api.list(name='foo/bar').results[0]
+        distribution = distributions_api.list(name='foo/bar').results[0]
+        self.addCleanup(repositories_api.delete, repository.pulp_href)
```
ipanova, line 41: why do we cleanup those resources? subsequent re-pushes should work  if repo/distribution  is already present

pulp_container/app/viewsets.py
```diff
@@ -595,8 +612,11 @@ def create(self, request, path):
         This methods handles the creation of an upload.
         """
         # TODO add repo push type to distinguish from sync repo type
-        distribution, _ = models.ContainerDistribution.objects.get_or_create(
-            name=path, base_path=path)
+        try:
```
ipanova, line 615: this try/expect is surplus here, get_or_create will whether get the exsiting distribution which is what is done on line 619 or create a new one.

Obviously , this will not work in case a base path is already taken and that distribution already serves some other repo. This case we will need to cover separately when we introduce sync/push repos

pulp_container/app/content.py
```diff
@@ -6,8 +6,6 @@
 
 registry = Registry()
 
-app.add_routes([web.get('/pulp/container/', registry.serve_v2)])
-app.add_routes([web.get('/pulp/container/_catalog', registry.list_repositories)])
```
ipanova, line 10: @dkliban  i think we can also move tags/list to the urls.py

dkliban: i agree

pulp_container/tests/functional/api/test_push_content.py
dkliban: Tests usually clean up anything they create. This test creates a repository and distribution.

pulp_container/app/urls.py
dkliban: good call. i'll make this change.

pulp_container/app/viewsets.py
dkliban: I did not mean to include this. I've removed it in the latest push.

PR #92, opened 2020-06-03
Create relations between content during push

ipanova:
closes #6827
https://pulp.plan.io/issues/6827


pulp_container/app/viewsets.py
```diff
@@ -831,9 +833,15 @@ def put(self, request, path, pk=None):
         # iterate over all the layers and create
         chunk = request.META['wsgi.input']
         artifact = self.receive_artifact(chunk)
+        with open(os.path.join(settings.MEDIA_ROOT, artifact.file.path)) as content_file:
```
fao89, line 836:
```python
from django.core.files.storage import default_storage as storage
content_file = storage.open(artifact.file.name)
```

PR #94, opened 2020-06-03
Add tags/list API endpoint

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/urls.py
+++ b/pulp_container/app/urls.py
@@ -5,6 +5,7 @@ from pulp_container.app.viewsets import (
     BlobUploads,
     CatalogView,
     Manifests,
+    TagsListView,
     VersionView
 )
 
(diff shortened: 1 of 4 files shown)


pulp_container/app/urls.py
```diff
@@ -23,6 +24,7 @@
 router.register(r'^v2/(?P<path>.+)/blobs/uploads\/?', BlobUploads, basename='docker-upload')
 router.register(r'^v2/(?P<path>.+)/blobs', Blobs, basename='blobs')
 router.register(r'^v2/(?P<path>.+)/manifests', Manifests, basename='manifests')
+router.register(r'^v2/(?P<path>.+)/tags/list', TagsList, basename='tags-list')
```
dkliban, line 27: you should not register this viewset with the router. Add it as a view to the urlpatterns below. 

This router creates multiple routes for the viewset. We just need a single one.

PR #98, opened 2020-06-04
Release 1.4.1

dkliban:


setup.py
```diff
@@ -10,7 +10,7 @@
 
 setup(
     name='pulp-container',
-    version='1.4.1.dev',
+    version='1.4.2.dev',
```
evgeni, line 13: shouldn't this (and all other places) say "1.4.1"? :)

evgeni: so, essentially, drop 3631159ed2ebfb1b3b2c0582ddd629f3d4c72440 from this PR :)

dkliban: There are multiple commits here. There is an intermediate commit with 1.4.1 and the changelog.

evgeni: from irc:

```
< Zhenech> so, you'll be tagging the intermediate?
< dkliban> that's right
< Zhenech> ewwww @ workflow
```

so shutting up :)

PR #103, opened 2020-06-15
Refactor token authentication

mdellweg:


mdellweg:
WIP: the redirect to content app preauthenticate part is not yet contained.

pulp_container/app/models.py
```diff
@@ -309,6 +312,14 @@ def get_repository_version(self):
         else:
             return None
 
+    def redirect_to_content_app(self, url):
+        """
+        Add preauthentication query string to redirect attempt.
+        """
+        if self.content_guard:
+            url = self.content_guard.cast().sign_url(url)
+        return HttpResponseRedirect(url)
```
lubosmj, line 321: Use `redirect()` instead of `HttpResponseRedirect`. It does the same thing and it looks a bit cleaner, see https://docs.djangoproject.com/en/3.0/topics/http/shortcuts/#redirect.

mdellweg:
I added the possibility to disable it again.
(It's an untested codepath though)

pulp_container/app/token_verification.py
```diff
@@ -1,186 +1,119 @@
...
+        if settings.get("TOKEN_AUTH_DISABLED", False):
+            return (AnonymousUser(), None)
 
-        headers = {
-            "Docker-Distribution-API-Version": "registry/2.0",
-            "Www-Authenticate": authenticate_header,
-        }
-        return headers
+        try:
+            authorization_header = request.headers["Authorization"]
```
ipanova, line 80: this code changes the behaviour where we return 401 even when hitting root endpoint (/v2 )

Some registries return 401 and others 200
https://registry-1.docker.io/v2/
https://quay.io/v2/

I am ok with both of approaches however I think that returning 401 and then asking for token auth, generating the token and then verifying it in the nuance is a surplus  work because the only thing that the client cares is 1) presence of registry 2.0 header and 4) whatever code is fine except 404 because what he looks at is the header to identify whether this is a v2 api registry otherwise he needs to fall back to the v1 api.
Thoughts?

CHANGES/6894.bugfix
```diff
@@ -0,0 +1 @@
+Refactored token_authentication to be available in api-app.
```
ipanova, line 1: how about -  Refactored token_authentication that now happens in pulpcore-api app

ipanova:
> I added the possibility to disable it again.
> (It's an untested codepath though)

I have tested and it does work. :balloon:

pulp_container/app/token_verification.py
mdellweg: I see both of the above examples returning 401.
Also, with the current implementation, i need to raise the `AuthenticationFailed` exception to spit out the `Www-Authenticate` header.

ipanova: ha, you are right, quay does return 401. I did through the browser so the "true" response kinda implied to me that it was a 200.

pulp_container/app/models.py
```diff
@@ -375,3 +386,54 @@ def append_chunk(self, chunk, chunk_size=None, save=True):
         self.file.close()  # Flush
         for algorithm in Artifact.DIGEST_FIELDS:
             setattr(self, algorithm, hashers[algorithm].hexdigest())
+
+
+def _gen_shared_secret():
```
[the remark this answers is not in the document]
mdellweg: Turns out djangos migration mechanism really does not like it. Sorry, i'm moving it out again.

mdellweg:
@ipanova want to rereview?

ipanova:
@mdellweg looks good to me, thank you!

PR #105, opened 2020-06-24
Push auth

mdellweg:
This is bases on #103

### DIFF
--- a/pulp_container/app/registry.py
+++ b/pulp_container/app/registry.py
@@ -242,7 +242,6 @@ class Registry(Handler):
         distribution = self._match_distribution(path)
         self._permit(request, distribution)
         repository_version = distribution.get_repository_version()
-        log.info(digest)
         try:
             ca = ContentArtifact.objects.get(
                 content__in=repository_version.content, relative_path=digest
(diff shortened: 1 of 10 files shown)


pulp_container/app/registry.py
```diff
@@ -242,7 +242,6 @@ async def get_by_digest(self, request):
         distribution = self._match_distribution(path)
         self._permit(request, distribution)
         repository_version = distribution.get_repository_version()
-        log.info(digest)
```
dkliban, line 245: Good. We were logging a lot :)

pulp_container/app/token_verification.py
```diff
@@ -24,17 +25,14 @@ def _decode_token(encoded_token):
...
-            decoded_token = {"access": []}
+        decoded_token = jwt.decode(encoded_token, public_key.read(), **JWT_DECODER_CONFIG)
     return decoded_token
 
 
 def _access_scope(request):
     repository_path = request.resolver_match.kwargs.get("path", "")
 
-    if request.method in ["GET", "HEAD", "OPTIONS"]:
+    if request.method in SAFE_METHODS:
```
dkliban, line 35: Nice use of predefined constant from DRF.

pulp_container/app/registry.py
mdellweg: This really looks like forgotten debugging code (no accompanying explanation around the bare digest).

pulp_container/app/token_verification.py
mdellweg: Yeah, i am still constantly learning about DRF. ;)

PR #106, opened 2020-06-29
Add ContainerPushRepository type

mdellweg:
fixes #6825
https://pulp.plan.io/issues/6825


pulp_container/app/viewsets.py
```diff
@@ -560,6 +564,63 @@ def __init__(self, blob, path, status, request, send_body=False):
...
+        dist_created = False
+        if create:
+            # TODO: Use serializer to create objects
+            distribution, dist_created = models.ContainerDistribution.objects.get_or_create(
+                base_path=path, defaults={"name": path},
+            )
+        else:
+            distribution = get_object_or_404(models.ContainerDistribution, base_path=path)
+        if dist_created:
+            repository = models.ContainerPushRepository.objects.create(name=path)
```
ipanova, line 608: can we add here try/except in case there is already a ContainerRepository with such name?

ipanova: also maybe we should delete the distribution if we fail with repo creation

mdellweg: Sure. What should be in the except block? Is it a 404, 409 or 500?
What if there is already a distribution, but no repository. Is auto creating the latter ok? Or does it mean someone reserved that name for later use?

mdellweg: We should put repo and distribution creation into a database transaction.

ipanova: and repo_version

ipanova: as we discussed per call - if existing distribution has no repo/repo_version assigned to it --> fail whole transaction

ipanova:
@mdellweg looks good to me, thank you for this work

PR #108, opened 2020-06-30
Replace whitelist with allowlist

fao89:
[noissue]


CHANGES.rst
```diff
@@ -192,7 +192,7 @@ Features
   `#4244 <https://pulp.plan.io/issues/4244>`_
```
bmbouter, line 192: I'm not involved in this project so feel free to disregard, but I don't think we can rewrite the history at all. Those older versions contain those older names no matter what.

ipanova:
@fao89 what prompted this rename?

CHANGES.rst
ipanova: that's correct. i don't think it depends on the project - changelog is not supposed to be rewritten

fao89:
> @fao89 what prompted this rename?

https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language

pulp_container/app/serializers.py
```diff
@@ -134,11 +134,11 @@ class ContainerRemoteSerializer(RemoteSerializer):
         allow_blank=False,
         help_text=_("Name of the upstream repository")
     )
-    whitelist_tags = serializers.ListField(
+    allowlist_tags = serializers.ListField(
         child=serializers.CharField(max_length=255),
         allow_null=True,
         required=False,
-        help_text=_("A list of whitelisted tags to sync")
+        help_text=_("A list of allowlisted tags to sync")
```
goosemania, line 141: maybe `allowed tags`?

goosemania:
I believe this needs to be tracked and there should be a changelog entry. It's a user-facing backward incompatible change.

fao89:
> I believe this needs to be tracked and there should be a changelog entry. It's a user-facing backward incompatible change.

I didn't realize it was a user-facing backward-incompatible change, I'll close this PR as it needs to be more planned

ipanova:
@fao89 Thank you for the PR, I will open a separate issue so we can sasess the impact and create roadmap with deprecation cycles .This is a user facing backwards incompatible change, our goal is to provide a not breaking user experience.

I will mentioned this PR in the issue to we can pick it up and re-open to not start the whole work from scratch.

pulp_container/app/serializers.py
ipanova: i would also suggest to consider `allowed_tag_list`

fao89:
> @fao89 Thank you for the PR, I will open a separate issue so we can asess the impact and create roadmap with deprecation cycles .This is a user facing backwards incompatible change, our goal is to provide a not breaking user experience.
> 
> I will mention this PR in the issue to we can pick it up and re-open to not start the whole work from scratch.

there is no need to reuse this PR, I think it worths to mention only the blog post:
https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language

PR #111, opened 2020-07-03
Refactor commons to ContainerRegistryApiMixin

mdellweg:
[noissue]

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -65,7 +65,6 @@ class UploadResponse(Response):
                 implementation of Response.
         """
         headers = {
-            "Docker-Distribution-Api-Version": "registry/2.0",
             "Docker-Upload-UUID": upload.pk,
             "Location": "/v2/{path}/blobs/uploads/{pk}".format(path=path, pk=upload.pk),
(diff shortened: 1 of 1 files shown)


pulp_container/app/registry_api.py
```diff
@@ -91,12 +90,8 @@ def __init__(self, manifest, path, request, status=200, send_body=False):
             send_body (bool): Whether a body should be sent with the response or just the headers.
         """
         artifact = manifest._artifacts.get()
-        if send_body:
```
ipanova, line 94: send_body is not used anywhere, any reason we are keeping it?

ipanova: i think we whether not send the content-length on a head request or if we do then we send the correct value we'd sent on a get request.

mdellweg: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13
I think we should send it. But i am happy to remove the send_body. (I would have thought it was meant to differentiate between HEAD and GET, but atm we always redirect on a GET). And this response never sends the body at all (i.e. `data=None` in the call to `super().__init__`.

PR #113, opened 2020-07-06
Redirect tag lookup to content app

mdellweg:
Also repaired schema conversion to work with s3 (django-storage).

fixes #6824
https://pulp.plan.io/issues/6824

### DIFF
--- a/pulp_container/tests/functional/api/test_pull_content.py
+++ b/pulp_container/tests/functional/api/test_pull_content.py
@@ -1,10 +1,11 @@
 # coding=utf-8
 """Tests that verify that images served by Pulp can be pulled."""
 import contextlib
+import requests
 import unittest
 from urllib.parse import urljoin
 
-from pulp_smash import cli, config, exceptions
+from pulp_smash import api, cli, config, exceptions
 from pulp_smash.pulp3.utils import (
     delete_orphans,
     get_content,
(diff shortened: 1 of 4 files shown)


pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -135,6 +138,32 @@ def test_api_returns_same_checksum(self):
...
+            queries.realm, params={"service": queries.service, "scope": queries.scope}
+        )
+        token = content_response.json()["token"]
+        content_response = self.client.get(
+            latest_image_url,
+            auth=BearerTokenAuth(token),
+            headers={"Accept": MEDIA_TYPE.MANIFEST_V1},
+        )
+        base_content_type = content_response.headers["Content-Type"].split(";")[0]
+        self.assertIn(base_content_type, {MEDIA_TYPE.MANIFEST_V1, MEDIA_TYPE.MANIFEST_V1_SIGNED})
```
mdellweg, line 165: Seems like we are returning `'application/vnd.docker.distribution.manifest.v1+prettyjws; charset=utf-8'` when asking for `'application/vnd.docker.distribution.manifest.v1+json'`. Is this intended?

dkliban: We probably shouldn't be doing that unless we are actually signing the manifest. Let's file a bug about it.

mdellweg: https://pulp.plan.io/issues/7113

PR #114, opened 2020-07-07
Create relations in a bulk

lubosmj:
closes #6940

### DIFF
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -459,18 +459,41 @@ class InterrelateContent(Stage):
         """
         Relate each item in the input queue to objects specified on the DeclarativeContent.
         """
-        async for dc in self.items():
+        async for batch in self.batches():
+            manifest_to_list_list = []
+            blob_list = []
+            config_blob_list = []
+            for dc in batch:
+                if dc.extra_data.get("relation"):
+                    manifest_to_list = self.relate_manifest_to_list(dc)
+                    manifest_to_list_list.append(manifest_to_list)
+                elif dc.extra_data.get("blob_relation"):
+                    blob = self.relate_blob(dc)
+                    blob_list.append(blob)
+                elif dc.extra_data.get("config_relation"):
+                    config_blob = self.relate_config_blob(dc)
+                    config_blob_list.append(config_blob)
+                elif dc.extra_data.get("man_relation"):
+                    self.relate_manifest_tag(dc)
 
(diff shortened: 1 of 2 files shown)


lubosmj:
To verify the results, I synced from the repository with 1,000 tags that was used during the performance testing. I synced from the local pulp2 repository to reduce the impact of downloaders.

- The sync time before my changes: ~16 minutes
- The sync time after my changes: ~6 minutes

There might be still a room for improvement. But after merging this PR along with https://github.com/pulp/pulpcore/pull/751 and https://github.com/pulp/pulpcore/pull/749, the performance should be comparable to pulp2.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -459,18 +459,35 @@ async def run(self):
...
+                    blob = self.relate_blob(dc)
+                    blob_list.append(blob)
+                elif dc.extra_data.get('config_relation'):
+                    config_blob = self.relate_config_blob(dc)
+                    config_blob_list.append(config_blob)
+                elif dc.extra_data.get('man_relation'):
+                    self.relate_manifest_tag(dc)
+
+            ManifestListManifest.objects.bulk_create(objs=manifest_to_list_list,
+                                                     ignore_conflicts=True,
```
mdellweg, line 480: Why is black not removing the hanging indent here?

mdellweg: I see now, you made the PR against master. Should it be 2.0? @dkliban

lubosmj: I changed the indentation and replaced all single quotes with double quotes to prevent possible conflicts when this commit will be merged into 2.0 as well.

mdellweg: That is in all cases a good idea. Thank you.
I still want to settle the question whether this goes into 2.0 or master.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -480,9 +503,9 @@ def relate_config_blob(self, dc):
             dc (pulpcore.plugin.stages.DeclarativeContent): dc for a Blob
```
mdellweg, line 504: Maybe we should document that they return something now.

ipanova:
It should go to master and cherry picked into 2.0

pulp_container/app/tasks/sync_stages.py
dkliban: We should merge into master and cherry-pick into 2.0 for now.

PR #115, opened 2020-07-08
Render exception as JSON

mdellweg:
fixes #7054

https://pulp.plan.io/issues/7054


dkliban:
I tested that doing a pull on non-existing repo gives a proper error message.
I tested that doing a push without authentication produces a nice error message.
I tested that once I login using admin credentials I am able to push.

PR #117, opened 2020-07-09
Merge instead of cherrypick

mdellweg:


mdellweg:
I think we decided to not go that way... Closing it.

PR #120, opened 2020-07-13
Make the model Tag artifactless

lubosmj:
closes #6816


dkliban:
We should probably add a migration that removes the ContentArtifacts for Tags.

lubosmj:
I added a migration which removes the relation to an Artifact.

Before:
```
In [6]: Tag.objects.get(name='1.26.1-uclibc').tagged_manifest._artifacts.get().sha256                                
Out[6]: 'ae007bdb45fc0d56e3d705b97640ac24844bcc9ce4c8b8493f216a57ab6af0d5'

In [7]: Tag.objects.get(name='1.26.1-uclibc')._artifacts.get().sha256                                                
Out[7]: 'ae007bdb45fc0d56e3d705b97640ac24844bcc9ce4c8b8493f216a57ab6af0d5'
```

After:
```
In [2]: Tag.objects.get(name='1.26.1-uclibc').tagged_manifest._artifacts.get().sha256                                 
Out[2]: 'ae007bdb45fc0d56e3d705b97640ac24844bcc9ce4c8b8493f216a57ab6af0d5'

In [3]: Tag.objects.get(name='1.26.1-uclibc')._artifacts.all()                                                        
Out[3]: <QuerySet []>
```

lubosmj:
@mdellweg, I have not checked the push API. That tick sign means that everything is fine now?

mdellweg:
@lubosmj sorry, the tick sign was for creating tags via the pulp-api using the serializers. I didn't find an empty box.

I had a look at the code of the registry-push, and nothing stuck out there. But i would be more confident if you could test that for real (or @dkliban saying it was never creating the ContentArtifact).

lubosmj:
@mdellweg, I tried to push a tag to the registry and and a tag object did not hold a reference to an image manifest's artifact anymore. The image manifest's artifact was referenced only by the corresponding manifest, as is should be.

PR #123, opened 2020-07-13
Return an HTTP 404 response when a manifest was not converted

lubosmj:
closes #7125

### DIFF
--- a/CHANGES/7125.bugfix
+++ b/CHANGES/7125.bugfix
@@ -1,0 +1,2 @@
+Fixed a runtime error which was triggered when a registry client sends an accept header with an
+inappropriate media type for a manifest and the conversion failed.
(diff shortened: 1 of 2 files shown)


CHANGES/7125.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed the error that caused a runtime error when a wrong accept header was specified during a pull
+request
```
mdellweg, line 2:
```suggestion
Fixed a runtime error when an old registry client sends accept header for manifest_v1 and the
conversion failed.
```

lubosmj: @mdellweg, I fixed the wording. The error was raised when using an inappropriate accept header. It does not matter whether it was send by an old client or a new client. Usually, the new client sends accept headers containing both `application/vnd.docker.distribution.manifest.v2+json` and `application/vnd.docker.distribution.manifest.list.v2+json` that is why we had not noticed this issue in the past. How do you like it now?

PR #126, opened 2020-07-17
Prevent distributing push repository versions

mdellweg:
fixes #7012
https://pulp.plan.io/issues/7012

### DIFF
--- a/CHANGES/7012.bugfix
+++ b/CHANGES/7012.bugfix
@@ -1,0 +1,1 @@
+Added a validation, that a push repository cannot be distributed by specifying a version.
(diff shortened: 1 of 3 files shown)


CHANGES/7012.feature
```diff
@@ -0,0 +1 @@
+Added a validation, that a push repository cannot be published by specifying a version.
```
ipanova, line 1: s/published/distributed

ipanova: i think that changelong should be a bugfix rather than a feature, wdyt?

PR #127, opened 2020-07-20
Forbid the methods PATCH and PUT for push repositories

lubosmj:
closes #7013

### DIFF
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -21,6 +21,7 @@ from pulpcore.plugin.viewsets import (
     CharInFilter,
     ContentFilter,
     ContentGuardViewSet,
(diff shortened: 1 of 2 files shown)


pulp_container/app/viewsets.py
```diff
@@ -401,9 +404,27 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
     parent_viewset = ContainerRepositoryViewSet
 
 
-class ContainerPushRepositoryViewSet(RepositoryViewSet):
+class NoUpdateRepositoryViewSet(
```
mdellweg, line 407: Just an idea: How about `ImmutableRepositoryViewSet`?

lubosmj:
This PR will be converted back to a normal PR after the next release of pulpcore.

PR #128, opened 2020-07-21
OpenAPI v3 with drf-spectacular

fao89:
Required PR: https://github.com/pulp/pulpcore/pull/785
[noissue]


mdellweg:
When running `http :24817/pulp/api/v3/docs/api.json | jq '.paths|keys'`, i see among others `"{blob_href}"`.
I believe this means that we do not properly namespace the href parameters of content models.
Yes, this was the case before, but now would be the perfect time to repair that, as this is a breaking change anyway.

PR #129, opened 2020-07-27
Merge 2.0 into master

dkliban:


mdellweg:
Without badly tricking the CI here, we will not make it pass. Should we merge this anyway?

requirements.txt
```diff
@@ -1,4 +1,5 @@
-pulpcore>=3.4
+pulpcore>=3.6
```
mdellweg, line 1:
```suggestion
pulpcore>=3.6.0.dev0
```
I needed this to pip install it with current pulpcore master.

PR #131, opened 2020-08-07
Delete push repositories with their distribution

mdellweg:
Both entities can be used to trigger the deletion, and the deletions are
coordinated in an atomic transaction.

fixes #7172
https://pulp.plan.io/issues/7172


pulp_container/app/viewsets.py
```diff
@@ -433,6 +455,30 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
...
+    def destroy(self, request, pk, **kwargs):
+        """
+        Delete a push repository with its distribution
+        """
+        distribution = self.get_object()
+        reservations = [distribution]
+        instance_ids = [
+            (distribution.pk, "container", "ContainerDistributionSerializer"),
+        ]
+        if distribution.repository and distribution.repository.PUSH_ENABLED:
```
ipanova, line 471: ```AttributeError: 'Repository' object has no attribute 'PUSH_ENABLED'```

You need to use `cast()`

pulp_container/app/tasks/base.py
```diff
@@ -0,0 +1,21 @@
+from django.db import transaction
+
+from pulpcore.app.apps import get_plugin_config
+
+
+def general_multi_delete(instance_ids):
+    """
+    Delete a list of model in a transaction
```
ipanova, line 8: s/model/models/

mdellweg: Wdyt about "model instances"?

pulp_container/app/viewsets.py
```diff
@@ -433,6 +455,30 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
     queryset = models.ContainerDistribution.objects.all()
     serializer_class = serializers.ContainerDistributionSerializer
 
+    @extend_schema(
+        description="Trigger an asynchronous delete task",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def destroy(self, request, pk, **kwargs):
+        """
+        Delete a push repository with its distribution
```
ipanova, line 464: Delete a distribution. If a push repository is associated to it, delete it as well.

PR #133, opened 2020-08-08
Use PulpTemporaryFile to store uploaded chunks

lubosmj:
closes #7218


pulp_container/app/migrations/0004_upload.py
```diff
@@ -14,26 +14,4 @@ class Migration(migrations.Migration):
...
-                ('sha384', models.CharField(max_length=96, null=True)),
-                ('sha512', models.CharField(max_length=128, null=True)),
-                ('repository', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='uploads', to='core.Repository')),
-            ],
-            options={
-                'abstract': False,
-            },
-        ),
-    ]
+    operations = []
```
mdellweg, line 17: Please do not modify existing migrations.

lubosmj: I was getting errors when I removed the function `generate_filename()`: 
https://github.com/pulp/pulp_container/blob/13800035477c6d43ebb78af33262c934551d5490/pulp_container/app/models.py#L362-L365.
That is why I purged this migration. Do you propose to leave the function there?

mdellweg: In that case, i guess yes.

CHANGES/7218.bugfix
```diff
@@ -0,0 +1,2 @@
+Refactored the registry's push API to not store uploaded chunks in /var/lib/pulp, but rather
```
ipanova, line 1: i'd rather mark this a feature then bugfix

mdellweg:
I am quite confused by the size of this PR. Can you elaborate, why subclassing is neccessary instead of "just" replacing the temporary_file with a pulp_temporary_file?
Did you fix additional issues with the registry upload api?

lubosmj:
@mdellweg, the problem here is that you cannot open a file in the append mode in S3. Therefore, you should not just simply replace the instance of a temporary file. You may either delete and write an updated file back or create a bunch of files that will be later assembled into a single file. I decided to go for the latter.

So, after merging the proposed changes, the registry will work in the following way (assuming that a database server is accessible to multiple Pulp instances):
1. Create an Upload object (identifying one upload within a repository).
2. Save each uploaded chunk to the S3 storage as a temporary file.
3. When the upload finishes, build an artifact from the temporary files.
4. Delete temporary files and the Upload object.

Speaking of additional issues with the upload API, I did not fix anything. It should work like it was working in the past (let's say for now that it was really working). When I was working on this change, the code was seemingly working because the test passed. However, I have to made a few additional changes because podman behaved differently compared to docker.

pulp_container/app/models.py
```diff
@@ -367,57 +367,52 @@ def generate_filename(instance, filename):
...
     """
-    Model for tracking Blob uploads.
+    A model for tracking Blob uploads.
     """
 
     repository = models.ForeignKey(Repository, related_name="uploads", on_delete=models.CASCADE)
+    cumulative_size = models.BigIntegerField(default=0)
+
+
+class BlobTemporaryUpload(PulpTemporaryFile):
```
mdellweg, line 377: Should this be an `UploadChunk` then?

pulp_container/app/models.py
```diff
@@ -367,57 +367,52 @@ def generate_filename(instance, filename):
...
+    """
+    A model used for storing uploaded blob chunks in a temporary file.
+    """
```
mdellweg, line 380: Would it make sense (and maybe the code easier) if you didn't inherit from `PulpTemporaryFile`, but have a one-to-one relation with cascaded delete here?
It would render this model to be a rich (with the additional `offset`) join table from `Upload` to `PulpTemporaryFile`.

pulp_container/app/__init__.py
```diff
@@ -1,8 +1,21 @@
...
+from pulp_container.app.signals import delete_from_storage
+
 
 class PulpContainerPluginAppConfig(PulpPluginAppConfig):
     """Entry point for the container plugin."""
 
     name = "pulp_container.app"
     label = "container"
+
+    def ready(self):
```
mdellweg, line 14: I think, this should be handled in `pulpcore` by using `PulpTemporaryFile`.

pulp_container/app/models.py
```diff
@@ -367,57 +367,52 @@ def generate_filename(instance, filename):
...
-    file = models.FileField(
-        max_length=255,
-        null=True,
-        upload_to=generate_filename,
-        storage=FileSystemStorage(location=settings.CHUNKED_UPLOAD_DIR),
-    )
+    def save_chunk(self, chunk, chunk_size=None):
+        """Save the passed chunk to a temporary file and update offset for a current upload."""
+        self._init_temporary_file(chunk)
+        self._update_upload_size(chunk, chunk_size)
```
mdellweg, line 388: Do those two methods not know `chunk` and `upload` from `self`?

pulp_container/app/models.py
```diff
@@ -367,57 +367,52 @@ def generate_filename(instance, filename):
...
+        with NamedTemporaryFile("ab") as temp_file:
+            while True:
+                subchunk = chunk.read(2000000)
+                if not subchunk:
+                    break
+                temp_file.write(subchunk)
 
+            temp_file.flush()
+
+            self.file = File(open(temp_file.name, "rb"))
```
mdellweg, line 401: Also this part i expect to be handled by using `PulpTemporaryFile`.

pulp_container/app/registry_api.py
```diff
@@ -442,41 +448,47 @@ def put(self, request, path, pk=None):
...
-                blob_artifact = ContentArtifact(
-                    artifact=artifact, content=blob, relative_path=digest
-                )
-                blob_artifact.save()
-            except IntegrityError:
-                pass
+        chunks = models.BlobTemporaryUpload.objects.filter(upload=upload).order_by("offset")
+        chunks_files = map(lambda chunk: chunk.file, chunks)
+
+        with NamedTemporaryFile("ab") as temp_file:
```
mdellweg, line 454: Isn't there a "turn this `PulpTemporaryFile` into an artifact primitive suitable for this?

pulp_container/app/models.py
lubosmj: No. Not in the time of initialization.

pulp_container/app/models.py
lubosmj: In this case, we are dealing specifically with a chunk object which behaves differently compared to an ordinary file object. I think you need to read chunk in chunks, like so: `chunk.read(X)`. In `PulpTemporaryFile`, we are dealing just with objects of the type `File` or `PulpTemporaryUploadedFile`.

lubosmj: This would require me to implement if else checks for the object's type in `PulpTemporaryFile`, would not?

pulp_container/app/registry_api.py
lubosmj: No, since I am constructing an Artifact object from a bunch of PulpTemporaryFile objects, like I did mention above.

mdellweg: Sorry, i thought you assembled all the chunks into a new PTFile, but then again you can just create the Artifact directly.

pulp_container/app/models.py
mdellweg: Isn't the chunk just a stream of raw data that needs to be written to a file (on whatever storage) for later use?

Maybe we should have broken this change down in two steps:
1. Handle storage without append
2. Move from local files to PTFile.

pulp_container/app/registry_api.py
lubosmj: I cannot create an Artifact directly because it is not possible to open files in the append mode. I have to iterate through chunks, merge them, and then create an Artifact on a single write.

pulp_container/app/models.py
lubosmj: Yes, it is.

1. The tests for S3 would not pass because I will be trying to open a file in the append mode.
2. This is what I actually did.

pulp_container/app/registry_api.py
mdellweg: Yeah, that is what i meant. I first thought you'd "upload" that (locally) assembled file into a new PTFile to make an artifact from it. But that step is clearly unnecessary, because you can create the artifact directly from that assembled file.

lubosmj:
I am putting this PR on hold due to the recent findings of errors in `PulpTemporaryFile`. The discussion about whether we want to handle the removal of temporary files in plugins or in pulpcore was moved here: https://github.com/pulp/pulpcore/pull/844.

lubosmj:
It looks like all the issues related to `PulpTemporaryFile` have been resolved. I will continue working on this PR soon.

lubosmj:
The issue has been addressed in pulpcore. Refer to https://github.com/pulp/pulpcore/pull/914.

lubosmj:
I am closing this PR in favour of https://github.com/pulp/pulpcore/pull/914. Additional changes (if any) will be made in a separate PR. The PR will reference the same issue number.

mdellweg:
@lubosmj with https://github.com/pulp/pulpcore/pull/914 merged, this can be picked up again, right?

PR #134, opened 2020-08-10
Replace whitelist_tags with include/exclude_tags

mdellweg:
fixes #6922
https://pulp.plan.io/issues/6922
fixes #7070
https://pulp.plan.io/issues/7070


pulp_container/app/serializers.py
```diff
@@ -139,11 +139,29 @@ class ContainerRemoteSerializer(RemoteSerializer):
...
+            """
+        ),
+    )
+    exclude_tags = serializers.ListField(
+        child=serializers.CharField(max_length=255),
+        allow_null=True,
+        required=False,
+        help_text=_(
+            """
+            A list of excluded tags to sync.
```
ipanova, line 160: A list of tags to exclude during sync

pulp_container/app/serializers.py
```diff
@@ -139,11 +139,29 @@ class ContainerRemoteSerializer(RemoteSerializer):
...
     )
-    whitelist_tags = serializers.ListField(
+    include_tags = serializers.ListField(
         child=serializers.CharField(max_length=255),
         allow_null=True,
         required=False,
-        help_text=_("A list of whitelisted tags to sync"),
+        help_text=_(
+            """
+            A list of included tags to sync.
```
ipanova, line 148: A list of tags to include during sync

pulp_container/tests/functional/api/test_sync.py
```diff
@@ -245,13 +256,12 @@ def sync_repository_with_whitelisted_tags(self, whitelist_tags):
         sync_response = self.repositories_api.sync(self.repository.pulp_href, repository_sync_data)
         monitor_task(sync_response.task)
         repository = self.repositories_api.read(self.repository.pulp_href)
-        self.assertIsNotNone(repository.latest_version_href)
 
-    def assert_synced_tags(self, whitelist_tags):
-        """Check if the created repository contains only the selected whitelisted tags."""
+    def assert_synced_tags(self, include_tags):
```
ipanova, line 260: what about adding a test where same tag is present in both places?
I think then this assertion will need to be changed because in this case list of tags wont be equal to `include_tags`

ipanova:
@mdellweg can you please update the docs here as well? https://github.com/pulp/pulp_container/blob/6928a0b94d228d7b49e680944eef14a1c0830555/docs/index.rst#features

pulp_container/tests/functional/api/test_sync.py
mdellweg: The 'complex filtering' test is doing this?
Maybe the name of this parameter is misleading. I can rename it to `expected_tags`.

ipanova: right, i see that now! +1 to rename

ipanova:
@mdellweg i think you need to re-generate the docs again, it's too many lines in this PR :scream:

mdellweg:
> @mdellweg i think you need to re-generate the docs again, it's too many lines in this PR scream

It is a generated asset. It should not have been there to begin with.

PR #136, opened 2020-08-11
Repair TOKEN_AUTH_DISABLED

mdellweg:
[noissue]


ipanova:
tested with true/false pull/push :taco:

PR #143, opened 2020-08-25
Issue one redirect for the S3 storage

lubosmj:
closes #6826

### DIFF
--- a/pulp_container/app/content.py
+++ b/pulp_container/app/content.py
@@ -6,9 +6,9 @@ from pulp_container.app.registry import Registry
 registry = Registry()
 
 app.add_routes(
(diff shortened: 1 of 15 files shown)


pulp_container/app/content.py
```diff
@@ -6,9 +6,19 @@
 registry = Registry()
 
 app.add_routes(
-    [web.get(r"/pulp/container/{path:.+}/blobs/sha256:{digest:.+}", registry.get_by_digest,)]
+    [
+        web.get(
+            r"/pulp/container/{path:.+}/blobs/sha256:{digest:.+}",
+            registry.get_by_digest,
+        )
+    ]
```
mdellweg, line 14: I think, by removing the last comma, we can keep it a one-liner. 
The question is, which one is more readable?

mdellweg (approved):

PR #145, opened 2020-08-31
Fixing black issues with new version of black

fao89:
[noissue]


lubosmj:
I think we already addressed the black formatting in the commit https://github.com/pulp/pulp_container/commit/9a6ae47fbeab1b069b8364b22a76f35e09914c4c.

PR #146, opened 2020-09-08
Use the path and query string of the URL to generate a 'validate_token'.

dkliban:
fixes: #7462
https://pulp.plan.io/issues/7462


mdellweg (approved):
I think, we should still include the querystring.

mdellweg (approved):
This looks good!

PR #148, opened 2020-09-08
Remove unused variable

ekohl:
In d23235cfc2e57266715ce1b124c757c7e7416663 the last use of this variable was removed.

I ran into this when trying to understand the use of CONTENT_ORIGIN but I'm unfamiliar with the codebase so I may be missing something.

### DIFF
--- a/pulp_container/app/authorization.py
+++ b/pulp_container/app/authorization.py
@@ -1,4 +1,3 @@
-import re
 import base64
 import hashlib
 import random
@@ -13,8 +12,6 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 
 TOKEN_EXPIRATION_TIME = 300
-
-KNOWN_SERVICES = [re.sub(r"(http://|https://)", "", settings.CONTENT_ORIGIN, count=1)]
 
 
 class AuthorizationService:


mdellweg:
Hey, thank you for finding this!
It seems the associated test to this variable has gone missing with the refactoring in line
https://github.com/pulp/pulp_container/commit/d23235cfc2e57266715ce1b124c757c7e7416663#diff-144ea72fc60076dffb457b0e24e1a420L52
Maybe we should add the test again, but it would need to test against the other address now.
@dkliban any thoughts?

ipanova:
@mdellweg seems like refactor removed the check of the `service` the token is generated against. we should add back https://github.com/pulp/pulp_container/commit/d23235cfc2e57266715ce1b124c757c7e7416663#diff-144ea72fc60076dffb457b0e24e1a420L179

ipanova:
I looked some more into the code and I think that there is no need to add the test back, plus we can remove the `KNOWN_SERVICES` variable. The audience(aka service)  will be checked  when token decoded https://github.com/pulp/pulp_container/blob/5010767c1a26e03c63cf6d18414b999c28afeeca/pulp_container/app/token_verification.py#L28 and error will be raised if audience will will not match https://pyjwt.readthedocs.io/en/latest/api.html?highlight=InvalidAudienceError#jwt.exceptions.InvalidAudienceError
That said, it is enough to check for service and scope if provided in the request, and we are doing it here https://github.com/pulp/pulp_container/blob/master/pulp_container/app/registry_api.py#L323

ipanova (approved):

mdellweg:
@ipanova Thank you for diving into this.

mdellweg (approved):

PR #151, opened 2020-09-17
Documented how include/exclude_tags options work with mirror=True/False.

ipanova:
closes #7380
https://pulp.plan.io/issues/7380

### DIFF
--- a/docs/workflows/sync.rst
+++ b/docs/workflows/sync.rst
@@ -84,6 +84,8 @@ sync with. You are telling pulp to fetch content from the remote and add to the 
     In the above example, the payload contains the field ``mirror=False``. This means that the
     sync will be run in the additive mode only. Set ``mirror`` to ``True`` and Pulp will pull
     in new content and remove content which was also removed from upstream.
+    The same logic will be applied when ``include/exclude_tags`` are specified together with 
+    the ``mirror`` command, but only on the subset of tags.
 
 .. note::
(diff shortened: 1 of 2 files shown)


docs/workflows/sync.rst
```diff
@@ -83,7 +83,9 @@ sync with. You are telling pulp to fetch content from the remote and add to the
 .. note::
     In the above example, the payload contains the field ``mirror=False``. This means that the
     sync will be run in the additive mode only. Set ``mirror`` to ``True`` and Pulp will pull
-    in new content and remove content which was also removed from upstream.
+    in new content and remove content which was also removed from upstream. Same logic will be
+    applied in case ``include/exclude_tags`` are specified together with ``mirror`` comment, but
+    only on the subset of tags.
```
RCMariko, line 88: Suggest slight tweak in English: The same logic will be
    applied when ``include/exclude_tags`` is specified with the ``mirror`` field, but
    only on the subset of tags.

And in this you mean when any value is provided for the "mirror" field or a specific one? (more of an English comment than technical, obviously)

ipanova: sorry i had a typo, it was meant to be `mirror command` and yes the same logic will be applied to the `mirror` value respectively.

RCMariko (approved):

mdellweg (approved):

PR #153, opened 2020-09-22
Documented how include/exclude_tags options work with mirror=True/False.

dkliban:
closes #7380
https://pulp.plan.io/issues/7380


bmbouter (approved):
Looks good to me. Thanks!

PR #155, opened 2020-09-23
Building changelog for 2.0.2

ipanova:
[noissue]

(cherry picked from commit 5c96295b61ed6cd8c7e353e9442bcf9916d60278)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,20 @@ Changelog
 
 .. towncrier release notes start
 
+2.0.2 (2020-09-22)
+==================
+
+
+Improved Documentation
+----------------------
+
+- Documented how include/exclude_tags options work with mirror=True/False.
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

dkliban:
This looks good @ipanova. We also need to change the minimum required version of pulpcore to 3.7. This can be done as a separate PR though.

mdellweg:
Shouldn't we keep the minimum version as low as possible? Is there a thing in core 3.7 we already rely on?

dkliban:
@pulpbot Yes, it's this change related to the OpenAPI schema https://github.com/pulp/pulpcore/commit/ee76dda3eef273c32e8101b3e65edb26cffed358

dkliban:
We decided that since the pulpcore version needs to be bumped, we should release 2.1.0 instead. I am closing this PR.

PR #158, opened 2020-09-24
Add namespace to distributions

mdellweg:
fixes #7089
https://pulp.plan.io/issues/7089

This is the first version to pass ci. Still missing are at least:
- [x] a data migration
~- [ ] validation of namespaces to be collision free with non namespaced distribution~ - The new design handles this on the database table level.


mdellweg:
I just realized, that if paths without a "/" still had a namespace (in that case with the same name), the uniqueness would be trivially given by the database.

pulp_container/app/serializers.py
```diff
@@ -230,6 +241,25 @@ def validate(self, data):
...
+
+        if "namespace" in validated_data:
+            if not base_path.startswith(validated_data["namespace"].name + "/"):
+                raise serializers.ValidationError(
+                    _("Selected namespace does not match first component of base_path.")
+                )
+        elif "/" in base_path:
+            namespace_name = base_path.split("/")[0]
+            try:
+                validated_data["namespace"] = models.Namespace.objects.get(name=namespace_name)
```
ipanova, line 254: how about using get_or_create instead?

pulp_container/app/serializers.py
```diff
@@ -230,6 +241,25 @@ def validate(self, data):
                         "repository version."
                     )
                 )
+        base_path = validated_data["base_path"]
+
+        if "namespace" in validated_data:
+            if not base_path.startswith(validated_data["namespace"].name + "/"):
```
ipanova, line 247: :taco:  :+1:

ipanova:
i know this change is not really related to the PR but can you please update and remove the publication reference https://github.com/pulp/pulp_container/pull/158/files#diff-90a314d7b2d5ddb20d360be02a4ec146L333

pulp_container/app/viewsets.py
```diff
@@ -482,3 +483,19 @@ class ContentRedirectContentGuardViewSet(ContentGuardViewSet):
...
+    mixins.CreateModelMixin,
+    mixins.DestroyModelMixin,
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+):
+    """
+    ViewSet for ContainerNamespaces.
+    """
+
+    endpoint_name = "pulp_container/namespaces"
```
ipanova, line 499: the changes look good to me, @dkliban do you want to take a look? do you have a different suggestion on the endpoint or we leave this one as is?

pulp_container/app/serializers.py
```diff
@@ -230,6 +241,25 @@ def validate(self, data):
...
+            try:
+                validated_data["namespace"] = models.ContainerNamespace.objects.get(
+                    name=namespace_name
+                )
+            except ObjectDoesNotExist:
+                namespace_serializer = ContainerNamespaceSerializer(data={"name": namespace_name})
+                namespace_serializer.is_valid(raise_exception=True)
+                validated_data["namespace"] = namespace_serializer.create(
+                    namespace_serializer.validated_data
+                )
```
jlsherrill, line 262: there's room here for a race condition, if two repos are being created with a new namespace at the same time it could result in one of the 'creates' to fail when the other task created it.  Not sure how often ya'll deal with that within pulp, but we see it all the time within katello

mdellweg: Sounds like we really need a sophisticated get_or_create on the serializer model.
I think we do that in >=3 places.

mdellweg: I added another except block to recognize a lost race.

pulp_container/app/models.py
```diff
@@ -330,11 +341,18 @@ class Meta:
...
     """
-    A container distribution defines how a publication is distributed by Pulp's webserver.
+    A container distribution defines how a repository version is distributed by Pulp's webserver.
     """
 
     TYPE = "container"
 
+    namespace = models.ForeignKey(
+        ContainerNamespace,
+        on_delete=models.CASCADE,
```
ipanova, line 351: :+1:

pulp_container/app/serializers.py
```diff
@@ -27,6 +29,26 @@
...
+class _GetOrCreateMixin:
+    @classmethod
+    def get_or_create(cls, natural_key, default_values=None):
+        try:
+            result = cls.Meta.model.objects.get(**natural_key)
+        except ObjectDoesNotExist:
+            data = {}
+            if default_values is not None:
+                data.update(default_values)
+            data.update(natural_key)
```
ipanova, line 41: data seems to be unused

ipanova:
```$ http GET :24817/pulp/api/v3/pulp_container/namespaces/
HTTP/1.1 200 OK
Connection: close
Content-Length: 112
Date: Mon, 26 Oct 2020 13:48:14 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "count": 4,
    "next": null,
    "previous": null,
    "results": [

        {
            "name": "bar"
        },
        {
            "name": "foo"
        }
    ]
}
```
I suggest adding pulp_href pulp_created and maybe list of distributions

pulp_container/app/viewsets.py
ipanova: i looked at 'django-admin show_urls' and i would not be opposed also having /namespaces/container/ as another option to consider

pulp_container/app/serializers.py
mdellweg: But it should be... Changing that.

pulp_container/app/serializers.py
```diff
@@ -111,6 +133,16 @@ def to_representation(self, value):
         return "".join([host.split("//")[-1], "/", value])
 
 
+class ContainerNamespaceSerializer(ModelSerializer, _GetOrCreateMixin):
+    """
+    Serializer for ContainerNamespaces.
+    """
+
+    class Meta:
+        fields = ModelSerializer.Meta.fields + ["name"]
```
ipanova, line 142: we should use a tuple here instead of list

mdellweg:
> I suggest adding pulp_href pulp_created and maybe list of distributions

As the list of distributions can get very large, i think it would be better to make the distribution filterable by namespace.

ipanova:
> > I suggest adding pulp_href pulp_created and maybe list of distributions
> 
> As the list of distributions can get very large, i think it would be better to make the distribution filterable by namespace.

That's a fine idea, for that we would need to add NamespaceFilter to the DistributionViewset.
Right now we are not serializing the namespace on the distribution, should we given that ^ change?

ipanova:
while i can filter the distributions by the name of the namespace i get 500 on the namespace endpoint
```
$ http GET :24817/pulp/api/v3/distributions/container/container/?namespace__name=my-repo1
HTTP/1.1 200 OK
Connection: close
Content-Length: 637
Date: Mon, 02 Nov 2020 17:32:35 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "base_path": "my-repo1",
            "content_guard": "/pulp/api/v3/contentguards/container/content_redirect/acbae288-e7cf-471f-bbd7-e88bb27300a6/",
            "name": "my-repo1",
            "namespace": "/pulp/api/v3/pulp_container/namespaces/4e565f62-aec7-48bb-8746-80cd412a69b5/",
            "pulp_created": "2020-11-02T17:27:59.566894Z",
            "pulp_href": "/pulp/api/v3/distributions/container/container/7252915c-53f6-4ba4-bcec-caff34c25d64/",
            "registry_path": "pulp3-source-fedora32.fluffy.example.com/my-repo1",
            "repository": "/pulp/api/v3/repositories/container/container-push/ade7c212-edee-4253-af06-738c6449862c/",
            "repository_version": null
        }
    ]
}

(pulp) [vagrant@pulp3-source-fedora32 _scripts]$ http GET :24817/pulp/api/v3/pulp_container/namespaces/4e565f62-aec7-48bb-8746-80cd412a69b5/
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Length: 27
Content-Type: text/html
Date: Mon, 02 Nov 2020 17:32:41 GMT
Server: gunicorn/20.0.4
Vary: Cookie
X-Frame-Options: SAMEORIGIN

<h1>Server Error (500)</h1>

```

mdellweg:
Somehow the `DetailIdentityField` derives the view as `"containernamespace-detail"`, while it is `"pulp_container/namespaces-detail"`. This needs more investigation.

pulp_container/app/serializers.py
```diff
@@ -111,6 +134,18 @@ def to_representation(self, value):
         return "".join([host.split("//")[-1], "/", value])
 
 
+class ContainerNamespaceSerializer(ModelSerializer, _GetOrCreateMixin):
+    """
+    Serializer for ContainerNamespaces.
+    """
+
+    pulp_href = IdentityField(view_name="pulp_container/namespaces-detail")
```
mdellweg, line 142: @ipanova this should fix it.

ipanova (approved):

PR #166, opened 2020-10-12
Remove outdated AUTHORS file.

ipanova:
[noissue]


ipanova:
@bmbouter do you know if there is some legal implications in removing this file?

bmbouter:
> @bmbouter do you know if there is some legal implications in removing this file? these are leftovers from pulp_docker repo.

I am not a lawyer, but I don't believe this file changes a contributors legal right to their copyright. The git log I believe serves as a more accurate source anyway.

dkliban (approved):

mdellweg (approved):

PR #169, opened 2020-10-16
Get API host from the request's context

lubosmj:
closes #7385

### DIFF
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -1,7 +1,6 @@
 from gettext import gettext as _
 import os
 
-from django.conf import settings
 from django.db import IntegrityError
 from django.core.exceptions import ObjectDoesNotExist
 
@@ -132,8 +131,8 @@ class RegistryPathField(serializers.CharField):
         """
         Converts a base_path into a registry path.
         """
-        host = settings.CONTENT_ORIGIN
-        return "".join([host.split("//")[-1], "/", value])
+        request = self.context["request"]
+        return f"{request.get_host()}/{value}"
 
 
 class ContainerNamespaceSerializer(ModelSerializer, _GetOrCreateMixin):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pulpcore>=3.7
+pulpcore>=3.9.0.dev
 ecdsa~=0.13.2
 pyjwkest~=1.4.0
 pyjwt[crypto]~=1.7.1
--- a/CHANGES/7385.bugfix
+++ b/CHANGES/7385.bugfix
@@ -1,0 +1,1 @@
+Fixed the value of registry_path in a container distribution.
--- a/pulp_container/app/token_verification.py
+++ b/pulp_container/app/token_verification.py
@@ -1,4 +1,3 @@
-import re
 import jwt
 
 from django.conf import settings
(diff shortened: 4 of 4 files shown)


lubosmj:
The tests are failing because one part of the issue was addressed in pulpcore. Refer to https://github.com/pulp/pulpcore/pull/968 to learn more.

mdellweg:
Depends on https://github.com/pulp/pulpcore/pull/968
Should we mark it as WIP until this has been merged?

pulp_container/app/serializers.py
```diff
@@ -109,8 +108,8 @@ def to_representation(self, value):
         """
         Converts a base_path into a registry path.
         """
-        host = settings.CONTENT_ORIGIN
-        return "".join([host.split("//")[-1], "/", value])
+        request = self.context["request"]
+        return f"{request.get_host()}/{value}"
```
ipanova, line 112: :+1:

ipanova:
changes look good to me, let's wait on the core pr to me merged.

pulp_container/app/serializers.py
dkliban: We need to provide two code paths here so that the plugin can stay compatible with older versions of pulpcore that don't pass in this request object. We should file a task to remove the support for the old behavior once we no longer support pulpcore 3.8.

lubosmj:
Tests are now passing because the additional change was merge in pulpcore. I believe we should now continue with the @dkliban's proposal.

pulp_container/app/serializers.py
dkliban: After some more discussion on IRC, we decided that it does not make sense to have the two code paths because the bug would still be present until pulpcore 3.9 is released. The other option is to release the pulpcore change as part of 3.8.1. If that does not pose any problems for other plugins, we will merge this PR after 3.8.1 is released.

mdellweg: Isn' it the other way around, and this change should bump the minimal supported pulpcore version to 3.8.0-dev?
And therefore only be released after 3.8?

ipanova: @mdellweg that's correct, we would need to raise the pulpcore requirement after 3.8.1 gets released

mdellweg: So i think, we need to raise that lower bound to 3.9(.dev?) with this PR. So that attempts to install it from source with the wrong version of pulpcore will fail already.
(we can drop the `dev` in the requirement on release)

mdellweg:
In the container meeting we discussed, that this PR should bump the core requirement to 3.9.dev.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-pulpcore>=3.7
+pulpcore>=3.9.dev
```
lubosmj, line 1: @mdellweg, I bumped the requirement here. This is what you wanted to see?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-pulpcore>=3.7
+pulpcore>=3.9.0.dev
```
lubosmj, line 1: @mdellweg, I bumped the requirement here. This is what you wanted to see?

ipanova:
@lubosmj can you make sure ci is green?

lubosmj:
@ipanova, sorry, Travis CI is now green.

mdellweg (approved):

PR #171, opened 2020-10-29
Fixed Renderer to handle properly Manifest and Blob responses.

ipanova:
closes #7620
https://pulp.plan.io/issues/7620

### DIFF
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -330,7 +330,6 @@ class ContainerFirstStage(Stage):
             url=blob_url,
             relative_path=digest,
             remote=self.remote,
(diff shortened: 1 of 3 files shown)


pulp_container/app/tasks/sync_stages.py
```diff
@@ -330,7 +330,6 @@ def create_blob(self, man_dc, blob_data):
             url=blob_url,
             relative_path=digest,
             remote=self.remote,
-            extra_data={"headers": V2_ACCEPT_HEADERS},
```
ipanova, line 333: This line is not related to the fix, however sending manifest headers for the blob requests is wrong.

mdellweg (approved):

dkliban (approved):

PR #172, opened 2020-11-02
Fixed a bug where Artifacts were missing some checksums after `podman…

ipanova:
… push`.

closes #7774
https://pulp.plan.io/issues/7774

### DIFF
--- a/CHANGES/7774.bugfix
+++ b/CHANGES/7774.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where Artifacts were missing sha224 checksum after `podman push`.
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -452,6 +452,7 @@ class BlobUploads(ContainerRegistryApiMixin, ViewSet):
                     sha256=upload.sha256,
                     sha384=upload.sha384,
                     sha512=upload.sha512,
+                    sha224=upload.sha224,
                     size=upload.file.size,
                 )
                 artifact.save()


ipanova:
@dkliban @mdellweg  we need to provide a migration along this fix as well

mdellweg:
you mean a data migration, agreed. If we need to make this backportable, we could provide it as an admin command. Adding the missing checksums is at least idempotent.

mdellweg:
Can we create that artifact using its serializer, to avoid such problems in the future?

ipanova:
> you mean a data migration, agreed. If we need to make this backportable, we could provide it as an admin command. Adding the missing checksums is at least idempotent.

There is some discussion on this ticket. Apparently we are going to add a django command along with the plugin providing the data migration https://pulp.plan.io/issues/7696
Once https://pulp.plan.io/issues/7696 and https://pulp.plan.io/issues/7561 will be addressed we would solve the issue of having missing checksums.
Speaking of the serializer, i don't know whether it would be able to cover all the cases mentioned in the discussion, at least we would need to extend it with the additional logic for the allowed_checksums. The pain point is that we cannot enforce plugins writers to  use the serialzer.

pulp_container/app/migrations/0010_fix_missing_blobs_checksums.py
```diff
@@ -0,0 +1,42 @@
...
+
+import hashlib
+from django.db import migrations, transaction
+
+
+CHUNK_SIZE = 1024 * 1024  # 1 Mb
+
+
+def fix_missing_blobs_checksums(apps, schema_editor):
+    with transaction.atomic():
```
mdellweg, line 11: Why is this atomic?

mdellweg (commented):
I realized, this will have a hard time to run, because migrations are called with pulpcore-admin and pulpcore-admin will already refuse to start if any artifact is improperly saved.

ipanova:
> I realized, this will have a hard time to run, because migrations are called with pulpcore-admin and pulpcore-admin will already refuse to start if any artifact is improperly saved.

yes, i had same realization after i have written the code.. we will need to find a different solution to this

mdellweg (approved):

PR #176, opened 2020-11-13
Store uploaded chunks as separate files in the shared storage

lubosmj:
closes #7218


pulp_container/app/models.py
```diff
@@ -7,20 +7,18 @@
...
 from pulpcore.plugin.models import (
-    Artifact,
+    BaseModel,
     Content,
     ContentGuard,
-    BaseModel,
     Remote,
     Repository,
     RepositoryVersionDistribution,
+    Upload,
```
mdellweg, line 21: Maybe import that with `Upload as CoreUpload` to avoid confusion.

pulp_container/app/registry_api.py
```diff
@@ -128,20 +128,21 @@ class UploadResponse(Response):
...
             content_length (int): The value for the Content-Length header.
             request (rest_framework.request.Request): Request object not used by this
                 implementation of Response.
         """
         headers = {
             "Docker-Upload-UUID": upload.pk,
-            "Location": "/v2/{path}/blobs/uploads/{pk}".format(path=path, pk=upload.pk),
-            "Range": "0-{offset}".format(offset=upload.file.size),
+            "Location": f"/v2/{path}/blobs/uploads/{upload.pk}",
+            "Range": f"0-{offset}",
```
mdellweg, line 145: Shouldn't this be `f"{offset}-{upload.file.size}"`?

pulp_container/app/models.py
```diff
@@ -385,60 +383,13 @@ def generate_filename(instance, filename):
     return time.strftime(filename)
 
 
-class Upload(BaseModel):
+class Upload(Upload):
```
mdellweg, line 386: Nice. ;)
I'd like to avoid naming them the same within the same file however.

mdellweg (commented):
Not a full review yet.

pulp_container/app/registry_api.py
```diff
@@ -428,12 +428,26 @@ def partial_update(self, request, path, pk=None):
...
+
+            upload.save()
+
         return UploadResponse(
-            upload=upload, path=path, content_length=upload.file.size, request=request
+            upload=upload,
+            path=path,
+            offset=chunk.size,
+            content_length=chunk.size,
+            request=request,
```
mdellweg, line 450: I am more than confused by the names and uses of these parameters.

pulp_container/app/registry_api.py
```diff
@@ -442,18 +456,18 @@ def put(self, request, path, pk=None):
...
-                artifact = Artifact(
-                    file=upload.file.name,
-                    md5=upload.md5,
-                    sha1=upload.sha1,
-                    sha256=upload.sha256,
-                    sha384=upload.sha384,
-                    sha512=upload.sha512,
-                    size=upload.file.size,
-                )
+                artifact = Artifact.init_and_validate(file)
```
mdellweg, line 470: :+1:

pulp_container/app/registry_api.py
```diff
@@ -428,12 +428,26 @@ def partial_update(self, request, path, pk=None):
...
         return UploadResponse(
-            upload=upload, path=path, content_length=upload.file.size, request=request
+            upload=upload,
+            path=path,
+            offset=chunk.size,
+            content_length=chunk.size,
+            request=request,
         )
 
     def put(self, request, path, pk=None):
```
mdellweg, line 453: The comment in the next line seems to be wrong.

pulp_container/app/registry_api.py
mdellweg: or `f"{0}-{upload.size}"`?

lubosmj:
Thank you for your review, @mdellweg. Your confusion should be now resolved.

pulp_container/app/registry_api.py
```diff
@@ -413,48 +412,44 @@ def partial_update(self, request, path, pk=None):
...
-                    file=upload.file.name,
-                    md5=upload.md5,
-                    sha1=upload.sha1,
-                    sha256=upload.sha256,
-                    sha384=upload.sha384,
-                    sha512=upload.sha512,
-                    sha224=upload.sha224,
-                    size=upload.file.size,
-                )
+                artifact = Artifact.init_and_validate(uploaded_file)
```
mdellweg, line 452: :+1:

pulp_container/app/registry_api.py
```diff
@@ -413,48 +412,44 @@ def partial_update(self, request, path, pk=None):
...
+        """
+        Create a blob from uploaded chunks.
+        """
         _, repository = self.get_dr_push(request, path)
 
         digest = request.query_params["digest"]
         upload = models.Upload.objects.get(pk=pk, repository=repository)
+        chunks = UploadChunk.objects.filter(upload=upload).order_by("offset")
+
+        with NamedTemporaryFile("ab") as temp_file:
```
mdellweg, line 443: Is there a `assemble_and_commit` method on core's Upload we could use here, now that we inherit it?

mdellweg (approved):
LGTM

pulp_container/app/registry_api.py
lubosmj: Not at the moment. In pulpcore, we have the same code just for assembling uploaded chunks: https://github.com/pulp/pulpcore/blob/08d446ba1a72f1eaec8f437e15ff799ee6efbdc2/pulpcore/app/tasks/upload.py#L28-L34. I can implement only the method `assemble()` because the committing is not same in both scenarios. What do you think?

mdellweg: Maybe later. Let's get this done first.

PR #179, opened 2020-11-17
Add AccessPolicy for ContainerRepository

mdellweg:
fixes #7706
https://pulp.plan.io/issues/7706


mdellweg:
@bmbouter I'd like to have your :eyes: on this early in the process.

pulp_container/app/migrations/0010_rbac_for_repositories.py
```diff
@@ -0,0 +1,85 @@
...
+        "effect": "allow",
+        "condition": "has_model_or_obj_perms:container.change_containerrepository",
+    },
+    {
+        "action": ["sync"],
+        "principal": "authenticated",
+        "effect": "allow",
+        "condition": [
+            "has_model_or_obj_perms:container.modify_repo_content",
+            "has_remote_param_model_or_obj_perms:container.view_containerremote",
```
mdellweg, line 44: This probably means that we depend on implementing the policy for Remotes first.

pulp_container/app/models.py
```diff
@@ -208,9 +214,13 @@ class ContainerRepository(Repository):
     TYPE = "container"
     CONTENT_TYPES = [Blob, Manifest, Tag]
     PUSH_ENABLED = False
+    ACCESS_POLICY_VIEWSET_NAME = "repositories/container/container"
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+        permissions = [
+            ("sync_containerrepository", "Can start a sync task"),
+        ]
```
mdellweg, line 223: WIP: There is definitely some permissions missing here.

pulp_container/app/viewsets.py
```diff
@@ -201,6 +201,71 @@ class ContainerRepositoryViewSet(RepositoryViewSet):
...
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:container.change_containerrepository",
+            },
+            {
+                "action": ["sync"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:container.sync_containerrepository",
+                    "has_remote_param_model_or_obj_perms:container.view_containerremote",
```
mdellweg, line 243: Is this line needed, or can it be handled by the permission scoped lookup?

pulp_container/app/viewsets.py
```diff
@@ -201,6 +201,71 @@ class ContainerRepositoryViewSet(RepositoryViewSet):
...
+                ],
+            },
+            {
+                "action": ["add", "remove", "tag", "untag", "copy_tags", "copy_manifests"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:container.modify_content_containerrepository",
```
mdellweg, line 251: Do we need to define permissions for the added content, too?

pulp_container/app/viewsets.py
ipanova: i think combination of these 2 conditions are correct

pulp_container/app/viewsets.py
ipanova: i don't understand how that would work? in this plugin we do not have the modify action

ipanova: Should it be this way for every action?
```
{
                "action": ["tag"],
                "principal": "authenticated",
                "effect": "allow",
                "condition": "has_model_or_obj_perms:container.tag_containerrepository",
            },
```

ipanova: theoretically all of the actions on the repo sync/tag/untag/add/remove/copy/ qualify you to view its content

ipanova: build_image is another action of the repo we need to register

mdellweg: That is exactly the question here: We need to define any non-statandard (crud) permission on the model class. [0]
Do we want a separate permission for each action on the viewset, or do we want to collect them into a single "modify_content" permission? Or should "tag" and "untag" be combined?

[0] https://github.com/pulp/pulp_container/pull/179/files#diff-1707426fbe1933a12da3490e42f0dbceea79e47f7c883708c0b0cb49755b1d94R219-R222

ipanova: @mdellweg I have read some more docs and i think having these actions grouped will suffice.

ipanova (approved):

PR #183, opened 2020-12-07
Add RBAC for container remotes.

ipanova:
closes #7707

### DIFF
--- a/CHANGES/7707.feature
+++ b/CHANGES/7707.feature
@@ -1,0 +1,1 @@
+Added access policy and permission management to the container remotes.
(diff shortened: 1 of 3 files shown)


mdellweg (commented):
This looks very good to me.
Do we have a blueprint how to test RBAC related changes?

ipanova:
> This looks very good to me.
> Do we have a blueprint how to test RBAC related changes?

When reviewing the core pr for rbac we used to test tasks. I would use same approach for testing other objects as well https://github.com/pulp/pulpcore/pull/815#issuecomment-670567070

mdellweg:
Sadly there is not a functional test example...

mdellweg (approved):

PR #190, opened 2020-12-16
Fixed tagging tests.

ipanova:
[noissue]


mdellweg (commented):
I do not see, how this is supposed to "fix" any test, as the third argument to the assert statements is only used as a message on a failed assertion.
But maybe that is what you wanted to fix, because the default output of `assertEqual` may be better.

ipanova:
@mdellweg this PR is not going to fix any failure because there are none, and this one i think was some sort of quirk since i cannot reproduce this locally plus as you can see ci is also green https://github.com/pulp/pulp_container/runs/1561323939?check_suite_focus=true#step:9:1147
by 'fix' i meant to properly use the assertEqual

mdellweg (approved):

PR #191, opened 2020-12-16
Add AccessPolicy for ContainerNamespace.

ipanova:
closes #7967

### DIFF
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -562,3 +562,43 @@ class ContainerNamespaceViewSet(
     endpoint_name = "pulp_container/namespaces"
     queryset = models.ContainerNamespace.objects.all()
     serializer_class = serializers.ContainerNamespaceSerializer
+    permission_classes = (AccessPolicyFromDB,)
+    queryset_filtering_required_permission = "container.view_containernamespace"
+
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["list"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": ["create"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_perms:container.add_containernamespace",
+            },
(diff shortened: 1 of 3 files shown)


pulp_container/app/viewsets.py
```diff
@@ -562,3 +562,43 @@ class ContainerNamespaceViewSet(
...
+                "condition": "has_model_or_obj_perms:container.delete_containernamespace",
+            },
+        ],
+        "permissions_assignment": [
+            {
+                "function": "add_for_object_creator",
+                "parameters": None,
+                "permissions": [
+                    "container.view_containernamespace",
+                    "container.delete_containernamespace",
```
mdellweg, line 600: I think, here we need the permission to manage Distributions in that namespace.
It will however only be checked in the actions statements of the Distribution access_policy.

ipanova: Yes, I plan to add it. I just have not manage to update the PR yet since we have discussed this on the meeting

ipanova:
@mdellweg i *think* this should be it.

pulp_container/app/viewsets.py
```diff
@@ -562,3 +562,44 @@ class ContainerNamespaceViewSet(
...
+                    "container.add_containerdistribution",
```
ipanova, line 601: now i am not sure, should we also add view/delete_distribution?

mdellweg: Good questions. I believe you get all the required permissions of the distributions you created for free. But you could be assigned to be the "manager of all nested distributions".

ipanova: we decided to manage distribution permissions auto-assignment in the follow up prs once we get policy for the distributions.

pulp_container/app/viewsets.py
mdellweg: We will add that later, because it's hard to code for something that does not exist.
Should be in place before it goes into master.

mdellweg (approved):

PR #192, opened 2020-12-19
Add access policy for ContainerDistributionViewSet

lubosmj:
ref #7937

### DIFF
--- a/pulp_container/app/access_policy.py
+++ b/pulp_container/app/access_policy.py
@@ -1,0 +1,30 @@
+from pulpcore.plugin.access_policy import AccessPolicyFromDB
+
+from pulp_container.app import models
+
+
+class DistributionAccessPolicyFromDB(AccessPolicyFromDB):
+    """
+    Access policy for DistributionViewSet which handles namespace permissions.
+    """
+
+    def has_add_namespace_perms(self, request, view, action):
+        """
+        Check whether a user can create a namespace or can read the namespace if it already exists.
+        """
+        namespace = request.data["base_path"].split("/")[0]
+        try:
+            namespace = models.ContainerNamespace.objects.get(name=namespace)
+        except models.ContainerNamespace.DoesNotExist:
+            return request.user.has_perm("container.add_containernamespace")
+        else:
+            return request.user.has_perm(
+                "container.view_containernamespace"
+            ) or request.user.has_perm("container.view_containernamespace", namespace)
+
+    def has_view_namespace_perms(self, request, view, action):
+        """
+        Test if a user is able read the namespace contained within a distribution.
+        """
+        namespace = view.get_object().namespace
+        return request.user.has_perm("container.view_containernamespace", namespace)
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -34,7 +34,7 @@ from pulpcore.plugin.viewsets import (
     OperationPostponedResponse,
 )
 
-from pulp_container.app import models, serializers, tasks
+from pulp_container.app import access_policy, models, serializers, tasks
 
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


pulp_container/app/viewsets.py
```diff
@@ -586,6 +586,49 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
...
+        ],
+        "permissions_assignment": [
+            {
+                "function": "add_for_object_creator",
+                "parameters": None,
+                "permissions": [
+                    "container.view_containerdistribution",
+                    "container.delete_containerdistribution",
+                    "container.change_containerdistribution",
+                    "container.add_containernamespace",
```
ipanova, line 626: i think this should not be present among the automatic permission assignment but rather put among the conditions for distribution creation i.e. I can create a distribution if i have model permission and I object permission on existing namespace and I have model permission in case there is not such namespace available.

pulp_container/app/viewsets.py
```diff
@@ -586,6 +586,49 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
...
+                "effect": "allow",
+            },
+            {
+                "action": ["create"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_perms:container.add_contrainerdistribution",
+            },
+            {
+                "action": ["retrieve"],
```
ipanova, line 606: please do not forget to add update/ partial_update actions

lubosmj:
As you can see, I just copied the whole `global_access_conditions.py` file and appended some custom functions to it. Maybe it is possible to inherit from `pulpcore.plugin.models.AccessPolicy` and add our own class to the field `permission_classes` defined within `ContainerDistributionViewSet`, like so:
```python
class DistributionViewSetAccessPolicy(AccessPolicy):
    statements = [
        {
            "action": ["list"],
            "principal": "authenticated",
            "effect": "allow",
        },
        {
            "action": ["create"],
            "principal": "authenticated",
            "effect": "allow",
            "condition": [
                "has_model_perms:container.add_containerdistribution",
                "has_add_namespace_perms",
            ],
        },
        {
            "action": ["retrieve"],
            "principal": "authenticated",
            "effect": "allow",
            "condition": [
                "has_view_namespace_perms",
            ],
        },
        {
            "action": ["update", "partial_update"],
            "principal": "authenticated",
            "effect": "allow",
            "condition": [
                "has_model_or_obj_perms:container.change_containerdistribution",
                "has_view_namespace_perms",
            ],
        },
        {
            "action": ["destroy"],
            "principal": "authenticated",
            "effect": "allow",
            "condition": [
                "has_model_or_obj_perms:container.delete_containerdistribution",
                "has_view_namespace_perms",
            ],
        },

    ]
    permissions_assignment = [
        {
            "function": "add_for_object_creator",
            "parameters": None,
            "permissions": [
                "container.view_containerdistribution",
                "container.delete_containerdistribution",
                "container.change_containerdistribution",
            ],
        },
    ]

    def has_add_namespace_perms(self, request, view, action):
        """
        Check whether a user can create a namespace or can read the namespace if it already exists.
        """
        namespace = request.data["base_path"].split("/")[0]
        try:
            namespace = models.ContainerNamespace.objects.get(name=namespace)
        except models.ContainerNamespace.DoesNotExist:
            return request.user.has_perm("container.add_containernamespace")
        else:
            return request.user.has_perm("container.view_containernamespace", namespace)


    def has_view_namespace_perms(self, request, view, action):
        """
        Test if a user is able read the namespace contained within a distribution.
        """
        namespace = view.get_object().namespace
        return request.user.has_perm("container.view_containernamespace", namespace)

...

class ContainerDistributionViewSet(BaseDistributionViewSet):

    ...

    permission_classes = (DistributionViewSetAccessPolicy, AccessPolicyFromDB)
```

However, I was not able to make it work.

ipanova:
@lubosmj you were on the right track, you should have subclassed from AccessPolicyFromDB  https://github.com/pulp/pulp_container/compare/rbac...ipanova:dist-rbac?expand=1

pulp_container/app/global_access_conditions.py
```diff
@@ -0,0 +1,221 @@
...
+        return True
+    if has_remote_param_obj_perms(request, view, action, permission):
+        return True
+    return False
+
+
+# Custom checks for namespaces within distributions
+
+
+def has_add_namespace_perms(request, view, action):
```
ipanova, line 203: at the next step we need to figure out how to enable this situation
```
$ http POST  :24817/pulp/api/v3/distributions/container/container/ base_path=library/production name=distribute-rhel7 --auth ina:goodpassword
HTTP/1.1 202 Accepted
Connection: close
Content-Length: 67
Correlation-ID: 878cf1b9197345e3a1240f01f3c671f7
Date: Tue, 22 Dec 2020 12:12:41 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "task": "/pulp/api/v3/tasks/d0357850-40e7-4d56-8ad0-dbc050374c50/"
}

(pulp) [vagrant@pulp3-source-fedora32 _scripts]$ http POST  :24817/pulp/api/v3/distributions/container/container/ base_path=library/dev name=dev-rhel7 --auth lena:badpassword
HTTP/1.1 403 Forbidden
Connection: close
Content-Length: 63
Correlation-ID: 420b9be882a847448c57bfc5431557e4
Date: Tue, 22 Dec 2020 12:13:40 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "detail": "You do not have permission to perform this action."
}

```
Both users have add namespace and add distribution perms assigned, however due to queryset scoping thay can perform operations only on the objects they have created.

lubosmj: With the current changes, your scenario is handled correctly. As I mentioned in our meeting, we should consider both the model-level permissions along with object-level permission at the same time. In particular, this had to be modified:
```diff
- return request.user.has_perm("container.view_containernamespace", namespace)
+ return request.user.has_perm(
+     "container.view_containernamespace"
+ ) or request.user.has_perm("container.view_containernamespace", namespace)
```

pulp_container/app/viewsets.py
```diff
@@ -586,6 +586,65 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
...
+                    "has_view_namespace_perms",
+                ],
+            },
+            {
+                "action": ["update", "partial_update"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:container.change_containerdistribution",
+                    "has_view_namespace_perms",
```
ipanova, line 623: this is arguable since if i have a view permission on the namespace it does not really mean i can change objects under it.
how about creating a 'has_manipulate_namespace_perms. That would give you permission to update/delete existing distributions under existing namespaces

pulp_container/app/access_policy.py
```diff
@@ -0,0 +1,30 @@
...
+        """
+        Check whether a user can create a namespace or can read the namespace if it already exists.
+        """
+        namespace = request.data["base_path"].split("/")[0]
+        try:
+            namespace = models.ContainerNamespace.objects.get(name=namespace)
+        except models.ContainerNamespace.DoesNotExist:
+            return request.user.has_perm("container.add_containernamespace")
+        else:
+            return request.user.has_perm(
```
ipanova, line 21: i guess i would then replace this with 'container.manipulate_distributions_namespace' This permission will be define on the namespace

pulp_container/app/viewsets.py
ipanova: https://github.com/pulp/pulp_container/pull/191#discussion_r544414661

ipanova:
This looks good, ty for working on this. Let's merge this as a start and add further necessary cheks later :clinking_glasses:

PR #194, opened 2020-12-21
Add RBAC tests for remotes

mdellweg:
This is not yet really testing much rbac related.
But it shows how to overcome the current inability to create users and assign permissions.

### DIFF
--- a/pulp_container/tests/functional/api/test_rbac_remotes.py
+++ b/pulp_container/tests/functional/api/test_rbac_remotes.py
@@ -1,0 +1,169 @@
+# coding=utf-8
+"""Tests that container remotes have RBAC."""
+from random import choice
+import unittest
+
+from pulp_smash import cli, config, utils
+from pulp_smash.pulp3.bindings import monitor_task
+from pulp_smash.pulp3.constants import ON_DEMAND_DOWNLOAD_POLICIES
+
+from pulp_container.tests.functional.utils import (
+    gen_container_remote,
+    skip_if,
+)
+
+from pulpcore.client.pulp_container import (
+    ApiClient as ContainerApiClient,
+    RemotesContainerApi,
+)
+from pulpcore.client.pulp_container.exceptions import ApiException
+
+
+CREATE_USER_CMD = [
+    "from django.contrib.auth import get_user_model",
+    "from django.contrib.auth.models import Permission",
+    "",
+    "user = get_user_model().objects.create(username='{username}')",
+    "user.set_password('{password}')",
+    "user.save()",
+    "for permission in {permissions!r}:",
+    "    if '.' in permission:",
+    "        app_label, codename = permission.split('.', maxsplit=1)",
+    "        perm = Permission.objects.get(codename=codename, content_type__app_label=app_label)",
+    "    else:",
+    "        perm = Permission.objects.get(codename=permission)",
+    "    user.user_permissions.add(perm)",
+]
+
+
+DELETE_USER_CMD = [
+    "from django.contrib.auth import get_user_model",
+    "get_user_model().objects.get(username='{username}').delete()",
+]
+
+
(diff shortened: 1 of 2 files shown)


lubosmj:
I am not reviewing this PR, but we will definitely benefit from it when it will be merged. Furthermore, we should also try to assign and remove particular object-level permissions for the user by utilizing https://django-guardian.readthedocs.io/en/stable/userguide/assign.html?highlight=assign_perm#assign-obj-perms to see how the code behaves.

ipanova:
This looks good for a start, adding a second user with different set of perms would be a good course of actions to continue on.

mdellweg:
I guess, adding object level permissions is a good thing we need to test, but we probably need a way to add those independently of the user creation process anyway. I'd leave that as an exercise for another PR.

pulp_container/tests/functional/api/test_rbac_remotes.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class RBACRemotesTestCase(unittest.TestCase):
+    """RBAC remotes."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Create class-wide variables and prepare api users."""
+        cls.user1 = gen_user(["container.add_containerremote"])
+        cls.user2 = gen_user(["container.view_containerremote"])
```
ipanova, line 27: does it makes sense to add one more test in this PR that checks user2 cannot create remotes?

pulp_container/tests/functional/api/test_rbac_remotes.py
```diff
@@ -0,0 +1,139 @@
...
+        # read without read permission
+        with self.assertRaises(ApiException):
+            remote = self.user3["remote_api"].read(self.remote.pulp_href)
+
+    @skip_if(bool, "remote", False)
+    def test_02_read_remotes(self):
+        """Read a remote by its name."""
+        page = self.user1["remote_api"].list(name=self.remote.name)
+        self.assertEqual(len(page.results), 1)
+        for key, val in self.remote.to_dict().items():
```
ipanova, line 78: how about skipping this? I seems like the main purpose is to confirm that the page has 1 result

pulp_container/tests/functional/api/test_rbac_remotes.py
```diff
@@ -0,0 +1,139 @@
...
+        body = gen_container_remote()
+        body["name"] = self.remote.name
+        with self.assertRaises(ApiException):
+            self.user1["remote_api"].create(body)
+
+    @skip_if(bool, "remote", False)
+    def test_02_read_remote(self):
+        """Read a remote by its href."""
+        remote = self.user1["remote_api"].read(self.remote.pulp_href)
+        for key, val in self.remote.to_dict().items():
```
ipanova, line 64: how about skipping this? It seems like the main purpose is to confirm that user can read a specific remote he looked for.

pulp_container/tests/functional/api/test_rbac_remotes.py
```diff
@@ -0,0 +1,139 @@
...
+        body = _gen_verbose_remote()
+        type(self).remote = self.user1["remote_api"].create(body)
+        for key in ("username", "password"):
+            del body[key]
+        for key, val in body.items():
+            with self.subTest(key=key):
+                self.assertEqual(self.remote.to_dict()[key], val, key)
+
+    @skip_if(bool, "remote", False)
+    def test_02_create_same_name(self):
```
ipanova, line 49: I am not entirely sure I see benefit in this exact test what would focus on the rbac testing itself, can you elaborate?

mdellweg: I must confess, i just took the tests from crud_remotes and added some rbac.
+1 to be more specific about what we test.

ipanova (approved):

PR #196, opened 2020-12-22
Add manage_namespace_distributions permissions.

ipanova:
[noissue]


pulp_container/app/access_policy.py
```diff
@@ -19,12 +20,12 @@ def has_add_namespace_perms(self, request, view, action):
             return request.user.has_perm("container.add_containernamespace")
         else:
             return request.user.has_perm(
-                "container.view_containernamespace"
-            ) or request.user.has_perm("container.view_containernamespace", namespace)
+                "container.manage_namespace_distributions"
+            ) or request.user.has_perm("container.manage_namespace_distributions", namespace)
 
     def has_view_namespace_perms(self, request, view, action):
```
ipanova, line 26: @lubosmj is there any benefir of keeping this check for perm?

lubosmj: @ipanova, I think we can get rid of that method.

Regarding the method `has_manage_namespace_distributions_perms()`, what do you think about providing an additional parameter `permission` that will be passed to `has_perm()`, like so: `request.user.has_perm(permission, namespace)`? When a user wants just to view a namespace, he or she does not need to have the `manage_namespace_distribution` permission, but only `view_containernamespace`.

pulp_container/app/access_policy.py
```diff
@@ -8,9 +8,10 @@ class DistributionAccessPolicyFromDB(AccessPolicyFromDB):
...
     """
 
-    def has_add_namespace_perms(self, request, view, action):
+    def has_manage_namespace_distributions_perms(self, request, view, action):
         """
-        Check whether a user can create a namespace or can read the namespace if it already exists.
+        Check whether a user can create a namespace or it can manage distributions
+        for existing namespace.
         """
         namespace = request.data["base_path"].split("/")[0]
```
lubosmj, line 16: The tests are failing because when you issue a request to retrieve a single distribution, you do not provide `base_path`. Maybe, it would be great if we could differentiate between the object creation and retrieval. Otherwise, just check whether `base_path` was provided and if not, use `view.get_object().namespace`.

pulp_container/app/access_policy.py
ipanova: I have left this method to decouple calls that require object-level perms on the namespace to perform update/patch/delete on existing distribution and model/object-level perms on the namespace when creating a new distribution.
Maybe we can improve this later when having a full picture of all other rbac objects

pulp_container/app/viewsets.py
```diff
@@ -611,7 +611,7 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
                 "effect": "allow",
                 "condition": [
                     "has_model_or_obj_perms:container.view_containerdistribution",
-                    "has_view_namespace_perms",
+                    # "has_namespace_obj_perms:container.view_containernamespace",
```
ipanova, line 614: I am not sure it is correct to require `has_namespace_obj_perms:container.view_containernamespace` if I already can view the distribution whether the one I have just created or I have a model perm to view all dists.

ipanova:
@mdellweg please review and feel free to merge and adjust as you see fit to move the work forward.

pulp_container/app/viewsets.py
mdellweg: You can add a second statement for "retrieve" that allows to view on the "manage_namespace_distributions" permission. Or we combine that kind of logic in a new `has_model_or_obj_or_namespace_perms`.
The latter cannot easily be adjusted by admins, however with the former it's much easier to screw up.

mdellweg (approved):
I'm sure this won't be the last version of it, but as we are collecting stuff on the rbac branch, i'd be ok with merging.

The hard part to grasp is to anticipate how an admin is supposed to build upon our default policy. and the `has_*_perms` methods are part of that particular interface.

PR #201, opened 2021-01-11
Add a note that explains the limitations of the schema conversion

lubosmj:
[noissue]


mdellweg:
Seems like you need to replace the api.json by a .gitkeep

docs/workflows/host.rst
```diff
@@ -105,3 +105,9 @@ Docker Output::
      https://docs.docker.com/get-started/
 
 
+.. note::
+    Having a container client which cannot handle requested manifests in the new format (schema 2),
```
mdellweg, line 109:
```suggestion
    When using a container client that cannot handle requested manifests in the new format (schema 2),
```

mdellweg (approved):

dkliban (approved):

PR #202, opened 2021-01-11
Consider RBAC during the token authentication

lubosmj:
ref #7937

### DIFF
--- a/pulp_container/app/authorization.py
+++ b/pulp_container/app/authorization.py
@@ -5,11 +5,16 @@ import uuid
 
 import jwt
 
+from collections import defaultdict
 from datetime import datetime
 
 from django.conf import settings
+
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
+
+from pulp_container.app.models import ContainerDistribution
+from pulp_container.app.access_policy import NamespacePermissionsChecker
 
 TOKEN_EXPIRATION_TIME = 300
 
@@ -22,17 +27,38 @@ class AuthorizationService:
     according to a user's scope.
     """
 
-    def generate_token(self, username, service, scope):
+    ANONYMOUS_USER = "AnonymousUser"
+    VALID_ACTIONS = ["pull", "push", "*"]
+
+    def __init__(self, user, service, scope):
+        """
+        Store class-wide variables and initialize a dictionary used for determining permissions.
+
+        Args:
+            user (django.contrib.auth.models.User): Requesting user.
+            service (str): Name of the service access is granted to.
+            scope (str): Scope of the resource that is to be accessed.
+
+        """
+        self.user = user
+        self.service = service
+        self.scope = scope
+
+        self.actions_permissions = defaultdict(
+            lambda: lambda *args: False,
+            {
+                "pull": self.has_pull_permissions,
+                "push": self.has_push_permissions,
+                "*": self.has_view_catalog_permissions,
+            },
+        )
+
+    def generate_token(self):
         """
         Generate a Bearer token.
 
         A signed JSON web token is generated in this method. The structure of the token is
         adjusted according the documentation https://docs.docker.com/registry/spec/auth/jwt/.
-
-        Args:
-            username (str): Requesting user.
-            service (str): Name of the service access is granted to.
-            scope (str): Scope of the resource that is to be accessed.
 
         Returns:
             dict: A newly generated Bearer token.
@@ -43,14 +69,14 @@ class AuthorizationService:
 
         current_datetime = datetime.now()
 
-        access = self.determine_access(username, scope)
+        access = self.determine_access()
         token_server = getattr(settings, "TOKEN_SERVER", "")
-        claim_set = self._generate_claim_set(
+        claim_set = self.generate_claim_set(
             access=[access],
-            audience=service,
+            audience=self.service,
             issued_at=int(current_datetime.timestamp()),
             issuer=token_server,
-            subject=username,
+            subject=self.user.username,
         )
 
         with open(settings.PRIVATE_KEY_PATH, "rb") as private_key:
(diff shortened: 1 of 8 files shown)


pulp_container/app/authorization.py
```diff
@@ -22,18 +29,43 @@ class AuthorizationService:
...
 
-    def generate_token(self, username, service, scope):
-        """
-        Generate a Bearer token.
+    ANONYMOUS_USER = "AnonymousUser"
+    VALID_ACTIONS = ["pull", "push", "*"]
 
-        A signed JSON web token is generated in this method. The structure of the token is
-        adjusted according the documentation https://docs.docker.com/registry/spec/auth/jwt/.
+    def __init__(self, username, service, scope):
```
ipanova, line 35: s/username/user

pulp_container/app/authorization.py
```diff
@@ -22,18 +29,43 @@ class AuthorizationService:
...
+        """
+        Store class-wide variables and initialize a dictionary used for determining permissions.
 
         Args:
             username (str): Requesting user.
```
ipanova, line 40: this needs update also

pulp_container/app/authorization.py
ipanova: how about we pass the user directly into the generate_token https://github.com/pulp/pulp_container/blob/rbac/pulp_container/app/registry_api.py#L327

pulp_container/app/authorization.py
```diff
@@ -92,32 +124,82 @@ def _split_into_encoded_groups(self, encoded_base32):
...
+        try:
+            distribution = ContainerDistribution.objects.get(base_path=path)
+        except ContainerDistribution.DoesNotExist:
+            return False
+        if self.user.has_perm("container.view_containerdistribution", distribution):
+            return True
+
+        return False
+
+    def has_push_permissions(self, path):
```
ipanova, line 165: does it make sense to consolidate this and has_manage_namespace_dist_perms into the codepath that is being used in a more generic way?

pulp_container/app/authorization.py
```diff
@@ -22,18 +27,39 @@ class AuthorizationService:
...
             service (str): Name of the service access is granted to.
             scope (str): Scope of the resource that is to be accessed.
 
+        """
+        self.user = user
+        self.service = service
+        self.scope = scope
+
+        self.actions_permissions = defaultdict(
+            lambda: lambda *args: False,
```
mdellweg, line 48: https://github.com/pulp/pulp_container/pull/203#discussion_r557505670

mdellweg: It took me a minute to understand the double lambda here. ;)

pulp_container/app/authorization.py
```diff
@@ -92,32 +118,70 @@ def _split_into_encoded_groups(self, encoded_base32):
...
-
-    def _generate_claim_set(self, issuer, issued_at, subject, audience, access):
+        typ, name, actions = self.scope.split(":")
+        actions = set(actions.split(","))
+
+        permitted_actions = set()
+        for action in actions:
+            has_permission = self.actions_permissions[action](name)
+            if has_permission:
+                permitted_actions.add(action)
```
mdellweg, line 139: https://github.com/pulp/pulp_container/pull/203#discussion_r557507443

you can do a set comprehension here i think:
(Suggested change in the link)

Also why make it a set in the first place when all you do is to turn into a list?

pulp_container/app/authorization.py
```diff
@@ -92,32 +118,70 @@ def _split_into_encoded_groups(self, encoded_base32):
...
+        Check if the user has permissions to pull from the repository specified by the path.
+        """
+        if self.user.has_perm("container.view_containerdistribution"):
+            return True
+
+        try:
+            distribution = ContainerDistribution.objects.get(base_path=path)
+        except ContainerDistribution.DoesNotExist:
+            return False
+        if self.user.has_perm("container.view_containerdistribution", distribution):
```
mdellweg, line 154: https://github.com/pulp/pulp_container/pull/203#discussion_r557510009
Maybe we should add a specific "pull" permission here, because "view_containerdistribution" includes reading the distributions properties.

pulp_container/app/authorization.py
```diff
@@ -92,32 +118,70 @@ def _split_into_encoded_groups(self, encoded_base32):
...
+            return True
+
+        return False
+
+    def has_push_permissions(self, path):
+        """
+        Check if the user has permissions to push to the repository specified by the path.
+        """
+        if not self.user.has_perm("container.add_containerdistribution"):
```
mdellweg, line 163: https://github.com/pulp/pulp_container/pull/203#discussion_r557511475
This seems to rule out access a bit too early. (Basically meaning, that you need both Class-level create permissions.)

pulp_container/app/authorization.py
```diff
@@ -92,32 +118,70 @@ def _split_into_encoded_groups(self, encoded_base32):
...
+        namespace = path.split("/")[0]
+        return NamespacePermissionsChecker.has_permissions(
+            namespace, self.user, "container.manage_namespace_distributions"
+        )
+
+    def has_view_catalog_permissions(self, path):
+        """
+        Check if the authenticated user is an administrator and is requesting the catalog endpoint.
+        """
+        if path == "catalog" and self.user.username == "admin":
```
mdellweg, line 175: https://github.com/pulp/pulp_container/pull/203#discussion_r557514184


maybe allow "self.user.is_staff()"
But should we not have a more open restriction here?
Should we do permission based queryset filtering?
@mdellweg

pulp_container/app/authorization.py
lubosmj: I thought it would be useful to use `set` here because we may receive a list of the same actions, i.e., `actions=pull,pull,push`. This way of permitting the actions should ignore duplicates.

pulp_container/app/authorization.py
lubosmj: The requirement was to allow only the administrator to perform such an action.

Regarding the permission based queryset filtering, I am not sure whether this is going to be resolved in a separate ticket here https://pulp.plan.io/issues/8068.

pulp_container/app/authorization.py
lubosmj: The comment you have referenced is highlighting different commands. What do you mean by the class-level create permissions? Do we have such a thing? We consider either model-level permissions or object-level permissions. Since the object does not exist yet, we need to check for this model-level permission, right?

pulp_container/app/authorization.py
mdellweg: right. I did not see this, but you can never be sure about user input. ;)

pulp_container/app/authorization.py
mdellweg: Yeah, i meant model-level permission. But yes, the repository might already exist, when you attempt to push to it.
The model-level permission is to be considered, if it did not exist before, or you do not have explicit object permissions.

pulp_container/app/authorization.py
mdellweg: :+1: other ticket seems fine.

ipanova: I wanted to check-in with AH first to understand what are their usecases. Docker hub seems to to deny access to this endpoint unless you have admin perms or maybe this is granted to specific users. I was not able to access this enpoint whether i was anonymous user or authed.
here is the follow up story https://pulp.plan.io/issues/8068 but it needs to be confirmed

pulp_container/app/authorization.py
ipanova: this seems  correct to me - if a user does not have model level perm for adding distributions that means he cannot push at all, regardless into which namespace.

pulp_container/app/authorization.py
ipanova: that is ok i think, why that would be a problem?

pulp_container/app/authorization.py
ipanova: @lubosmj this should not be a concern since we are controlling this in the auth header https://github.com/pulp/pulp_container/blob/master/pulp_container/app/token_verification.py#L100
But i see an improvement here to return `pull,push` https://github.com/pulp/pulp_container/blob/master/pulp_container/app/token_verification.py#L34

pulp_container/app/authorization.py
mdellweg: but he could have object level permissions to push to a distribution/push_repo. And we completely rule this out here.

ipanova: we are removing post for push-repos

ipanova: are you confusing this with namespace/distribution? i might not following you

mdellweg: given the name of this function, i believe it is questioning, whether a user can do a push operation on a specific path.
And whenever a distribution exists at that path, we should allow any user with object level "push_permission" (whatever that means in detail) for that distribution to push.
Only if there is not yet such a distribution, we need to ask if it's ok to create one.

ipanova: i see what you're saying. can we make any neede changes in the follow up PR? I would rather merge this as is to unblock folks

ipanova (approved):
There are some suggestions that I hope we can address in the follow up pr.

pulp_container/app/authorization.py
mdellweg: sure.

ipanova (approved):
I am merging this pr, in the follow up pr we need to (1)add a specific pull permission for distribution, (2) enable push for a user who has only object-level perms for existing distributions
(3) return pull,push https://github.com/pulp/pulp_container/blob/master/pulp_container/app/token_verification.py#L34

PR #203, opened 2021-01-13
Add RBAC to the push repository endpoint

goosemania:


pulp_container/app/models.py
```diff
@@ -250,6 +250,7 @@ class ContainerPushRepository(Repository):
     TYPE = "container-push"
     CONTENT_TYPES = [Blob, Manifest, Tag]
     PUSH_ENABLED = True
+    ACCESS_POLICY_VIEWSET_NAME = "repositories/container/push"
```
ipanova, line 253: repositories/container/container_push

mdellweg (commented):
Sorry, it seems like i forgot to submit this review yesterday.

pulp_container/app/viewsets.py
```diff
@@ -533,14 +533,44 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
...
-class ContainerPushRepositoryViewSet(ImmutableRepositoryViewSet):
+class ContainerPushRepositoryViewSet(ReadOnlyRepositoryViewSet):
     """
     ViewSet for a container push repository.
-    """
 
+    POST and DELETE are disallowed because a push repository is tightly coupled with a
+    ContainerDistribution which handles it automatically.
+    Created - during push operation, removed - with ContainerDistribution removal
+    """
```
mdellweg, line 543: Hey, i like this approach.
I think it solves https://pulp.plan.io/issues/8014 and https://pulp.plan.io/issues/7839

goosemania: @mdellweg , @ipanova , are those duplicates? They look like duplicates to me.
I'd close one as a dup, and open a new issue to cover the DELETE method.
What do you think?

goosemania: @mdellweg , @ipanova , are those duplicates? They look like duplicates to me.
I'd close one as a dup, and open a new issue to cover the DELETE method.
What do you think?

ipanova: @goosemania either way is fine. I would personally close #7839 as dup and update the title for #8014 that we are making push repos readonly - this will cover post and delete changes.

goosemania: That works as well.

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -4,52 +4,140 @@
...
+    def test_push_with_dist_perms(self):
+        """
+        Test that it's enough to have container distribution perms to perform push.
+
+        It also checks read abilities for users with different set of permissions.
+        """
+        repo_name = "test/perms"
+        local_url = "/".join([self.registry_name, f"{repo_name}:2.0"])
+        image_path = f"{DOCKERHUB_PULP_FIXTURE_1}:manifest_a"
+        self._push(image_path, local_url, self.user_creator)
```
goosemania, line 115: this give me an error at the moment
FWIW, I have only podman installed and it's 1.6.4.

```
self = <pulp_container.tests.functional.api.test_push_content.PushRepoTestCase testMethod=test_push_with_dist_perms>

    def test_push_with_dist_perms(self):
        """
        Test that it's enough to have container distribution perms to perform push.
    
        It also checks read abilities for users with different set of permissions.
        """
        repo_name = "test/perms"
        local_url = "/".join([self.registry_name, f"{repo_name}:2.0"])
        image_path = f"{DOCKERHUB_PULP_FIXTURE_1}:manifest_a"
>       self._push(image_path, local_url, self.user_creator)

pulp_container/tests/functional/api/test_push_content.py:115: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pulp_container/tests/functional/api/test_push_content.py:77: in _push
    cls.registry.push(local_url)
../pulp-smash/pulp_smash/cli.py:966: in _dispatch_command
    result = self._client.run(cmd, sudo=True)
../pulp-smash/pulp_smash/cli.py:316: in run
../pulp-smash/pulp_smash/cli.py:54: in code_handler
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = CompletedProcess(args=('sudo', 'podman', 'push', 'localhost:24817/test/perms:2.0'), returncode=125, stdout='', stderr=...r uploading manifest 2.0 to localhost:24817/test/perms: denied: Access to the requested resource is not authorized.\n')

    def check_returncode(self):
        """Raise an exception if ``returncode`` is non-zero.
    
        Raise :class:`pulp_smash.exceptions.CalledProcessError` if
        ``returncode`` is non-zero.
    
        Why not raise ``subprocess.CalledProcessError``? Because stdout and
        stderr are not included when str() is called on a CalledProcessError
        object. A typical message is::
    
            "Command '('ls', 'foo')' returned non-zero exit status 2"
    
        This information is valuable. One could still make
        ``subprocess.CalledProcessError`` work by overloading ``args``:
    
        >>> if isinstance(args, (str, bytes)):
        ...     custom_args = (args, stdout, stderr)
        ... else:
        ...     custom_args = tuple(args) + (stdout, stderr)
        >>> subprocess.CalledProcessError(args, returncode)
    
        But this seems like a hack.
    
        In addition, it's generally good for an application to raise expected
        exceptions from its own namespace, so as to better abstract away
        dependencies.
        """
        if self.returncode != 0:
            raise exceptions.CalledProcessError(
>               self.args, self.returncode, self.stdout, self.stderr
            )
E           pulp_smash.exceptions.CalledProcessError: Command ('sudo', 'podman', 'push', 'localhost:24817/test/perms:2.0') returned non-zero exit status 125.
E           
E           stdout: 
E           
E           stderr: Getting image source signatures
E           Copying blob sha256:ace0eda3e3be35a979cec764a3321b4c7d0b9e4bb3094d20d3ff6782961a8d54
E           Copying blob sha256:c86ba25590b0e721b9d5a2ca8d2d8e4cd8d694ad2a4235785eb8ac111e95725a
E           Copying config sha256:2ea2d398465fee77f3f80929d6ee391634e548a98eaab5cfb64dba5a25da13b8
E           Writing manifest to image destination
E           Error: Error copying image to the remote destination: Error writing manifest: Error uploading manifest 2.0 to localhost:24817/test/perms: denied: Access to the requested resource is not authorized.

../pulp-smash/pulp_smash/cli.py:129: CalledProcessError

```

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -4,52 +4,140 @@
...
+            except exceptions.CalledProcessError:
+                pass
+
+        cls.registry.pull(image_path)
+
+        if user:
+            cls.registry.logout(cls.registry_name)
+
+    @classmethod
+    def _push(cls, image_path, local_url, user):
```
mdellweg, line 68: This function is performing much more than a simple push atm. It makes reading the tests very confusing.

pulp_container/tests/functional/constants.py
```diff
@@ -70,3 +70,5 @@
 """
 
 DOCKERHUB_PULP_FIXTURE_1 = "pulp/test-fixture-1"
+
+LOCAL_REGISTRY_ADMIN = {'username': 'admin', 'password': 'password'}
```
mdellweg, line 74: I know it reflects the way it was, but shouldn't this password be retrieved from smash config?

pulp_container/tests/functional/api/test_push_content.py
goosemania: I can copy-paste it into every test or rename to tag and push.
It's worsens readability of the test significantly (for me personally) to have all those lines of code in every test just for the sake of making push while being logged in.

I'm also thinking that I might just drop all the tests when they work anyway, since the private/public repo work will change those entirely.

pulp_container/app/viewsets.py
```diff
@@ -100,6 +102,14 @@ class Meta:
         fields = BaseDistributionViewSet.filterset_class.Meta.fields
 
 
+class ContainerNamespaceFilter(BaseFilterSet):
```
goosemania, line 105: I had to add it to be able to search for namespaces to  clean them up in tests.
Let me know if it's worth a dedicated issue/log entry/pr and I can separate it from this never-ending-solve-all-the-other-issues-around PR.

pulp_container/tests/functional/api/test_token_authentication.py
```diff
@@ -121,6 +121,10 @@ def test_pull_image_with_real_container_client(self):
 
         image_url = urljoin(self.cfg.get_base_url(), self.distribution.base_path)
         image_with_tag = f"{image_url}:manifest_a"
+
+        registry_name = urlparse(self.cfg.get_base_url()).netloc
+        admin_user, admin_password = self.cfg.pulp_auth
+        registry.login("-u", admin_user, "-p", admin_password, registry_name)
```
goosemania, line 127: I don't know how this test worked before. 
I believe my changes didn't affect it. If you have any thoughts, please do share.

My wild guess is that it ran in some order that the user was logged into registry from previous tests.
Maybe because in my tests I explicitly logout, it started failing. :shrug:

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,145 @@
...
-        local_url = "/".join([registry_name, "foo/bar:1.0"])
-        # Be sure to not being logged in
-        try:
-            registry.logout(registry_name)
-        except exceptions.CalledProcessError:
-            pass
-        # Pull an image with large blobs
-        registry.pull("centos:7")
+    @classmethod
+    def tag_and_push(cls, image_path, local_url, user):
```
goosemania, line 77: @mdellweg , is it any better? or would you prefer to copy-paste these 5 lines of code + comments to every test?

pulp_container/app/viewsets.py
mdellweg: It's fine for me. We should be able to assume there are filters for every viewset...

pulp_container/tests/functional/api/test_push_content.py
mdellweg: Sorry, i was mistaken by the `registry.rmi` call. So to get it straight: This function tags a local image to the registry, logs in pushes the image and removes it from the _local_ storage. That seems like a nice primitive in tests.

I thought that it also removed it from the registry right away, and it didn't occur to me how we could test pulling it with another user afterwards. So maybe `_push` was a good name.

If you want to start testing that a push operation can be denied, you can wrap it in a `try` `finally` block.
(suggestions below)

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,145 @@
...
-            registry.push(local_url)
+        cls.registry.tag(image_path, local_url)
         # Log in
-        registry.login("-u", "admin", "-p", "password", registry_name)
+        cls.registry.login("-u", user["username"], "-p", user["password"], cls.registry_name)
         # Push successfully
-        registry.push(local_url)
-        # Pull while logged in
-        registry.pull(local_url)
+        cls.registry.push(local_url)
```
mdellweg, line 86:
```suggestion
        try:
            cls.registry.push(local_url)
```

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,145 @@
...
         # Untag local copy
-        registry.rmi(local_url)
-        # Pull while logged out
-        registry.pull(local_url)
-        # cleanup
-        distribution = distributions_api.list(name="foo/bar").results[0]
-        self.addCleanup(distributions_api.delete, distribution.pulp_href)
+        cls.registry.rmi(local_url)
+
+        cls.registry.logout(cls.registry_name)
```
mdellweg, line 90:
```suggestion
            cls.registry.rmi(local_url)

            cls.registry.logout(cls.registry_name)
```

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,145 @@
...
+        cls.user_admin = {"username": admin_user, "password": admin_password}
+        cls.user_creator = gen_user(
+            [
+                "container.add_containerdistribution",
+                "container.view_containerdistribution",
+                "container.add_containernamespace",
+            ]
+        )
+        cls.user_reader = gen_user(["container.view_containerpushrepository"])
+        cls.user_helpless = gen_user([])
```
ipanova, line 45: :sob:

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,145 @@
...
+    def tearDownClass(cls):
+        """Delete api users."""
+        del_user(cls.user_creator)
+        del_user(cls.user_reader)
+        del_user(cls.user_helpless)
+
+    @classmethod
+    def _pull(cls, image_path, user=None):
+        """
+        Pull using specified user.
```
ipanova, line 60: can you please extend this doc block and add that we also ensuring we login with user if specified and logout after every pull

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,145 @@
...
+        """Test that one can't push without being logged in."""
+        local_url = "/".join([self.registry_name, "foo/bar:1.0"])
+
+        # Try to push without permission
+        with self.assertRaises(exceptions.CalledProcessError):
+            self.registry.push(local_url)
+
+    def test_push_with_dist_perms(self):
+        """
+        Test that it's enough to have container distribution perms to perform push.
```
ipanova, line 114: this test shows that it is enough to have container distribution + namespace perms to push.
It covers the case when the namespace did not exist before push.

Can we add one more test what shows it is enough to have container dists perms + manage_dists_within_the_namespace when there is already an existing namespace? That would require to create a user_collaborator

ipanova: on the other hand, this push workflow has a lot of combinations, we can also add more extensive tests in the next PR pushes.

pulp_container/tests/functional/api/test_token_authentication.py
ipanova: before rbac anynoumos user could pull, now that we have rbac + we logout after every operation it is very possible that your wild guess is not that wild ;)

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,186 @@
...
+        """
+        repo_name = "team/owner"
+        local_url = "/".join([self.registry_name, f"{repo_name}:2.0"])
+        image_path = f"{DOCKERHUB_PULP_FIXTURE_1}:manifest_a"
+        self._push(image_path, local_url, self.user_creator)
+
+        collab_repo_name = "team/owner"
+        local_url = "/".join([self.registry_name, f"{collab_repo_name}:2.0"])
+        image_path = f"{DOCKERHUB_PULP_FIXTURE_1}:manifest_b"
+        self._push(image_path, local_url, self.user_collaborator)
```
ipanova, line 187: :+1:

ipanova (approved):
@goosemania thanks a lot for working on this, these tests so much greatly improve our position with regards to the correctness of the written code.

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -6,49 +6,186 @@
...
+        Ensure we login with a specified user and logout after the push.
+        A local tag is removed for cleanup purposes.
+        """
         # Tag it to registry under test
-        registry.tag("centos:7", local_url)
+        cls.registry.tag(image_path, local_url)
+        # Log in
+        cls.registry.login("-u", user["username"], "-p", user["password"], cls.registry_name)
+        try:
+            # Push successfully
```
mdellweg, line 104: This comment should be removed, IMHO. Because this function can just as well test an unsuccessful push operation.

pulp_container/tests/functional/api/test_push_content.py
mdellweg: I am wondering if the manage_namespace permission alone should do the trick here.
Sure it would be better to have it as a object permission. But that shouldn't block us to get this PR merged.

mdellweg (approved):

mdellweg:
A question that crossed my mind while thinking about #205 : With Push repos being read-only, how can i precreate one to mark it private?

ipanova:
> A question that crossed my mind while thinking about #205 : With Push repos being read-only, how can i precreate one to mark it private?

privacy of the registry repos will be managed with the flag set on the distribution. This will decide whether the registry repo will be visible from registry api for pull and push.
Push-repo will be auto created meaning that owner who created the distribution will be able to read it from pulp api. And all other users who subsequently manually would get added to the according groups.

That's my understanding.

mdellweg:
That is correct, the question i am unsure of is, whether a distribution that is created without a repo(-version) will automatically cocreate a push repo. That would mean, that you can no longer use an "empty" distribution as a placeholder to reserve the distribution-path.

PR #205, opened 2021-01-18
Add private flag to distribution

mdellweg:
This allows the user to specify its container repositories to be private
or world readable.

TODO:
- [x] Use the private flag to allow pull access

fixes #8102
https://pulp.plan.io/issues/8102


mdellweg:
The specified TODO depends on https://github.com/pulp/pulp_container/pull/203

ipanova:
> The specified TODO depends on #203

If  my thinking is correct these code changes should not depend on the push-repo at all, since all the logic is going through the distribution

mdellweg:
At some point in the Push code, the distribution will be looked up to check for the permission. That is the place where i expect this flag to introduce a shortcut.

I think, this is it:
https://github.com/pulp/pulp_container/pull/203/files#diff-c9c63122c0e9c033b596c104536853afe2505833c6d33f9ed1e06bf7322ef19fR160

mdellweg:
I will look into how to test this tomorrow.

mdellweg:
I realized, i want to base the tests for this on the ones in #203

mdellweg:
I am struggling with the api client again... But this is what i came up with as a test.

pulp_container/app/authorization.py
```diff
@@ -132,11 +132,7 @@ def determine_access(self):
         typ, name, actions = self.scope.split(":")
         actions = set(actions.split(","))
 
-        permitted_actions = set()
-        for action in actions:
-            has_permission = self.actions_permissions[action](name)
-            if has_permission:
-                permitted_actions.add(action)
+        permitted_actions = {action for action in actions if self.actions_permissions[action](name)}
```
mdellweg, line 135: I know, this is unrelated.

pulp_container/app/authorization.py
```diff
@@ -151,6 +147,8 @@ def has_pull_permissions(self, path):
             distribution = ContainerDistribution.objects.get(base_path=path)
         except ContainerDistribution.DoesNotExist:
             return False
+        if not distribution.private:
+            return True
```
mdellweg, line 151: This is the real meat in this PR.

pulp_container/app/registry_api.py
```diff
@@ -300,6 +300,15 @@ def get_dr_push(self, request, path, create=False):
...
                     raise RepositoryInvalid(name=path, message="Repository is read-only.")
+            elif create:
+                with transaction.atomic():
+                    repo_serializer = serializers.ContainerPushRepositorySerializer(
+                        data={"name": path}, context={"request": request}
+                    )
+                    repo_serializer.is_valid(raise_exception=True)
+                    repository = repo_serializer.create(repo_serializer.validated_data)
+                    distribution.repository = repository
+                    distribution.safe()
```
mdellweg, line 311: From IRC:
<x9c4> So you would create a private empty distro and then when pushing to it you decide it's fate to serve a push repo? Sounds fine to me.
<ipanova> yes, that would be my take on it

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -188,3 +189,43 @@ def test_push_to_existing_namespace(self):
...
+        self._push(image_path, local_url, self.user_creator)
+
+        self._pull(local_url, self.user_creator)
+        with self.assertRaises(exceptions.CalledProcessError):
+            self._pull(local_url, self.user_helpless)
+
+        distribution.private = False
+        distribution_response = self.user_creator["distribution_api"].update(
+            distribution.pulp_href, distribution
+        )
```
mdellweg, line 224: I think i am doing something wrong here.

pulp_container/app/authorization.py
ipanova: i believe this check should happen couple of lines above.
If distribution is not marked as private, there is no need to check on any permissions pull permissions https://github.com/pulp/pulp_container/pull/205/files#diff-c9c63122c0e9c033b596c104536853afe2505833c6d33f9ed1e06bf7322ef19fR143

pulp_container/app/models.py
```diff
@@ -371,6 +373,9 @@ class ContainerDistribution(
         related_name="container_distributions",
         null=True,
     )
+    private = models.BooleanField(
+        default=False, help_text=_("Prevent repository to be pulled by anonymous users.")
```
ipanova, line 377: we should phrase this somehow differently, if repo is private no only anonymous user cannot pull, but as well any authed user unless explicitly added by the owner.

pulp_container/app/serializers.py
```diff
@@ -265,25 +265,30 @@ def validate(self, data):
...
+                    )
+            else:
+                validated_data["namespace"] = ContainerNamespaceSerializer.get_or_create(
+                    {"name": namespace_name}
                 )
-        else:
-            validated_data["namespace"] = ContainerNamespaceSerializer.get_or_create(
-                {"name": namespace_name}
-            )
+        elif self.partial is False:
```
ipanova, line 282: what corner case are we covering here? base_path is a required field to my knowledge

pulp_container/app/registry_api.py
```diff
@@ -300,6 +300,15 @@ def get_dr_push(self, request, path, create=False):
                 repository = repository.cast()
                 if not repository.PUSH_ENABLED:
                     raise RepositoryInvalid(name=path, message="Repository is read-only.")
+            elif create:
```
ipanova, line 303: :+1:

pulp_container/app/viewsets.py
```diff
@@ -760,3 +760,37 @@ class ContainerNamespaceViewSet(
             },
         ],
     }
+
+    @extend_schema(
+        description="Trigger an asynchronous delete task",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def destroy(self, request, pk, **kwargs):
```
ipanova, line 768: we should leave this as a synchronous call. podman does not have a way how to deal with the async tasks, @dkliban PTAL

ipanova: we have also async distribution deletion. we need to brainstorm this more.

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -188,3 +189,44 @@ def test_push_to_existing_namespace(self):
         # cleanup, namespace removal also removes related distributions
         namespace = self.namespace_api.list(name="team").results[0]
         self.addCleanup(self.namespace_api.delete, namespace.pulp_href)
+
+    def test_private_repository(self):
```
ipanova, line 193: can you add one more test?
Test that you can create a private distribution and push to it.
Test that the same user_creator can pull, but user_helpless cannot.
Test the user_reader can pull form private repo.
Test that the other user_helpless  can pull after marking it non-private.

pulp_container/app/models.py
mdellweg:
```suggestion
        default=False, help_text=_("Restrict pull access to explicitly authorized users. Defaults to unrestricted pull access.")
```
Like that?

pulp_container/app/serializers.py
mdellweg: I tried to do a partial update to change "private" and it blew up, because "base_path" wasn't part of the `validated_data`.
This elif branch is maybe a safetynet. It should never be reached. I agree. Maybe i should raise an Implementation Error of some kind instead of a ValidationError to get a proper 500 server response.
WDYT?

pulp_container/app/viewsets.py
mdellweg: podman is never going to hit this, as it is part of the pulp api.

pulp_container/app/models.py
ipanova: :fire:

pulp_container/tests/functional/api/test_push_content.py
mdellweg: Apart from "Test the user_reader can pull form private repo." that sounds exactly like the test you are looking at. I can add this action to it, if you like ?!

pulp_container/app/serializers.py
ipanova: if it is a patch request, why do we enforce users to provide base_path if i want to only update the 'name', for example.
can we just leave the if branch on line 269 without the elif branch on line 282?

pulp_container/app/authorization.py
mdellweg: This is the first line after having a grasp on the distribution object. Do you mean to move that too?

pulp_container/tests/functional/api/test_push_content.py
ipanova: yes i meant adding only the  Test the user_reader can pull form private repo to add, I basically added it to the list of what you are already have.

pulp_container/tests/functional/api/test_push_content.py
mdellweg: I needed to fix the serializer for partial updates to make this work.

pulp_container/app/serializers.py
mdellweg: That's what i wanted to say if drf is implemented correctly, 282 should never be reached. I may have been to cautious. Removing it now...

pulp_container/app/authorization.py
mdellweg: I thought it was convenient to see this flag being evaluated as part of the question for pull permissions.
Being rather commonly set to True, can rearrange that a bit however.

mdellweg: In the long run, i see `has_pull_permission` being transformed into an access policy.

ipanova: i think even in the short run. AH plan to have their own policy, i moved this story to mvp milestone https://pulp.plan.io/issues/8075

pulp_container/app/viewsets.py
ipanova: @mdellweg I will be happy to merge this PR and address the async/sync/locking issue separately. I have filed an issue for it https://pulp.plan.io/issues/8151

ipanova (approved):
looks good to me,please make the ci green before merging

pulp_container/app/authorization.py
```diff
@@ -132,25 +132,24 @@ def determine_access(self):
...
+        permitted_actions = {action for action in actions if self.actions_permissions[action](name)}
 
         return {"type": typ, "name": name, "actions": list(permitted_actions)}
 
     def has_pull_permissions(self, path):
         """
         Check if the user has permissions to pull from the repository specified by the path.
         """
-        if self.user.has_perm("container.view_containerdistribution"):
+        if self.has_push_permissions(path):
```
ipanova, line 143: :+1:

mdellweg: Actually this will double the checking for push_permission if you issue `["push","pull"]`. I tried to get the  tests pass that way.

pulp_container/app/authorization.py
ipanova: seems like after rebase the code looks different. we should check distribution.private flag in the first place before checking for any perms.. i know this will be re-worked again in other prs but let's keep the logic in order in every pr, wdyt?

pulp_container/app/authorization.py
mdellweg: And, as expected, your last changes to the push pull permission logic fixed it for me.

mdellweg:
And after realizing, that the `user_reader` should not be able to `pull`, but the `user_dist_consumer` might, it should turn green now.

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -236,3 +237,48 @@ def test_push_to_existing_namespace(self):
...
+        with self.assertRaises(exceptions.CalledProcessError):
+            self._pull(local_url, self.user_helpless)
+
+        distribution.private = False
+        distribution_response = self.user_creator["distribution_api"].partial_update(
+            distribution.pulp_href, {"private": False}
+        )
+        monitor_task(distribution_response.task)
+
+        self._pull(local_url, self.user_reader)
```
ipanova, line 279: :+1:

PR #207, opened 2021-01-18
Adds granular permissions for push repositories     

dkliban:
The RegistryAccessPolicy is a combination of AccessPolicies for ContainerNamespace,
    ContainerDistribution, and ContainerPushRepository viewsets.
    
    The default policy requires users to have "container.add_containernamespace" and
    "container.add_containerdistribution" permissions in order to push a new repository
    into a new namespace. The permissions also enable user to use the Pulp API to create
    a namespace and distribution.
    
    The ContainerNamespaceViewset provides a create_namespace_group() function that can
    be used to assign permissions to a group of users. The default permission assignment
    for a namespace creates three groups Owners, Collaborators, and Consumers.
    
    Namespace Owners get the following permissions:
    
         "container.view_containernamespace",
         "container.delete_containernamespace",
         "container.namespace_add_containerdistribution",
         "container.namespace_delete_containerdistribution",
         "container.namespace_view_containerdistribution",
         "container.namespace_pull_containerdistribution",
         "container.namespace_push_containerdistribution",
         "container.namespace_change_containerdistribution",
         "container.namespace_view_containerpushrepository",
    
    Namespace Collaborators get the following permissions:
    
         "container.view_containernamespace",
         "container.namespace_add_containerdistribution",
         "container.namespace_delete_containerdistribution",
         "container.namespace_view_containerdistribution",
         "container.namespace_pull_containerdistribution",
         "container.namespace_push_containerdistribution",
         "container.namespace_change_containerdistribution",
         "container.namespace_view_containerpushrepository",
    
    Namespace Consumers get the following permissions:
    
         "container.view_containernamespace",
         "container.namespace_view_containerdistribution",
         "container.namespace_pull_containerdistribution",
         "container.namespace_view_containerpushrepository",
    
    The ContainerDistributionsViewset provides a create_distribution_group() function
    that can be used to assign permissions to a group of users. The default permission
    assignment for a newly created ContainerDistribution consists of three groups:
    Owners, Collaborators, and Consumers.
    
    ContainerDistribution Owners get the following permissions:
    
         "container.view_containerdistribution",
         "container.pull_containerdistribution",
         "container.push_containerdistribution",
         "container.delete_containerdistribution",
         "container.change_containerdistribution",
         "container.view_containerpushrepository",
    
    ContainerDistribution Collaborators get the following permissions:
    
         "container.view_containerdistribution",
         "container.pull_containerdistribution",
         "container.push_containerdistribution",
         "container.view_containerpushrepository",
    
    ContainerDistribution Consumers get the following permissions:
    
         "container.view_containerdistribution",
         "container.pull_containerdistribution",
         "container.view_containerpushrepository",
    
    The ContainerPushRepositoryViewset provides an add_perms_to_distribution_group()
    function to assign ContainerPushRepository permissions to the groups associated
    with the ContainerDistribution that serves the specific ContainerPushRepository.
    
    closes: #8101
    https://pulp.plan.io/issues/8101


pulp_container/app/models.py
```diff
@@ -186,7 +189,39 @@ class Meta:
         unique_together = (("name", "tagged_manifest"),)
 
 
-class ContainerNamespace(BaseModel, AutoAddObjPermsMixin, AutoDeleteObjPermsMixin):
+class CreateGroupAndAddPermissionsMixin(AutoAddObjPermsMixin):
```
goosemania, line 192: Nice!
Should we have the same for group removal? To ensure that no perms are inherited from the removed objects with the same field value?
It's critical for namespaces if we use its name and not pk.

ipanova: +1 it would be good to remove groups and perms whenever namespace/distribution also gets deleted.

dkliban: We should. I only considered the removal of the permissions which already occurs. However, the groups themselves need to be removed also. I'll add another mixin for that.

pulp_container/app/access_policy.py
```diff
@@ -17,8 +17,8 @@ def has_permissions(namespace, user, permission):
         try:
             namespace = models.ContainerNamespace.objects.get(name=namespace)
         except models.ContainerNamespace.DoesNotExist:
-            # check model permissions for namespace creation
-            return user.has_perm("container.add_containernamespace")
+            # Anyone can create a new namespace
+            return True
```
mdellweg, line 21: Na, we don't want that. If anyone should be able to create a new namespace, than the access_policy needs to reflect that.

pulp_container/app/models.py
```diff
@@ -188,7 +195,63 @@ class Meta:
...
+
+    To use this mixin, your model must support ``django-lifecycle``.
+    """
+
+    @hook("before_delete")
+    def delete_groups_and_user_obj_perms(self):
+        """Delete all groups associated with this object and user object permissions."""
+
+        group_pks = [o.group_id for o in GroupObjectPermission.objects.filter(object_pk=self.pk)]
+        Group.objects.filter(pk__in=group_pks).delete()
```
mdellweg, line 250: This will delete all groups with any permission on that object. Even the ones not created by us.

pulp_container/app/models.py
```diff
@@ -199,7 +262,12 @@ class ContainerNamespace(BaseModel, AutoAddObjPermsMixin, AutoDeleteObjPermsMixi
     class Meta:
         unique_together = (("name",),)
         permissions = [
-            ("manage_namespace_distributions", "Can manage distributions in a namespace"),
+            ("namespace_add_containerdistribution", "Add a distribution to a namespace"),
+            ("namespace_delete_containerdistribution", "Delete a distribution from a namespace"),
+            ("namespace_view_containerdistribution", "View a distribution in a namespace"),
+            ("namespace_pull_containerdistribution", "Pull from all distributions in a namespace"),
+            ("namespace_push_containerdistribution", "Push to all distributions in a namespace"),
+            ("namespace_change_containerdistribution", "Change all distributions in a namespace"),
```
mdellweg, line 270: :+1:

pulp_container/app/models.py
dkliban: I'll add better filtering for this

pulp_container/app/viewsets.py
```diff
@@ -587,6 +589,29 @@ class ContainerPushRepositoryViewSet(ReadOnlyRepositoryViewSet):
         ],
     }
 
+    def get_queryset(self):
```
dkliban, line 592: I had to add custom filtering for the repositories so that we could consider permissions on the namespace and the private field on the distribution. We will need to do something similar for the distributions viewset.

dkliban:
@ipanova @mdellweg @goosemania Even though there are still some test failures, the functionality we discussed is there. 

I discovered that during a push the hook for ContainerPushRepositories is called before the on create hook for ContainerDistributions so the logic in the group permission assignment is a bit weird. 

Please provide feedback on how to improve this.

pulp_container/app/access_policy.py
goosemania: Yeah, I'd appreciate some background behind the decision of allowing creation of namespaces by anyone.

pulp_container/app/models.py
goosemania: How about deleting only groups which are created automatically by us? all the default groups: owner, collaborator, consumer?
I'm not sure we want to touch groups created by the users, even if the only permission which is left is for the object which just got removed.
Imagine I have a group of users, teamA, and I give them only read access to a specific object. When this object is removed, they no longer have access to anything but that's fine, I don't want to recreate the group and add all the users again, I just want to give them new permissions eventually.

pulp_container/app/models.py
goosemania: unfortunately, it's the maximum recursive depth for the thumb-ups :disappointed:

pulp_container/app/models.py
```diff
@@ -239,7 +326,9 @@ def finalize_new_version(self, new_version):
         validate_repo_version(new_version)
 
 
-class ContainerPushRepository(Repository, AutoAddObjPermsMixin, AutoDeleteObjPermsMixin):
+class ContainerPushRepository(
+    Repository, CreateGroupAndAddPermissionsMixin, AutoDeleteObjPermsMixin
```
goosemania, line 330: We are not deleting groups here because all groups are related to the ContainerDistribution, is my understanding correct?

pulp_container/app/models.py
```diff
@@ -358,7 +447,7 @@ class Meta:
 
 
 class ContainerDistribution(
-    RepositoryVersionDistribution, AutoAddObjPermsMixin, AutoDeleteObjPermsMixin
+    RepositoryVersionDistribution, CreateGroupAndAddPermissionsMixin, AutoDeleteGroupsMixin
```
goosemania, line 450: Should we keep `AutoAddObjPermsMixin`, `AutoDeleteObjPermsMixin` everywhere as well?
So permission auto-assignment works for users who change the access policy.
And even without a change to access policy the permissions are cleaned properly from the groups created and assigned to by a user.
What do you think?

Can 2 hooks of the same type be performed? (2 after_delete and 2 after_create)?

ipanova:
> @ipanova @mdellweg @goosemania Even though there are still some test failures, the functionality we discussed is there.
> 
> I discovered that during a push the hook for ContainerPushRepositories is called before the on create hook for ContainerDistributions so the logic in the group permission assignment is a bit weird.
> 
> Please provide feedback on how to improve this.

We can try flip the order here https://github.com/pulp/pulp_container/blob/master/pulp_container/app/registry_api.py#L284 if that won't help we could try having nested transaction where we'd be able to control the order of execution

pulp_container/app/models.py
ipanova: I agree, let's auto remove only groups we have created. Rest of the groups would need to be deleted by the owner of the groups 'manually'

pulp_container/app/models.py
```diff
@@ -188,7 +195,81 @@ class Meta:
...
+    This mixin provides the ``create_group`` which creates groups with object level permissions.
+    """
+
+    def create_group(self, permissions, parameters):
+        """
+        Creates a group and optionally adds the current user to it.
+
+        The parameters are specified as a dictionary with the following keys:
+
+        "name_prefix" - the first part of the group name
```
ipanova, line 211: group_type is missing

pulp_container/app/authorization.py
```diff
@@ -174,15 +178,17 @@ def has_push_permissions(self, path):
...
-                return False
             namespace = path.split("/")[0]
             return NamespacePermissionsChecker.has_permissions(
-                namespace, self.user, "container.manage_namespace_distributions"
+                namespace, self.user, "container.namespace_add_containerdistribution"
             )
         # this is an existing distribution. User should have model or object perms
-        if self.user.has_perm("container.push_containerdistribution") or self.user.has_perm(
-            "container.push_containerdistribution", distribution
+        if (
```
ipanova, line 186: we should probably add model perms of `container.namespace_push_containerdistribution` to this logical statement

ipanova:
I believe we should not allow anyone to create a distribution via pulp api, we should add back ' add_distribution'
This should not affect 'push' operation since I think our code does not interact with that part of the code at all
https://github.com/pulp/pulp_container/blob/bb4d32d06da7b64069ed50c24cc7f6258a0817ab/pulp_container/app/viewsets.py#L696

pulp_container/app/authorization.py
```diff
@@ -151,23 +151,37 @@ def determine_access(self):
...
         """
         Check if the user has permissions to pull from the repository specified by the path.
         """
         try:
             distribution = ContainerDistribution.objects.get(base_path=path)
         except ContainerDistribution.DoesNotExist:
-            return False
+            namespace_name = path.split("/")[0]
+            try:
+                namesapce = ContainerNamespace.objects.get(name=namespace_name)
```
mdellweg, line 175:
```suggestion
                namespace = ContainerNamespace.objects.get(name=namespace_name)
```

pulp_container/app/authorization.py
```diff
@@ -176,16 +190,16 @@ def has_push_permissions(self, path):
...
-        view = namedtuple("FakeView", ["action", "get_object"])("push", lambda: distribution)
-        return self.access_policy.has_permission(request, view)
+            namespace_name = path.split("/")[0]
+            try:
+                namesapce = ContainerNamespace.objects.get(name=namespace_name)
+            except ContainerNamespace.DoesNotExist:
+                # Check if user is allowed to create a new namespace
+                return self.has_permission(None, "POST", "create", {"name": namespace_name})
+            # Check if user is allowed to create a new distribution in the namespace
+            return self.has_permission(namesapce, "GET", "create_distribution", {})
```
mdellweg, line 200:
```suggestion
            return self.has_permission(namespace, "POST", "create_distribution", {})
```

pulp_container/app/viewsets.py
```diff
@@ -642,33 +696,29 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
                 "action": ["create"],
                 "principal": "authenticated",
                 "effect": "allow",
-                "condition": [
-                    "has_model_perms:container.add_containerdistribution",
-                    "has_manage_namespace_dist_perms:container.manage_namespace_distributions",
-                ],
```
mdellweg, line 648: Having no condition here, means every authenticated user can create whatever not yet existing distribution (even in existing namespaces).
It should be
(namespace does not exist) or (user.has_namespace_perm(create_distribution))

pulp_container/app/access_policy.py
```diff
@@ -32,32 +32,40 @@ class DistributionAccessPolicyMixin:
...
+            return request.user.has_perm(permission, view.get_object())
 
     def obj_exists(self, request, view, action):
         """
         Check if the distribution exists.
         """
         return view.get_object() is not None
 
-    def is_private(self, request, view, action):
+    def is_public(self, request, view, action):
```
mdellweg, line 64: I think this is very confusing, that you need to set the `private` flag on the distribution to trigger this `is_public` check.

pulp_container/tests/unit/test_serializers.py
```diff
@@ -21,6 +24,14 @@ def setUp(self):
         self.push_repository_href = "/pulp/api/v3/repositories/container/container-push/{}/".format(
             self.push_repository.pk
         )
+        self.user = get_user_model().objects.create(username="user1", is_staff=False)
+        _set_current_user(self.user)
+
+    def tearDown(self):
+        """Delete the user."""
+        super(TestContainerDistributionSerializer, self).tearDown()
```
mdellweg, line 32:
```suggestion
        super().tearDown()
```
Welcome to world of python 3! ;)

pulp_container/tests/functional/utils.py
```diff
@@ -95,6 +100,30 @@ def del_user(user):
...
+
+
+def add_user_to_namespace_group(user, namespace_name, group_type, as_user):
+    """Add the user to either owner, collaborator, or consumer group of a namespace."""
+    namespace_collaborator_group = (
+        as_user["groups_api"]
+        .list(name="container.namespace.{}.{}".format(group_type, namespace_name))
+        .results[0]
+    )
+    as_user["group_users_api"].create(
```
mdellweg, line 121: This requires RBAC for the group endpoints, right?

ipanova:
> ```
> The RegistryAccessPolicy is a combination of AccessPolicies for ContainerNamespace,
> ContainerDistribution, and ContainerPushRepository viewsets.
> 
> The default policy requires users to have "container.add_containernamespace" and
> "container.add_containerdistribution" permissions in order to push a new repository
> into a new namespace. The permissions also enable user to use the Pulp API to create
> a namespace and distribution.
> 
> The ContainerNamespaceViewset provides a create_namespace_group() function that can
> be used to assign permissions to a group of users. The default permission assignment
> for a namespace creates three groups Owners, Collaborators, and Consumers.
> 
> Namespace Owners get the following permissions:
> 
>      "container.view_containernamespace",
>      "container.delete_containernamespace",
>      "container.namespace_add_containerdistribution",
>      "container.namespace_delete_containerdistribution",
>      "container.namespace_view_containerdistribution",
>      "container.namespace_pull_containerdistribution",
>      "container.namespace_push_containerdistribution",
>      "container.namespace_change_containerdistribution",
>      "container.namespace_view_containerpushrepository",
> 
> Namespace Collaborators get the following permissions:
> 
>      "container.view_containernamespace",
>      "container.namespace_add_containerdistribution",
>      "container.namespace_delete_containerdistribution",
>      "container.namespace_view_containerdistribution",
>      "container.namespace_pull_containerdistribution",
>      "container.namespace_push_containerdistribution",
>      "container.namespace_change_containerdistribution",
>      "container.namespace_view_containerpushrepository",
> 
> Namespace Consumers get the following permissions:
> 
>      "container.view_containernamespace",
>      "container.namespace_view_containerdistribution",
>      "container.namespace_pull_containerdistribution",
>      "container.namespace_view_containerpushrepository",
> 
> The ContainerDistributionsViewset provides a create_distribution_group() function
> that can be used to assign permissions to a group of users. The default permission
> assignment for a newly created ContainerDistribution consists of three groups:
> Owners, Collaborators, and Consumers.
> 
> ContainerDistribution Owners get the following permissions:
> 
>      "container.view_containerdistribution",
>      "container.pull_containerdistribution",
>      "container.push_containerdistribution",
>      "container.delete_containerdistribution",
>      "container.change_containerdistribution",
>      "container.view_containerpushrepository",
> 
> ContainerDistribution Collaborators get the following permissions:
> 
>      "container.view_containerdistribution",
>      "container.pull_containerdistribution",
>      "container.push_containerdistribution",
>      "container.view_containerpushrepository",
> 
> ContainerDistribution Consumers get the following permissions:
> 
>      "container.view_containerdistribution",
>      "container.pull_containerdistribution",
>      "container.view_containerpushrepository",
> 
> The ContainerPushRepositoryViewset provides an add_perms_to_distribution_group()
> function to assign ContainerPushRepository permissions to the groups associated
> with the ContainerDistribution that serves the specific ContainerPushRepository.
> 
> closes: #8101
> https://pulp.plan.io/issues/8101
> ```

Thank you for summarizing, this is very helpful! we should definitely capture this in our docs.
Couple of questions/suggestions:
Namespace Owners and Namespace Collaborators should also get `container.namespace_modify_content_containerpushrepository` otherwise they would not be able to tag/untag repos

Having `container.view_containerpushrepository` on the distribution viewset. how does that work? since this permission is defined on the push repository model and you can;t really re-use on another model

pulp_container/app/authorization.py
dkliban: thank you

pulp_container/app/viewsets.py
dkliban: I added the condition back

pulp_container/tests/functional/utils.py
dkliban: yes it does

pulp_container/tests/unit/test_serializers.py
dkliban: Thank you!

dkliban:
@ipanova The logic for adding `container.view_containerpushrepository` permission is a bit confusing in the code. I actually now think that I should add another method to the ContainerDistributionViewset that will accept ContainerPushRepository related permissions separately. Right now the `create_distribution_group()` method on the ContainerDistribution model checks if the permission name ends with `containerpushrepository` and separates those permissions into a list of their own. Then it checks if there is a ContainerPushRepository associated with the distribution. If there is, it adds the permissions for that object.

pulp_container/app/access_policy.py
```diff
@@ -6,46 +6,34 @@
...
-    """
-
-    def has_manage_namespace_dist_perms(self, request, view, action, permission):
-        """
-        Check whether a user can create a namespace or it can manage distributions
-        for existing namespace.
-        """
-        namespace = request.data["base_path"].split("/")[0]
-        return NamespacePermissionsChecker.has_permissions(namespace, request.user, permission)
+            distribution = models.ContainerDistribution.objects.get(repository=obj)
```
ipanova, line 24: Theoretically a repository can be distributed by mutiple distributions.
Another distribution can be created via pulp API and repo manually assigned to the distribution.
I suggest:
```
dists_qs = models.ContainerDistribution.objects.filter(repository=obj)
for dist in dists_qs:
    if request.user.has_perm(permission, dist.namespace):
        return True
```

pulp_container/app/access_policy.py
```diff
@@ -60,20 +48,21 @@ def is_private(self, request, view, action):
...
-    Access policy for DistributionViewSet which handles namespace permissions.
+    Access policy for ContainerDistributionViewSet and ContainerPushRepositoryViewSet which handles
+    namespace permissions.
     """
 
 
-class RegistryAccessPolicy(AccessPolicy, DistributionAccessPolicyMixin):
+class RegistryAccessPolicy(AccessPolicy, NamespaceAccessPolicyMixin):
     """
     An AccessPolicy that loads statements from the container distribution viewset.
```
ipanova, line 60: and namespace viewsets

pulp_container/app/authorization.py
```diff
@@ -176,16 +190,16 @@ def has_push_permissions(self, path):
...
-        view = namedtuple("FakeView", ["action", "get_object"])("push", lambda: distribution)
-        return self.access_policy.has_permission(request, view)
+            namespace_name = path.split("/")[0]
+            try:
+                namespace = ContainerNamespace.objects.get(name=namespace_name)
+            except ContainerNamespace.DoesNotExist:
+                # Check if user is allowed to create a new namespace
+                return self.has_permission(None, "POST", "create", {"name": namespace_name})
+            # Check if user is allowed to create a new distribution in the namespace
+            return self.has_permission(namespace, "GET", "create_distribution", {})
```
ipanova, line 200: s/GET/POST

pulp_container/app/migrations/0017_add_granular_perms.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('container', '0016_add_delete_versions_permission'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='containernamespace',
+            options={'permissions': [('namespace_add_containerdistribution', 'Add a distribution to a namespace'), ('namespace_delete_containerdistribution', 'Delete a distribution from a namespace'), ('namespace_view_containerdistribution', 'View a distribution in a namespace'), ('namespace_pull_containerdistribution', 'Pull from all distributions in a namespace'), ('namespace_push_containerdistribution', 'Push to all distributions in a namespace'), ('namespace_change_containerdistribution', 'Change all distributions in a namespace'), ('namespace_view_containerpushrepository', 'View all push repositories in a namespace'), ('namespace_modify_content_containerpushrepository', 'Modify content in all push repositories in a namespace')]},
```
ipanova, line 15: this is nitpick, but since we expose this to the user, can we add the distription so it has same pattern not to create confusion?
i.e Delete any distribution from  namespace
Pull from any distribution in a namespace..etc

pulp_container/app/models.py
```diff
@@ -188,18 +195,66 @@ class Meta:
...
+            name="{}.{}".format("container.namespace.owners", self.name)
+        )
+        assign_perm("auth.change_group", owners_group, group)
+        assign_perm("auth.view_group", owners_group, group)
+        if add_user_to_group:
+            current_user.groups.add(group)
+        self.add_for_groups(permissions, group.name)
+
+    @hook("before_delete")
+    def delete_groups_and_user_obj_perms(self):
```
ipanova, line 235: @dkliban I have added the AutoDeleteObjPermMixin and it seems like hook in the core is called fist, and then this one. Shall we add this mixin so we are sure we are also removing `GroupObjectPermission.`?

ipanova:
https://github.com/pulp/pulp_container/pull/207/files#diff-31c44b2812e50f3171248db027f439e8e64248cb84067fc7c305de74cdb9a86fR660 should this be `has_namespace_or_obj_perms`

pulp_container/app/viewsets.py
```diff
@@ -648,7 +650,7 @@ class ContainerPushRepositoryViewSet(TagOperationsMixin, ReadOnlyRepositoryViewS
                 "action": ["retrieve"],
                 "principal": "authenticated",
                 "effect": "allow",
-                "condition": "has_model_or_obj_perms:container.view_containerpushrepository",
+                "condition": "has_namespace_or_obj_perms:container.view_containerpushrepository",
             },
             {
                 "action": ["tag", "untag"],
```
ipanova, line 656: should this be `has_namespace_or_obj_perms:container.modify_content_containerpushrepository`

pulp_container/app/viewsets.py
```diff
@@ -661,8 +663,33 @@ class ContainerPushRepositoryViewSet(TagOperationsMixin, ReadOnlyRepositoryViewS
...
+                },
+                "permissions": [
+                    "container.view_containerpushrepository",
+                    "container.modify_content_containerpushrepository",
+                ],
+            },
+            {
+                "function": "add_perms_to_distribution_group",
+                "parameters": {
+                    "group_type": "consumers",
```
ipanova, line 690: consumers should not have `container.modify_content_containerpushrepository`

pulp_container/app/viewsets.py
ipanova: +1 not necessarily part of this pr, but commenting now for the sake of capturing a TODO

pulp_container/app/access_policy.py
dkliban: That makes sense. I'll make the change.

pulp_container/app/models.py
dkliban: Django lifecycle does not have any docs about decorating two methods with the same 'when' condition. I prefer to just add the         `GroupObjectPermission.objects.filter(object_pk=self.pk).delete()` to this method.

pulp_container/app/access_policy.py
```diff
@@ -21,8 +21,10 @@ def has_namespace_obj_perms(self, request, view, action, permission):
         if type(obj) == models.ContainerDistribution:
             namespace = obj.namespace
         else:
-            distribution = models.ContainerDistribution.objects.get(repository=obj)
-            namespace = distribution.namespace
+            dists_qs = models.ContainerDistribution.objects.filter(repository=obj)
```
ipanova, line 24:
```
def has_namespace_obj_perms(self, request, view, action, permission):
        """
        Check if a user has object-level perms on the namespace associated with the distribution
        or repository.
        """
        obj = view.get_object()
        if type(obj) == models.ContainerDistribution:
            namespace = obj.namespace
        else:
            dists_qs = models.ContainerDistribution.objects.filter(repository=obj)
            for dist in dists_qs:
                if request.user.has_perm(permission, dist.namespace):
                    return True
            return False

```

pulp_container/app/viewsets.py
ipanova: i filed an issue https://pulp.plan.io/issues/8206

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -184,26 +194,6 @@ def test_push_with_view_perms(self):
         with self.assertRaises(exceptions.CalledProcessError):
             self._push(image_path, local_url, self.user_reader)
 
-    def test_push_with_no_perms(self):
```
ipanova, line 187: why this test got removed?

ipanova (approved):

ipanova:
@dkliban this looks good to me! thank you forgetting this big chunk of work done!

PR #208, opened 2021-01-20
Add separate push and pull perms.

ipanova:
[noissue]


mdellweg:
Your commit message is a bit misleading ;)

pulp_container/app/authorization.py
```diff
@@ -133,6 +133,13 @@ def determine_access(self):
         actions = set(actions.split(","))
 
         permitted_actions = set()
+        if "push" in actions:
+            has_permission = self.actions_permissions["push"](name)
+            if has_permission:
+                permitted_actions.add("push")
+                permitted_actions.add("pull")
+
+            return {"type": typ, "name": name, "actions": list(permitted_actions)}
```
mdellweg, line 142: What's the reason for this codepath? In my experience podman will ask for push and pull which will be handled in the loop after this.

pulp_container/app/viewsets.py
```diff
@@ -634,7 +634,14 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
                 "effect": "allow",
                 "condition": [
                     "has_model_or_obj_perms:container.view_containerdistribution",
-                    # "has_namespace_obj_perms:container.view_containernamespace",
+                ],
+            },
+            {
+                "action": [],
```
mdellweg, line 640: I think there should be an action. Even if it is not defined on _that_ viewset.

pulp_container/app/authorization.py
ipanova: imagine you are pushing something for the first time. You don't have the distribution created just yet. This code path is needed to skip has_pull_permission which will fail unless the user has a global 'pull' permission..which does not really sound logical to give to a user ahead of push the global pull permission. This actually was the attempt to fix the but where in order to push we neded to give add_namespace, add_distro and view_distro perms https://github.com/pulp/pulp_container/pull/203/files#diff-18f14684e2e23997ce0169f5a491f334fee88119dfbe8587594359fe72c10bc0R40
In reality it should be enough to give add_namespace and  add_distro only

pulp_container/app/viewsets.py
ipanova: actually this was on purpose. If i will keep 'get' for example i will be able to still see the distribution endpoint with pull permissions. The goal was to separate view and pull perms. Meaning that if a user can pull content of a distribution it does not mean that he can do GET on pulp api distribution endpoint and see data.

ipanova: > I think there should be an action. Even if it is not defined on _that_ viewset.

i don't think so https://rsinger86.github.io/drf-access-policy/statement_elements/#action that;s why i left it empty

mdellweg: How is this policy then used anywhere?

ipanova:
These changes were made based on this convo
https://github.com/pulp/pulp_container/pull/202#pullrequestreview-570442775
https://github.com/pulp/pulp_container/pull/202#discussion_r558176347
https://github.com/pulp/pulp_container/pull/202#discussion_r558175229

pulp_container/app/viewsets.py
ipanova: we are checking on users perms during push/pull operations

mdellweg: I believe `user.has_perm` is already one level beneath the access policy (It is the technology used by the policies rules.).

pulp_container/app/viewsets.py
```diff
@@ -643,7 +642,6 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
                 "effect": "allow",
                 "condition": [
                     "has_model_or_obj_perms:container.change_containerdistribution",
-                    "has_namespace_obj_perms:container.manage_namespace_distributions",
```
mdellweg, line 646: why are we removing them here?

ipanova: I thought to leave manage_namescpace_distributions only when we add new ones or remove existing. For update operations it feels like it is enough to grant to the user this specific object-level perms. WDYT?

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -37,30 +37,39 @@ def setUpClass(cls):
...
+                "container.pull_containerdistribution",
+                "container.push_containerdistribution",
                 "container.view_containerpushrepository",
+                "container.manage_namespace_distributions",
             ]
         )
         cls.user_reader = gen_user(["container.view_containerpushrepository"])
         cls.user_helpless = gen_user([])
 
-        cls._pull("centos:7")
```
ipanova, line 55: any reason we were pulling such a big image?

mdellweg: I think that was because the test orinigally meant to be triggering multiple upload chunks.

mdellweg: We may restore the "old" test and repurpose this one for RBAC and use a tiny image.

pulp_container/app/viewsets.py
mdellweg: Yeah, it's hard. I thought we might need to replicate all the distribution permissions on the namespace level.
So when you are a consumer of that namespace, you can pull from all repos therein, including future ones.

pulp_container/app/models.py
```diff
@@ -393,6 +393,10 @@ def redirect_to_content_app(self, url):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+        permissions = [
+            ("pull_containerdistribution", "Can pull on a container distribution"),
+            ("push_containerdistribution", "Can push into a container distribution"),
```
goosemania, line 398: Since it's a user-facing  description, can we ask any of the native speakers to review this wording?
I'm not sure. "Pull on" sounds odd to my ear.
I though that one can "pull X from a registry/repo" or "push X to a registry/repo".
Is the container distribution here used as a synonym to a registry/repo?

pulp_container/app/authorization.py
goosemania, line 142: What happens if there is push and pull requested (aka in `actions`) but `has_permission` is `False` after checking the push perm?
It looks to me that we won't give even pull perms in such case, is it intentional? Or maybe I'm missing something.

ipanova: no, your observations are right, i need to rework logic a bit, ty

goosemania: It might be worth adding a test which covers that case.

pulp_container/app/viewsets.py
ipanova: i am hoping this will be managed by the created groups and assigned to those set of permissions.

mdellweg (approved):
Looks good so far. Let's merge to unblock all the followups.

PR #209, opened 2021-01-22
Return the digest of an unsigned manifest

lubosmj:
closes #7923

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -36,6 +36,7 @@ from pulp_container.app import models, serializers
 from pulp_container.app.authorization import AuthorizationService
 from pulp_container.app.redirects import FileStorageRedirects, S3StorageRedirects
 from pulp_container.app.token_verification import TokenAuthentication, TokenPermission
+from pulp_container.app.utils import get_accepted_media_types
 
 
 log = logging.getLogger(__name__)
@@ -546,6 +547,9 @@ class Manifests(RedirectsMixin, ContainerRegistryApiMixin, ViewSet):
             else:
                 manifest = models.Manifest.objects.get(digest=pk, pk__in=repository_version.content)
         except (models.Tag.DoesNotExist, models.Manifest.DoesNotExist):
+            raise ManifestNotFound(reference=pk)
+
+        if manifest.media_type not in get_accepted_media_types(request.headers):
             raise ManifestNotFound(reference=pk)
 
         return ManifestResponse(manifest, path, request)
--- a/pulp_container/app/schema_convert.py
+++ b/pulp_container/app/schema_convert.py
@@ -16,6 +16,7 @@ from pulp_container.constants import MEDIA_TYPE
 log = logging.getLogger(__name__)
 
 FS_Layer = namedtuple("FS_Layer", "layer_id uncompressed_digest history")
+ConversionResult = namedtuple("ConversionResult", "text digest content_type")
 
 
 class Schema2toSchema1ConverterWrapper:
@@ -29,16 +30,29 @@ class Schema2toSchema1ConverterWrapper:
         self.name = path
 
     def convert(self):
-        """Convert a manifest to schema 1."""
+        """Convert a manifest to schema 1.
+
+        Raises:
+            RuntimeError: If the conversion was not successful.
+
+        Returns:
+            ConversionResult: A converted manifest, corresponding digest, and content type.
+
+        """
         if self.tag.tagged_manifest.media_type == MEDIA_TYPE.MANIFEST_V2:
-            return self._convert_schema(self.tag.tagged_manifest)
+            schema_with_signature, digest = self._convert_schema(self.tag.tagged_manifest)
+            return ConversionResult(schema_with_signature, digest, MEDIA_TYPE.MANIFEST_V1_SIGNED)
         elif self.tag.tagged_manifest.media_type == MEDIA_TYPE.MANIFEST_LIST:
             legacy = self._get_legacy_manifest()
             if legacy.media_type in self.accepted_media_types:
                 # return legacy without conversion
-                return legacy, False, legacy.digest
+                legacy_schema = _jsonDumps(_get_manifest_dict(legacy))
+                return ConversionResult(legacy_schema, legacy.digest, legacy.media_type)
             elif legacy.media_type == MEDIA_TYPE.MANIFEST_V2:
-                return self._convert_schema(legacy)
+                schema_with_signature, digest = self._convert_schema(legacy)
+                return ConversionResult(
+                    schema_with_signature, digest, MEDIA_TYPE.MANIFEST_V1_SIGNED
+                )
             else:
                 raise RuntimeError()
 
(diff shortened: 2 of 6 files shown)


lubosmj:
I tested the changes with the function `calculate_digest` we use in pulp2 (https://github.com/pulp/pulp_docker/blob/cfe04bdb569a7b91a16ba01fd5e78079851de670/plugins/pulp_docker/plugins/models.py#L102):

```
In [1]: docker_content_digest = "sha256:58af68b09aba5ec25871122b589d3988193f3fc7bf0751eb0f10bc39b38e0449"

In [2]: with open("devel/backup/pulpunitmixin_latest_v1", "rb") as f:
   ...:     manifest = f.read()
   ...: 

In [3]: from pulp2_manifest_digest_checks import UnitMixin

In [4]: UnitMixin.calculate_digest(manifest) == docker_content_digest
Out[4]: True
```

lubosmj:
Furthermore, I was not able to reproduce the issue @ipanova experienced in https://pulp.plan.io/issues/7923#note-12. Due to that, the code which handles the redirects remains untouched.

pulp_container/app/registry_api.py
```diff
@@ -548,6 +549,9 @@ def head(self, request, path, pk=None):
         except (models.Tag.DoesNotExist, models.Manifest.DoesNotExist):
             raise ManifestNotFound(reference=pk)
 
+        if manifest.media_type not in get_accepted_media_types(request.headers):
+            raise ManifestNotFound(reference=pk)
```
mdellweg, line 553: :+1:

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -165,6 +167,17 @@ def test_api_performes_schema_conversion(self):
         base_content_type = content_response.headers["Content-Type"].split(";")[0]
         self.assertIn(base_content_type, {MEDIA_TYPE.MANIFEST_V1, MEDIA_TYPE.MANIFEST_V1_SIGNED})
 
+        header_digest = content_response.headers["Docker-Content-Digest"]
+        converted_manifest = json.loads(content_response.content)
+        converted_manifest.pop("signatures")
+        manifest_string = json.dumps(
+            converted_manifest, indent=3, sort_keys=True, separators=(",", ": ")
+        ).encode("utf-8")
```
mdellweg, line 175: Is this the approved way to write the manifest in order to produce the correct signature?

pulp_container/app/schema_convert.py
```diff
@@ -54,7 +52,14 @@ def _convert_schema(self, manifest):
             )
         except ValueError:
             raise RuntimeError()
-        return converter.convert()
+
+        converted_schema, schema_with_signature = converter.convert()
```
mdellweg, line 56: Doesn't convert return three values?
I guess i am confused by `Schema2toSchema1ConverterWrapper` vs `Schema2toSchema1Converter`.

mdellweg (commented):
Do i understand it correctly, that in the head call we do not perform the conversion and therefore do not return a digest?

pulp_container/app/schema_convert.py
lubosmj: No, it returns two values. See the method `convert` of the class `Schema2toSchema1Converter`. The method `_convert_schema` of the class `Schema2toSchema1ConverterWrapper` returns three values.

mdellweg: Can we simplify this in a follow up PR? It seems a bit unnecessary to write a wrapper around a class that isn't used elsewhere whilst both are under our control.
But, please don't mix such a simplification with a change in behavior, but make it two PR's.

pulp_container/tests/functional/api/test_pull_content.py
lubosmj: Well, I suppose, yes. Since we do the same operation in the reverse order when we are building a converted manifest: https://github.com/pulp/pulp_container/blob/070238111d9e6b297272f94ac1e8d9bc42c89c1b/pulp_container/app/schema_convert.py#L215-L222

https://github.com/pulp/pulp_container/blob/070238111d9e6b297272f94ac1e8d9bc42c89c1b/pulp_container/app/schema_convert.py#L203-L204

pulp_container/app/schema_convert.py
lubosmj: Will do.

ipanova:
> Do i understand it correctly, that in the head call we do not perform the conversion and therefore do not return a digest?

That's correct. HEAD request asks for existing manifest. Conversion of the manifest into some other media type creates a new manifest.

pulp_container/tests/functional/api/test_pull_content.py
mdellweg: Than we could use the same code in line 229 of schema_conversion.py instead of the rather hacky string manipulation.

lubosmj: No, because the signature is usually appended to the bottom of the payload. In the `dumps` call, there is used the argument `sort_keys=True`. This would change the order of the keys.

The base64 encoded field `formatLength` (encoded in the field `protected`) present the signature describes how many bytes correspond to the payload. Once we change the order, the clients could trim out the contents of the actual payload.

In the test, we can call the same command because we popped the signature from a dictionary-like object. Not all container clients will parse the content in a similar way in the future, right?

mdellweg: Ah, i see.

pulp_container/app/schema_convert.py
```diff
@@ -29,18 +30,29 @@ def __init__(self, tag, accepted_media_types, path):
...
-            return converted_schema, True, self.tag.tagged_manifest.digest
+            schema_with_signature, digest = self._convert_schema(self.tag.tagged_manifest)
+            return ConversionResult(schema_with_signature, digest, MEDIA_TYPE.MANIFEST_V1_SIGNED)
         elif self.tag.tagged_manifest.media_type == MEDIA_TYPE.MANIFEST_LIST:
             legacy = self._get_legacy_manifest()
             if legacy.media_type in self.accepted_media_types:
                 # return legacy without conversion
-                return legacy, False, legacy.digest
+                legacy_schema = _jsonDumps(_get_manifest_dict(legacy))
+                return ConversionResult(legacy_schema, legacy.digest, legacy.media_type)
```
mdellweg, line 50: So the boolean wasn't even used before?

mdellweg (approved):
This looks right to me.

pulp_container/app/schema_convert.py
lubosmj: It was used only in this place: https://github.com/pulp/pulp_container/blob/070238111d9e6b297272f94ac1e8d9bc42c89c1b/pulp_container/app/registry.py#L199-L200

Because of that, when we were trying to get, for example, manifest_v2 from manifest_list_v2, it returned an HTTP 404 error because the conversion was not performed. Still, it was a valid request and it should rather return the extracted am64 manifest.

ipanova (approved):

pulp_container/app/schema_convert.py
lubosmj: After a thorough examination, I am not going to do this. The wrapper class serves as a manager that executes the conversion only when it is required. Moving the part that determines the conversion to the class `Schema2toSchema1Converter` will violate the single responsibility principle. Furthermore, it will make the unit testing more complicated without additional monkey patching.

Maybe, we could just rename the class `Schema2toSchema1ConverterWrapper` to `Schema2toSchema1ConverterHandler` or `Schema2toSchema1ConverterManager`?

PR #211, opened 2021-01-25
Push pull ap

mdellweg:
~This is far from a working example~, but i was wondering if there's a way we can use the access_policy framework already in place to decide on the token permissions.
That way we could make the question whether a not-private repo is world readable or only available to logged in users part of the distribution access policy.

### DIFF
--- a/pulp_container/app/authorization.py
+++ b/pulp_container/app/authorization.py
@@ -5,16 +5,18 @@ import uuid
 
 import jwt
 
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from datetime import datetime
 
 from django.conf import settings
+from django.http import HttpRequest
+from rest_framework.request import Request
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 
 from pulp_container.app.models import ContainerDistribution
-from pulp_container.app.access_policy import NamespacePermissionsChecker
+from pulp_container.app.access_policy import RegistryAccessPolicy
 
 TOKEN_EXPIRATION_TIME = settings.get("TOKEN_EXPIRATION_TIME", 300)
 
@@ -43,6 +45,7 @@ class AuthorizationService:
         self.user = user
         self.service = service
         self.scope = scope
+        self.access_policy = RegistryAccessPolicy()
 
         self.actions_permissions = defaultdict(
             lambda: lambda *args: False,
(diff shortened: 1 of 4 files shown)


pulp_container/app/authorization.py
```diff
@@ -132,29 +136,21 @@ def determine_access(self):
...
-            return True
-
-        return False
+        # Fake the pull request
+        request = Request(HttpRequest())
+        request.method = "GET"
+        request.user = self.user
+        # TODO Fake the corresponding view
+        view = APIView()
+        return self.access_policy.has_permission(request, view)
```
bmbouter, line 153: I would keep going with this for now until we have a better solution. I've been working on extending rbac to not require NamedModelViewSet but it's not ready just yet. The issue I'm working on is: https://pulp.plan.io/issues/8048

mdellweg: It doesn't matter here, because we need the permissions for the registry endpoints in the token granting endpoint (totally weird).

mdellweg:
I'm sure the policy does not yet reflect, what we want as a default, but the permission code to move it into the policy might be ready for a review.

ipanova:
@mdellweg I like it. Will you be able to move catalog to the policy as well?

mdellweg:
> @mdellweg I like it. Will you be able to move catalog to the policy as well?

Yes sure. I can make it in this PR.

pulp_container/app/viewsets.py
```diff
@@ -667,11 +675,21 @@ class ContainerDistributionViewSet(BaseDistributionViewSet):
                 ],
             },
             {
-                "action": [],
+                "action": ["push"],
                 "principal": "authenticated",
                 "effect": "allow",
                 "condition": [
                     "has_model_or_obj_perms:container.push_containerdistribution",
+                    "obj_exists",
```
mdellweg, line 683: I think, i know now, what was missing.
:facepalm:

pulp_container/app/authorization.py
```diff
@@ -156,45 +159,51 @@ def has_pull_permissions(self, path):
...
-            )
-        # this is an existing distribution. User should have model or object perms
-        if self.user.has_perm("container.push_containerdistribution") or self.user.has_perm(
-            "container.push_containerdistribution", distribution
-        ):
-            return True
-        return False
+            distribution = None
+
+        # Fake the pull request
```
dkliban, line 181: s/pull/push/

dkliban (approved):

pulp_container/app/authorization.py
mdellweg: Let's only "Fake the request" then we don't introduce more copy and paste errors.

PR #212, opened 2021-01-25
Add RBAC to repository versions

goosemania:
closes #8017
https://pulp.plan.io/issues/8017


pulp_container/app/access_policy.py
```diff
@@ -43,3 +46,101 @@ def has_namespace_obj_perms(self, request, view, action, permission):
...
+
+        Returns:
+            True if the user has the Permission named by the ``permission`` at the model-level or on the
+                argument on the ``repository`` parameter at the object-level. False otherwise.
+        """
+        if self.has_model_perms(request, view, action, permission):
+            return True
+        if self.has_repoversion_repo_obj_perms(request, view, action, permission):
+            return True
+        return False
```
goosemania, line 113: @bmbouter , I wrote those in a generic way, should I add them to the global ones in pulpcore right away? pulp_container requires 3.10 anyway.

Shorter name which describes that it can be used for repo versions only and that repo perms needs to be provided are welcome.

pulp_container/app/viewsets.py
```diff
@@ -545,6 +545,33 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
...
+                    "container.view_containerrepository",
+            },
+            {
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition":
+                    "(has_repoversion_repo_model_or_obj_perms:container.delete_containerrepository "
+                    "or has_repoversion_repo_model_or_obj_perms:"
+                        "container.modify_content_containerrepository)",
```
goosemania, line 571: Any formatting suggestions?
black does this:
```
                "condition":
￼                "(has_repoversion_repo_model_or_obj_perms:container.delete_containerrepository "
￼                "or has_repoversion_repo_model_or_obj_perms:"
￼                "container.modify_content_containerrepository)",
```
or goes over 100 chars

pulp_container/app/viewsets.py
```diff
@@ -594,6 +621,30 @@ class ContainerPushRepositoryVersionViewSet(RepositoryVersionViewSet):
...
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition":
+                    "has_repoversion_repo_model_or_obj_perms:container.view_containerpushrepository"
+            },
+            {
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_pushrepoversion_delete_perms",
```
goosemania, line 644: I made this one non-customizable. 
If I got perms right and if we want to make this one more customizable, I can introduce lookups by distribution and by namespace and we'll have a looooong condition with ORs here.

Any opinions?

pulp_container/app/access_policy.py
```diff
@@ -43,3 +46,101 @@ def has_namespace_obj_perms(self, request, view, action, permission):
...
+        Args:
+            request (rest_framework.request.Request): The request being made.
+            view (subclass rest_framework.viewsets.GenericViewSet): The view being checked for
+                authorization.
+            action (str): The action being performed, e.g. "destroy".
+        """
+        sufficient_model_perms = (
+            "container.delete_containernamespace",
+            "container.manage_namespace_distributions",
+            "container.delete_containerdistribution",
```
goosemania, line 128: I check these model ones and then the same object ones. Did I get the right set of perms? Any of them is enough, right?

pulp_container/app/viewsets.py
goosemania: Alternatively, we can add a delete_containerpushrepo to the auto-assignment section of the push repo viewset and check it here.
I came up with this only now, while writing the comments.

pulp_container/app/access_policy.py
goosemania: Will move to pulpcore

pulp_container/app/access_policy.py
ipanova: maybe i am wrong, but should not we just require model or obj perms of the repo,same as with the regular container repo?
We just need to make sure that the created Groups during push operation contains correct set of permissions assigned to each group(owner, collaborator, consumer)

pulp_container/app/viewsets.py
ipanova: can we then split this into 2 rules?
```
            {
                "action": ["destroy"],
                "principal": "authenticated",
                "effect": "allow",
                "condition":
                    "has_repoversion_repo_model_or_obj_perms:container.delete_containerrepository "
            },
            {
                "action": ["destroy"],
                "principal": "authenticated",
                "effect": "allow",
                "condition":
                    " has_repoversion_repo_model_or_obj_perms:"
                        "container.modify_content_containerrepository)",
            },
```

This should be equivalent

pulp_container/app/access_policy.py
goosemania: See my other comment https://github.com/pulp/pulp_container/pull/212/files#r563954354
We currently do not use delete permissions on push repos at all. We can start using them but it's the reason I started checking distributions and namespaces.

pulp_container/app/viewsets.py
ipanova: adding 'delete_repo_versions' in the auto-assignment section of the container push repo would make sense to me.

ipanova: after thinking a bit more..i am not sure it is a good idea to allow to delete repo versions from push repo.
example: you perform podman push. At the end of the operation  you have 3 repo versions, because that image had 2 blobs and 1 manifest. If the user will remove last repo version it will make the repo unusable. Since the user does not know these subtle details of 'complete' incomplete versions i suggest we do not allow deleting versions at all.
Unless we can enable 'delete any version expect the last one" ( last version is always served of the container push repo)

goosemania: It make sense to me, I'll remove destroy section from here.

pulp_container/app/access_policy.py
```diff
@@ -43,3 +43,27 @@ def has_namespace_obj_perms(self, request, view, action, permission):
...
+        sufficient_repo_perms = (
+            "container.delete_containerrepository",
+            "container.sync_containerrepository",
+            "container.modify_content_containerrepository",
+            "container.build_image_containerrepository",
+        )
+        # any permission will do
+        for perm in sufficient_repo_perms:
+            if request.user.has_perm(perm) or request.user.has_perm(perm, repository):
+                return True
```
goosemania, line 67: One option is to check all permissions here and be able to tweak them without changing a default policy statements in the db. The downside is that a user won't see these details if they try to read the default policy via API.

pulp_container/app/viewsets.py
```diff
@@ -545,6 +545,57 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
...
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_repo_attr_model_or_obj_perms:container.view_containerrepository",
+            },
+            {
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_delete_containerrepository_versions_perms",
+            },
```
goosemania, line 568: This is still option one.

pulp_container/app/viewsets.py
```diff
@@ -545,6 +545,57 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
...
+            #     "condition":
+            #         "has_repo_attr_model_or_obj_perms:container.build_image_containerrepository",
+            # },
+            # {
+            #     "action": ["destroy"],
+            #     "principal": "authenticated",
+            #     "effect": "allow",
+            #     "condition":
+            #         "has_repo_attr_model_or_obj_perms:container.modify_content_containerrepository",
+            # },
```
goosemania, line 596: This is option two. 
A default policy can't be changed easily and there are many statements for the same action/principal/effect.
A good thing is that a user will see those perms in detail if they try to read the access policy via API.

goosemania: Option three is to put all conditions in one loooong string ad have one item for a  `destroy` action.

goosemania:
@ipanova , @dkliban , @mdellweg , please choose which option out of three is a preference for the container team and I will go with that.

In no particular order, the options (see details in my other comments) are:
1. custom check
2. many items for a destroy action with one condition
3. one item for a destroy action with a many conditions put in one string

pulp_container/app/access_policy.py
mdellweg: What about we call this helper `has_modify_containerrepository_version_perms` (i.e. rather what it is, and not what we intend to use it for in the default policy). The policy would then read "You can delete RV, if you can add RV by any means."
And still i'd like to give the admin the tools (a `container.delete_containerrepository_version` permission) to be able to say "Just being able to add content does not mean you can delete any repo version."

pulp_container/app/viewsets.py
mdellweg: Option 4:
```json
            {
                "action": ["destroy"],
                "principal": "authenticated",
                "effect": "allow",
                "condition":
                    "has_repo_attr_model_or_obj_perms:container.delete_containerrepository"
                    " or has_repo_attr_model_or_obj_perms:container.delete_version_containerrepository",
            },
```

pulp_container/app/viewsets.py
```diff
@@ -591,9 +642,31 @@ class ContainerPushRepositoryViewSet(ReadOnlyRepositoryViewSet):
...
+                "effect": "allow",
+            },
+            {
+                "action": ["retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_repo_attr_model_or_obj_perms:container.view_containerpushrepository",  # noqa
+            },
+        ],
+    }
```
mdellweg, line 669: :+1:

pulp_container/app/viewsets.py
ipanova: my understanding is, in case admin would like to change the policy, he will hit this endpoint`http GET  :24817/pulp/api/v3/access_policies` and be able to issue a PUT/PATCH request with changes. If we go with option1. How we would be able to do so? Unless we will change list of `sufficient_repo_permissions` which are hardcoded in the code?
```
$ http GET  :24817/pulp/api/v3/access_policies/92015dd5-a312-49f6-820a-58b91919cce9/
HTTP/1.1 200 OK
Allow: GET, PUT, PATCH, HEAD, OPTIONS
Connection: close
Content-Length: 2128
Correlation-ID: 188fd71293a146bc83dd0393f8490a02
Date: Thu, 28 Jan 2021 11:12:14 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "permissions_assignment": [
        {
            "function": "add_for_object_creator",
            "parameters": null,
            "permissions": [
                "container.view_containerrepository",
                "container.change_containerrepository",
                "container.delete_containerrepository",
                "container.sync_containerrepository",
                "container.modify_content_containerrepository",
                "container.build_image_containerrepository"
            ]
        }
    ],
    "pulp_created": "2021-01-27T19:15:01.424436Z",
    "pulp_href": "/pulp/api/v3/access_policies/92015dd5-a312-49f6-820a-58b91919cce9/",
    "statements": [
        {
            "action": [
                "list"
            ],
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "create"
            ],
            "condition": "has_model_perms:container.add_containerrepository",
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "retrieve"
            ],
            "condition": "has_model_or_obj_perms:container.view_containerrepository",
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "destroy"
            ],
            "condition": "has_model_or_obj_perms:container.delete_containerrepository",
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "update",
                "partial_update"
            ],
            "condition": "has_model_or_obj_perms:container.change_containerrepository",
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "sync"
            ],
            "condition": [
                "has_model_or_obj_perms:container.sync_containerrepository",
                "has_remote_param_model_or_obj_perms:container.view_containerremote"
            ],
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "remove",
                "tag",
                "untag"
            ],
            "condition": [
                "has_model_or_obj_perms:container.modify_content_containerrepository"
            ],
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "add"
            ],
            "condition": [
                "has_model_or_obj_perms:container.modify_content_containerrepository",
                "has_content_param_model_or_obj_perms"
            ],
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "copy_tags",
                "copy_manifests"
            ],
            "condition": [
                "has_model_or_obj_perms:container.modify_content_containerrepository",
                "has_source_param_model_or_obj_perms:container.view_containerrepository"
            ],
            "effect": "allow",
            "principal": "authenticated"
        },
        {
            "action": [
                "build_image"
            ],
            "condition": [
                "has_model_or_obj_perms:container.build_image_containerrepository"
            ],
            "effect": "allow",
            "principal": "authenticated"
        }
    ],
    "viewset_name": "repositories/container/container"
}

```

pulp_container/app/access_policy.py
ipanova: you have a point here - maybe this is not correct to enable a user to remove a repo version if he can sync the repo. Now i wonder whether we should allow this only if he can delete repo or maybe go with option 4

mdellweg: I mostly see this function as a convenience wrapper, as the policy is able to reflect the same "has_any_of_those_permissions" logic.
That being said, i'm not opposed to have such convenience.
That it allows us to change the policies meaning without even touching it, i'm undecided whether that is nice or problematic. But I could be calmed by giving it a descriptive name.

pulp_container/app/migrations/0013_add_delete_versions_perm.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('container', '0012_add_container_namespace_permissions'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='containerrepository',
+            options={'default_related_name': '%(app_label)s_%(model_name)s', 'permissions': [('sync_containerrepository', 'Can start a sync task'), ('modify_content_containerrepository', 'Can modify content in a repository'), ('build_image_containerrepository', 'Can use the image builder in a repository'), ('delete_containerrepository_versions', 'Can delete repository versions')]},
```
goosemania, line 15: I need to rebase, waiting for other PRs with migrations to be merged since mine is small.

pulp_container/app/viewsets.py
```diff
@@ -545,6 +546,29 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
...
+                "action": ["retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_repo_attr_model_or_obj_perms:container.view_containerrepository",
+            },
+            {
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:delete_containerrepository_versions",
```
ipanova, line 568: apologies if this is not ready for review yet, i have briefly skimmed and noticed we miss an OR here for delete_container_repo  permission

goosemania:
Needs a rebase, waiting on rbac branch to be fixed

pulp_container/tests/functional/api/test_rbac_repo_versions.py
```diff
@@ -0,0 +1,269 @@
...
+        monitor_task(response.task)
+
+        response = self.user_repov_remover["repo_version_api"].delete(create_new_repo_version())
+        monitor_task(response.task)
+
+        response = self.user_repo_remover["repo_version_api"].delete(create_new_repo_version())
+        monitor_task(response.task)
+
+
+class PushRepoVersionTestCase(unittest.TestCase):
```
goosemania, line 155: I wonder if I should move push tests to the module where push content tests are, to reduce the setup time for tests.

goosemania:
Rebased. Green.

pulp_container/app/viewsets.py
```diff
@@ -266,6 +269,35 @@ def untag(self, request, pk):
...
+        try:
+            perm = self.queryset_filtering_required_repo_permission
+        except AttributeError:
+            pass
+        else:
+            repo_version = qs.first()
+            repo = repo_version.repository.cast()
+            if not (self.request.user.has_perm(perm) or self.request.user.has_perm(perm, repo)):
+                raise Http404(_("detail not found"))
+        return qs
```
mdellweg, line 298: :+1:

mdellweg (approved):
LGTM

pulp_container/app/viewsets.py
```diff
@@ -545,12 +578,48 @@ def build_image(self, request, pk):
...
+                "condition": "has_repo_attr_model_or_obj_perms:container.view_containerrepository",
+            },
+            {
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_repo_attr_model_or_obj_perms:container.delete_containerrepository_versions",  # noqa
+                    "has_repo_attr_model_or_obj_perms:container.view_containerrepository",
+                ],
```
ipanova, line 610: :+1

pulp_container/tests/functional/api/test_rbac_repo_versions.py
ipanova: push content tests are already quite long, i believe eventually we goona split those.
My personal preference would be to keep tests organized by 'topic' even though we are going to sacrifice on the setup time.

pulp_container/tests/functional/api/test_rbac_repo_versions.py
```diff
@@ -0,0 +1,269 @@
...
+        """Delete api users and things created in setUpclass."""
+        namespace = cls.namespace_api.list(name="testrv").results[0]
+        cls.namespace_api.delete(namespace.pulp_href)
+        delete_orphans()
+        del_user(cls.user_creator)
+        del_user(cls.user_reader)
+        del_user(cls.user_helpless)
+
+    @classmethod
+    def _pull(cls, image_path, user=None):
```
ipanova, line 203: but i was thinking to move this into the registry base test so we do not copy paste them every time when we write a new module with pull/push tests.

ipanova (approved):
Thank you! :cake: :rocket:

PR #213, opened 2021-01-26
Add support for tagging and untagging for push repositories

lubosmj:
Note that in this commit, there are not handled race conditions where a container client simulateneously tries to push a tag that is being created in an asynchronous task invoked via the Pulp API.

closes #8104

### DIFF
--- a/CHANGES/8104.feature
+++ b/CHANGES/8104.feature
@@ -1,0 +1,1 @@
+Added support for tagging and untagging manifests for push repositories.
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -207,7 +207,66 @@ class ContainerRemoteViewSet(RemoteViewSet):
     }
 
 
-class ContainerRepositoryViewSet(RepositoryViewSet):
+class TagOperationsMixin:
+    """
+    A mixin that adds functionality for creating and deleting tags.
+    """
+
+    @extend_schema(
+        description="Trigger an asynchronous task to tag an image in the repository",
+        summary="Create a Tag",
+        responses={202: AsyncOperationResponseSerializer},
+        request=serializers.TagImageSerializer,
+    )
+    @action(detail=True, methods=["post"], serializer_class=serializers.TagImageSerializer)
+    def tag(self, request, pk):
+        """
+        Create a task which is responsible for creating a new tag.
+        """
+        repository = self.get_object()
+        request.data["repository"] = repository
+
+        serializer = serializers.TagImageSerializer(data=request.data, context={"request": request})
+        serializer.is_valid(raise_exception=True)
+
+        manifest = serializer.validated_data["manifest"]
+        tag = serializer.validated_data["tag"]
+
+        result = enqueue_with_reservation(
+            tasks.tag_image,
+            [repository, manifest],
+            kwargs={"manifest_pk": manifest.pk, "tag": tag, "repository_pk": repository.pk},
+        )
+        return OperationPostponedResponse(result, request)
+
+    @extend_schema(
+        description="Trigger an asynchronous task to untag an image in the repository",
+        summary="Delete a tag",
+        responses={202: AsyncOperationResponseSerializer},
+        request=serializers.UnTagImageSerializer,
+    )
+    @action(detail=True, methods=["post"], serializer_class=serializers.UnTagImageSerializer)
+    def untag(self, request, pk):
+        """
+        Create a task which is responsible for untagging an image.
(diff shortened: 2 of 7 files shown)


pulp_container/app/viewsets.py
```diff
@@ -207,7 +206,99 @@ class ContainerRemoteViewSet(RemoteViewSet):
...
+        return OperationPostponedResponse(result, request)
+
+    @extend_schema(
+        description="Trigger an asynchronous task to copy tags",
+        summary="Copy tags",
+        responses={202: AsyncOperationResponseSerializer},
+        request=serializers.TagCopySerializer,
+    )
+    @action(detail=True, methods=["post"], serializer_class=serializers.TagCopySerializer)
+    def copy_tags(self, request, pk):
```
ipanova, line 274: The goal of this story is to enable only /tag and /untag endpoints on the Push Container repo.

pulp_container/app/viewsets.py
```diff
@@ -547,13 +552,9 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
     parent_viewset = ContainerRepositoryViewSet
 
 
-class ContainerPushRepositoryViewSet(ReadOnlyRepositoryViewSet):
+class ContainerPushRepositoryViewSet(TagOperationsMixin, RepositoryViewSet):
```
ipanova, line 555: I'm afraid we cannot use RepositoyViewSet since it adds the DELETE, only list/get/post should be allowed.

lubosmj: Since I cannot access `AsyncUpdateMixin` from `pulpcore` in order to define custom `ModifiableRepositoryViewSet`, I defined all allowed methods in `http_method_names` manually. This eventually forbids the DELETE operation.

CHANGES/8104.feature
```diff
@@ -0,0 +1 @@
+Added support for creating, removing, and copying tags for push repositories.
```
ipanova, line 1: how about -Added support for tagging and untagging manifests for push repositories

pulp_container/tests/functional/api/test_tagging_images.py
```diff
@@ -174,27 +207,69 @@ def test_05_untag_second_image_again(self):
...
+    def test_02_tag_second_image_with_same_tag(self):
+        """Check if the existing tag correctly references a new manifest."""
+        tagged_image = f"{self.registry_repository_name}:manifest_b"
+        self.registry.pull(tagged_image)
+        local_image_b = self.registry.inspect(tagged_image)
+        self.registry.rmi(tagged_image)
+
+        manifest_b = self.get_manifest_by_tag("manifest_b")
+        self.tag_image(manifest_b, "new_tag")
+        self.registry.pull(tagged_image)
```
ipanova, line 262: you are pulling here still on manifest_b tag, you should update 'tagged_image' string to point to 'new_tag'

pulp_container/app/viewsets.py
mdellweg: So `AsyncUpdateMixin` should become part of the plugin api? But we don't want "update" either on push repos.

ipanova: It will suffice to leave ReadOnlyRepositoryViewSet. This will make sure you cannot perform post/put/patch/delete on the repository object but will still enable POST action for tag which you have defined in TagOperationsMixin.

pulp_container/app/viewsets.py
```diff
@@ -547,18 +552,16 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionViewSet):
...
     """
     ViewSet for a container push repository.
-
```
ipanova, line 553: oh, and please bring back this comment.

pulp_container/tests/functional/api/test_tagging_images.py
lubosmj: Oh, I changed this test recently and forgot to update `tagged_image`. My bad.

pulp_container/app/viewsets.py
```diff
@@ -207,7 +207,66 @@ class ContainerRemoteViewSet(RemoteViewSet):
     }
 
 
-class ContainerRepositoryViewSet(RepositoryViewSet):
+class TagOperationsMixin:
+    """
+    A mixin that adds functionality for creating, deleting, and copying tags.
```
ipanova, line 212: remove the copy tags part please

ipanova (approved):
Thank you, looks great :fire:

PR #215, opened 2021-01-27
Add tag management permissions to the push repo.

ipanova:
[noissue]


pulp_container/app/viewsets.py
```diff
@@ -586,7 +586,7 @@ class ContainerPushRepositoryViewSet(TagOperationsMixin, ReadOnlyRepositoryViewS
                 "principal": "authenticated",
                 "effect": "allow",
                 "condition": [
-                    "has_model_or_obj_perms:container.modify_content_containerrepository",
+                    "has_model_or_obj_perms:container.manage_tags_containerpushrepository",
```
ipanova, line 589: should we have this as granular as possible or rather generalize in this case? for example we could a generic permission
`modify_content_containerrepository` that would allow any content modification whether tagging, or some other action that involves content alteration  for push *and* normal repos.

pulp_container/app/models.py
```diff
@@ -256,6 +256,9 @@ class ContainerPushRepository(Repository, AutoAddObjPermsMixin, AutoDeleteObjPer
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+        permissions = [
+            ("modify_content_containerrepository", "Can modify content in a repository"),
```
goosemania, line 260: Interesting, I'm not sure I understand then how it works.
We have the same permission defined in different classes? Does it work?

If we defined it once, why do we need it in 2 places? For auto-assignment I guess

goosemania:
@ipanova, what do you think about adding a test for a push repo?
Tests saved me many times already with this rbac work.

pulp_container/app/models.py
mdellweg: This does nothing for auto-assignment afaik.
I think it should be `modify_content_containerpushrepository`.

ipanova:
> @ipanova, what do you think about adding a test for a push repo?
> Tests saved me many times already with this rbac work.

what test do you have in mind?

pulp_container/app/models.py
ipanova: It would not work until i would add this permission to this model for ContainerPushRepo.
We had this permission defined on ContainerRepo model but not on the Push repo model.

And yes, i needed to add it to the model because of auto-assignment otherwise i was observing this error:
```
(pulp) [vagrant@pulp3-source-fedora32 _scripts]$ pjournal 
-- Logs begin at Thu 2020-11-19 21:47:04 UTC. --
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/guardian/shortcuts.py", line 133, in assign_perm
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:     return model.objects.assign_perm(perm, user, obj)
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/guardian/managers.py", line 39, in assign_perm
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:     permission = Permission.objects.get(content_type=ctype, codename=perm)
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:     return getattr(self.get_queryset(), name)(*args, **kwargs)
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:   File "/usr/local/lib/pulp/lib64/python3.8/site-packages/django/db/models/query.py", line 406, in get
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]:     raise self.model.DoesNotExist(
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]: django.contrib.auth.models.Permission.DoesNotExist: Permission matching query does not exist.
Jan 28 20:07:38 pulp3-source-fedora32.fluffy.example.com gunicorn[488801]: 127.0.0.1 - - [28/Jan/2021:20:07:38 +0000] "POST /v2/test/this/blobs/uploads/ HTTP/1.1" 500 27 "-" "libpod/2.2.1"

```

ipanova: @mdellweg it works fine, user.has_perm on the object does give me true after i have just created the repo with that user

goosemania:
> > @ipanova, what do you think about adding a test for a push repo?
> > Tests saved me many times already with this rbac work.
> 
> what test do you have in mind?

have a push repo and try to add or remove tags by the owner, by the user with the modify permission only, by the user without modify permissions (this one should fail).

pulp_container/app/models.py
ipanova: after some investigation it turned out that we cannot create two permission with the same codename that would be used on different objects.

pulp_container/app/viewsets.py
```diff
@@ -586,7 +586,7 @@ class ContainerPushRepositoryViewSet(TagOperationsMixin, ReadOnlyRepositoryViewS
                 "principal": "authenticated",
                 "effect": "allow",
                 "condition": [
-                    "has_model_or_obj_perms:container.modify_content_containerrepository",
+                    "has_model_or_obj_perms:container.modify_content_containerpushrepository",
```
ipanova, line 589: should we add another condition to modify_content? without the view_repo permission you won't even be able to `get` it

```
$ http POST  :24817/pulp/api/v3/repositories/container/container-push/c62453c6-2216-4453-9e6e-bfe9cef43f6b/tag/ tag=lala digest='sha256:2c1099ea9644e527fa5f15401c4d9ba62dbf30aa70473e72dfde85450cd96dbd' --auth ina:goodpassword
HTTP/1.1 404 Not Found
Allow: POST, OPTIONS
Connection: close
Content-Length: 23
Correlation-ID: b458c1ba9830487c9466e603195d4ebe
Date: Fri, 29 Jan 2021 21:48:02 GMT
Server: gunicorn/20.0.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "detail": "Not found."
}

```

ipanova: this also stands got other repo actions like sync/copy/etc

mdellweg: I don't think, we have any update-needs-view-permission in any of the other policies.

goosemania: I think it would be good to be explicit, so when users look up the policy they see which permissions are needed.

dkliban (approved):

PR #217, opened 2021-01-29
content rbac

ipanova:


pulp_container/app/viewsets.py
```diff
@@ -113,8 +113,51 @@ class Meta:
...
+push_repo = Repo("container.container-push", "container.view_containerpushrepository")
+
+class ContainerQuerySetViewSet(viewsets.GenericViewSet):
+    def get_queryset(self):
+        qs = super().get_queryset()
+        has_model_push_repo = self.request.user.has_perm(push_repo.perm)
+        has_model_repo = self.request.user.has_perm(mirror_repo.perm)
+        if has_model_push_repo and has_model_repo:
+            return qs
+        #if has_model_repo:
```
ipanova, line 129: I commented this out because it might get false-positive results.
If a user has model level perms for push-repo but only some object-level perms on mirror-repo, we won't show any mirror-repo results.

ipanova: this was an attempt to not go trough lines 140+

ipanova:
@goosemania @dkliban @mdellweg  please take a look and let me know your thoughts.

pulp_container/app/access_policy.py
```diff
@@ -43,3 +44,45 @@ def has_namespace_obj_perms(self, request, view, action, permission):
...
+class ContainerContentAccessPolicyFromDB(AccessPolicyFromDB):
+    """
+    Access policy for Tags/Manifests/Blobs viewsets.
+    """
+
+    def has_repo_query_params(self, request, view, mirror_repo, push_repo):
+        """
+        Checks if the requests' query_params contain repository_version.
+
+        This is used in the quryset scoping for content.
```
mdellweg, line 58:
```suggestion
        This is used in the queryset scoping for content.
```

pulp_container/app/access_policy.py
```diff
@@ -43,3 +44,45 @@ def has_namespace_obj_perms(self, request, view, action, permission):
...
         namespace = view.get_object().namespace
         return request.user.has_perm(permission, namespace)
+
+
+class ContainerContentAccessPolicyFromDB(AccessPolicyFromDB):
+    """
+    Access policy for Tags/Manifests/Blobs viewsets.
+    """
+
+    def has_repo_query_params(self, request, view, mirror_repo, push_repo):
```
mdellweg, line 54: I'm in doubt, that this function can be used in the access policies statements. Not sure whether that's a reason to move it elsewhere, though.

pulp_container/app/viewsets.py
```diff
@@ -114,7 +116,50 @@ class Meta:
         }
 
 
-class TagViewSet(ReadOnlyContentViewSet):
+Repo = namedtuple("Repo", "type perm")
+mirror_repo = Repo("container.container", "container.view_containerrepository")
+push_repo = Repo("container.container-push", "container.view_containerpushrepository")
+
+
+class ContainerQuerySetViewSet(viewsets.GenericViewSet):
```
mdellweg, line 124:
```suggestion
class ContainerContentQuerySetMixin():
```
Not sure if this works properly with `super()`, but worth a try IMHO.

pulp_container/app/viewsets.py
```diff
@@ -114,7 +116,50 @@ class Meta:
...
+    def get_queryset(self):
+        qs = super().get_queryset()
+        has_model_push_repo = self.request.user.has_perm(push_repo.perm)
+        has_model_repo = self.request.user.has_perm(mirror_repo.perm)
+        if has_model_push_repo and has_model_repo:
+            return qs
```
mdellweg, line 130: This would allow to see content that is not associated to any repo.

pulp_container/app/viewsets.py
```diff
@@ -114,7 +116,50 @@ class Meta:
...
+        # if has_model_repo:
+        #    return qs.model.objects.filter(repositories__pulp_type=mirror_repo.type)
+        # if has_model_push_repo:
+        #    return qs.model.objects.filter(repositories__pulp_type=push_repo.type)
+        query_params = self.request.query_params
+        if query_params and "repository_version" in query_params.keys():
+            policy = self.permission_classes[0]
+            repos_can_see, push_repos_can_see = policy.has_repo_query_params(
```
mdellweg, line 138: At the very least, the name of this function confuses me. It sounds like it's supposed to return bool.

pulp_container/app/viewsets.py
```diff
@@ -114,7 +116,50 @@ class Meta:
...
+                self, self.request, self, mirror_repo, push_repo
```
mdellweg, line 139: I believe, we can hardcode those two repo types in that function, no?

pulp_container/app/viewsets.py
```diff
@@ -114,7 +116,50 @@ class Meta:
...
+            from guardian.shortcuts import get_objects_for_user
+
+            push_repos_can_see = get_objects_for_user(
+                self.request.user, push_repo.perm, klass=push_repo_qs
+            ).values_list("pk")
+            repos_can_see = get_objects_for_user(
+                self.request.user, mirror_repo.perm, klass=repo_qs
+            ).values_list("pk")
+        content_qs = qs.model.objects.filter(repositories__in=repos_can_see)
+        push_content_qs = qs.model.objects.filter(repositories__in=push_repos_can_see)
```
mdellweg, line 158: Wondering, if we can make that one filter:
```suggestion
        result_qs = qs.model.objects.filter(repositories__in=repos_can_see + push_repos_can_see)
```

Also by referring to `qs.model.objects` we are resetting every query that was implemented by `super().get_queryset()` before.

pulp_container/app/viewsets.py
ipanova: That is a good observation. I can remove this if we want to display to the user only the content present in the repos he can access. He won't be able to see any 'orphaned' content'. This PR does not solve the 'orphaned content' question. If I leave this then any authed user will be able to see orphaned content.

What would be the preference?

mdellweg: I don't have a preference either way. Maybe never show orphaned content?

goosemania: I think we can leave it as is and change when we resolve orphaned content question.
IMO, content would need to have its own perms to provide enough flexibility and cover many edge cases.

goosemania: I think we can leave it as is and change when we resolve orphaned content question.
IMO, content would need to have its own perms to provide enough flexibility and cover many edge cases.

pulp_container/tests/functional/api/test_rbac_repo_content.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        push_repository_rv = self.pushrepository_api.read(
+            self.push_repository.pulp_href
+        ).latest_version_href
+        push_repository2_rv = self.pushrepository_api.read(
+            self.push_repository2.pulp_href
+        ).latest_version_href
+        repository_rv = self.repository_api.read(self.repository.pulp_href).latest_version_href
+
+        self.assertEqual(self.user_creator["tags_api"].list().count, 10)
```
ipanova, line 133: I tested tags only in this test, let me know if i should also add blobs/manifests. I found adding them redundant since they user same policy and queryset

mdellweg: Sounds good to me.

pulp_container/app/viewsets.py
```diff
@@ -119,7 +124,103 @@ class Meta:
...
+
+            push_repos_can_see = get_objects_for_user(
+                self.request.user, repo_info.push_perm, klass=push_repo_qs
+            ).values_list("pk")
+            mirror_repos_can_see = get_objects_for_user(
+                self.request.user, repo_info.mirror_perm, klass=repo_qs
+            ).values_list("pk")
+            content_qs = qs.model.objects.filter(
+                Q(repositories__in=push_repos_can_see) | Q(repositories__in=mirror_repos_can_see)
+            )
```
goosemania, line 218: I agree on the logic here. I wonder if the queries are too heavy because a lot of filtering is happening by field on a different model causing large joins.
Maybe we can get all repos user has access to and then filter content, is it any worse?
E.g.
```
allowed_push_repos = get_objects_for_user(self.request.user, repo_info.push_perm, klass=models.ContainerPushRepository).only('pk')
allowed_mirror_repos = get_objects_for_user(self.request.user, repo_info.push_perm, klass=models.ContainerRepository).only('pk')
content_qs = qs.model.objects.filter(
￼                Q(repositories__in=allowed_push_repos) | Q(repositories__in=allowed_mirror_repos)
```
It's just avoiding searching repos by content. The cost is dealing with more repos in a subsequent query. I think it's likely ok. Any thoughts or do I miss anything?

I'd ask @dralley on the query performance assessment here.

dralley: I agree.  You can use ```.only("pk")``` just in case if you want.

ipanova: fresh look always helps! i do think it this approach should be better.

pulp_container/app/viewsets.py
```diff
@@ -119,7 +124,99 @@ class Meta:
...
+            return qs
+        query_params = self.request.query_params
+        if query_params and "repository_version" in query_params:
+            repo_pks = self._repo_query_params(self.request, self, repo_info)
+            content_qs = qs.model.objects.filter(repositories__in=repo_pks)
+        else:
+            allowed_push_repos = get_objects_for_user(
+                self.request.user,
+                repo_info.push_perm,
+                klass=models.ContainerPushRepository.objects.all(),
```
goosemania, line 205: Did you add this because specifying just model didn't work?
I'm again unsure about queryset evaluation, I think that all() might cause it to be executed right away while filtering by model is potentially more optimized.
I read here that you can specify a model https://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#get-objects-for-user
```
klass – may be a Model, Manager or QuerySet object. If not given this parameter would be computed based on given params.
```

goosemania (approved):
Great job! Painful task it was, well done!

pulp_container/app/viewsets.py
ipanova: I had an impression klass accepts only a queryset that;s why i used it. Let me try with the model

PR #219, opened 2021-02-01
Change is_private check to is_public

dkliban:
drf-access-policy expects the condition to be a method name. 'not is_private' is not a method name on the viewset. 

[noissue]

### DIFF
--- a/pulp_container/app/access_policy.py
+++ b/pulp_container/app/access_policy.py
@@ -53,11 +53,11 @@ class DistributionAccessPolicyMixin:
         """
         return view.get_object() is not None
 
-    def is_private(self, request, view, action):
+    def is_public(self, request, view, action):
         """
         Check if the distribution is marked private.
         """
-        return view.get_object().private
+        return not view.get_object().private
 
 
 class DistributionAccessPolicyFromDB(AccessPolicyFromDB, DistributionAccessPolicyMixin):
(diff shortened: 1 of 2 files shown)


ipanova (approved):

mdellweg:
This is weird. It works for me and i traced with an exception that the condition was actually called. Also there is support for boolean operations implemented:

https://github.com/rsinger86/drf-access-policy/blob/master/rest_access_policy/access_policy.py#L171

mdellweg:
Also i really want to keep the name the same as the field on the model to not confuse any user.

ipanova:
> Also i really want to keep the name the same as the field on the model to not confuse any user.

stupid idea, can it be the syntax problem ? https://rsinger86.github.io/drf-access-policy/object_level_permissions/ see it is wrapped into parenthesis

ipanova:
@mdellweg even though yesterday we have seen with @goosemania  this problem https://pulp.plan.io/issues/8187

mdellweg:
> @mdellweg even though yesterday we have seen with @goosemania this problem https://pulp.plan.io/issues/8187

That might explain it. I'd consider that a bug in drf-access-policy.

dkliban:
I was experiencing the problem because I was using an older version of drf-access-policy. Closing this PR.

PR #220, opened 2021-02-02
Add pagination for the _catalog registry endpoint

mdellweg:
fixes #7974
https://pulp.plan.io/issues/7974

### DIFF
--- a/CHANGES/7974.feature
+++ b/CHANGES/7974.feature
@@ -1,0 +1,1 @@
+Added pagination to the _catalog and the tags/list endpoint in the registry API.
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -8,6 +8,7 @@ import json
 import logging
 import hashlib
 import re
+from urllib.parse import urlparse, urlunparse, parse_qs, urlencode
 from tempfile import NamedTemporaryFile
 
 from django.core.files.storage import default_storage as storage
@@ -27,9 +28,13 @@ from rest_framework.exceptions import (
     ParseError,
     ValidationError,
 )
+from rest_framework.generics import ListAPIView
+from rest_framework.pagination import BasePagination
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.renderers import BaseRenderer, JSONRenderer
 from rest_framework.response import Response
+from rest_framework.serializers import ModelSerializer
+from rest_framework.settings import api_settings
 from rest_framework.viewsets import ViewSet
 from rest_framework.views import APIView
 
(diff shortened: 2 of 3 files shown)


lubosmj (changes requested):
Awesome! I have tested the changes and they seem to work most of the time:
```
(pulp) [vagrant@pulp3-source-fedora32 backup]$ http :24817/v2/namespace/busybox/tags/list?n=10
HTTP/1.1 200 OK
Allow: GET, HEAD, OPTIONS
Connection: close
Content-Length: 47
Correlation-ID: a62ebd01565847d5bb666b0ccf7ac2af
Date: Fri, 05 Feb 2021 12:54:39 GMT
Docker-Distribution-Api-Version: registry/2.0
Link: <http://localhost:24817/v2/namespace/busybox/tags/list?n=10&last=1-glibc>; rel="next"
Server: gunicorn/20.0.4
Vary: Accept
X-Frame-Options: SAMEORIGIN

{
    "name": "namespace/busybox",
    "tags": [
        "1-glibc"
    ]
}


(pulp) [vagrant@pulp3-source-fedora32 backup]$ http :24817/v2/namespace/busybox/tags/list?n=-1
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Length: 27
Content-Type: text/html
Correlation-ID: 03767513c8ba4814a2fa7d2c13094656
Date: Fri, 05 Feb 2021 12:54:46 GMT
Server: gunicorn/20.0.4
X-Frame-Options: SAMEORIGIN

<h1>Server Error (500)</h1>
```

Also, I am not sure whether we want to mimic the behavior of dockerhub or not, but when I query the tags/list endpoint, it does not return the `Link` header if the number of entries is not (presumably) large enough:
```
(pulp) [vagrant@pulp3-source-fedora32 backup]$ http GET https://registry-1.docker.io/v2/lubosmj/mirror/tags/list?n=1 --auth-type=jwt --auth=blabla
HTTP/1.1 200 OK
Content-Length: 56
Date: Fri, 05 Feb 2021 12:58:53 GMT
Docker-Distribution-Api-Version: registry/2.0
Strict-Transport-Security: max-age=31536000

{
    "name": "lubosmj/mirror",
    "tags": [
        "123",
        "3.8",
        "latest"
    ]
}

(pulp) [vagrant@pulp3-source-fedora32 backup]$ http GET https://registry-1.docker.io/v2/lubosmj/mirror/tags/list --auth-type=jwt --auth=blabla
HTTP/1.1 200 OK
Content-Length: 56
Date: Fri, 05 Feb 2021 12:44:48 GMT
Docker-Distribution-Api-Version: registry/2.0
Strict-Transport-Security: max-age=31536000

{
    "name": "lubosmj/mirror",
    "tags": [
        "123",
        "3.8",
        "latest"
    ]
}
```

pulp_container/app/registry_api.py
```diff
@@ -345,36 +349,132 @@ def get(self, request):
...
+        try:
+            self.n = int(request.query_params.get("n"))
+        except Exception:
+            self.n = self.page_size
+        last = request.query_params.get("last")
+        self.url = request.build_absolute_uri()
+
+        if last:
+            queryset = queryset.filter(base_path__gt=last)
+        return queryset.order_by("base_path")[: self.n]
```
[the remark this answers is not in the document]
mdellweg: I thought about that and then there were subtle differences. But now that we know how it looks, i might be able to extract a base class.

pulp_container/app/registry_api.py
```diff
@@ -345,36 +349,132 @@ def get(self, request):
...
+        query_params["last"] = repositories_names[-1]
+        parsed_url[4] = urlencode(query_params)
+        url = urlunparse(parsed_url)
+        log.info(url)
+        headers = {
+            "Link": f'<{url}>; rel="next"',
+        }
+        return Response(headers=headers, data={"repositories": list(repositories_names)})
+
+    page_size = 16  # TODO what is a reasonable default here?
```
ipanova, line 399: i have seen 50 on quay

ipanova:
@mdellweg did you happen to stumble across this? https://github.com/tbeadle/django-rest-framework-link-header-pagination

pulp_container/app/registry_api.py
ipanova, line 380: we can return here qs.value_list('base_path', flat=true) so then later you can create a list and not iterate trough qs.

pulp_container/app/registry_api.py
```diff
@@ -345,36 +349,132 @@ def get(self, request):
...
+
+    def get_paginated_response(self, data):
+        """
+        Prepare the paginated container _catalog response.
+        """
+        repositories_names = [repo["base_path"] for repo in data]
+        parsed_url = list(urlparse(self.url))
+        query_params = parse_qs(parsed_url[4])
+        query_params["n"] = str(self.n)
+        query_params["last"] = repositories_names[-1]
```
ipanova, line 390: this will give en error if repo_names is empty list

pulp_container/app/registry_api.py
```diff
@@ -345,36 +349,132 @@ def get(self, request):
...
+        parsed_url[4] = urlencode(query_params)
+        url = urlunparse(parsed_url)
+        log.info(url)
+        headers = {
+            "Link": f'<{url}>; rel="next"',
+        }
+        return Response(headers=headers, data={"repositories": list(repositories_names)})
```
ipanova, line 397: repo_name is already a list

ipanova: i see a problem with this approach - the link header should be present in the response only when it indicates that not all the results have been displayed.
This way we will always have link in the header which is wrong.

ipanova:
```
(pulp) [vagrant@pulp3-source-fedora32 ~]$ http GET :24817/v2/test/this/tags/list
HTTP/1.1 200 OK
Allow: GET, HEAD, OPTIONS
Connection: close
Content-Length: 40
Correlation-ID: 39902f76fc6a4b2696d31040ceeba9ec
Date: Wed, 17 Feb 2021 17:33:41 GMT
Docker-Distribution-Api-Version: registry/2.0
Link: <http://localhost:24817/v2/test/this/tags/list?n=16&last=mytag1.8>; rel="next"
Server: gunicorn/20.0.4
Vary: Accept
X-Frame-Options: SAMEORIGIN

{
    "name": "test/this",
    "tags": [
        "mytag1.8"
    ]
}

```

pulp_container/app/registry_api.py
```diff
@@ -345,36 +349,132 @@ def get(self, request):
...
+        Analyse the pagination parameters and prepare the queryset.
+        """
+        try:
+            self.n = int(request.query_params.get("n"))
+        except Exception:
+            self.n = self.page_size
+        last = request.query_params.get("last")
+        self.url = request.build_absolute_uri()
+
+        if last:
```
ipanova, line 378: to be honest i don't have a 100% understanding on how `last` works, how is it handled the case when some non-existent vale has been provided in the request query param?

pulp_container/app/registry_api.py
ipanova: however i agree on the approach to paginate the response even if the client did not sent request with pagination.

ipanova:
imo, i would make this as simple as possible and:
1. return a paginated response no matter whether client has asked for it
2. support only `n` so a client can ask for a custom pagination other then the one set in our settings( it is 100)
3. the link header should be present only when not the full response has been produced

pulp_container/app/registry_api.py
mdellweg: Well i suspect, that is why you specify last and not "first of next query". It does not matter if it got deleted in the meantime. You just list, what comes later in alphabethical order.

pulp_container/app/registry_api.py
mdellweg: That needs fixing for sure!

pulp_container/app/registry_api.py
mdellweg: Not sure if we can get away with this. If you follow the next link here (provided i fixed the empty list problem) you'll get an empty list, and you stop iterating.
But i will try to find a solution without this extra call anyway.

pulp_container/app/registry_api.py
mdellweg: Hmm i turned this endpoint into a ListViewset, and it now uses a serializer. I think this will not work well with `value_list`.
Maybe `only("base_path")` can be an improvement still.

pulp_container/app/registry_api.py
mdellweg: Now, with the empty list, there will be no next link anymore. But as long as we return something, i do not see, where i can get the information that there would be more.

ipanova: one more thing i would suggest is to return in the link header a relative URI instead of absolute. According to RFC both are support though

pulp_container/app/registry_api.py
```diff
@@ -378,36 +383,136 @@ def get(self, request):
...
+        """
+        Analyse the pagination parameters and prepare the queryset.
+        """
+        try:
+            self.n = int(request.query_params.get("n"))
+        except Exception:
+            self.n = self.page_size
+        else:
+            if self.n > 10 * self.page_size:
+                self.n = 10 * self.page_size
```
mdellweg, line 411: For the greedy.
In the end the pagination is not really a service to the user, but a precautionary measure against DOS.

ipanova:
@lubosmj how funny or sad it might sounds but i am pretty sure dockehub does not support pagination :) I have tried to list tags from mongo and it returned me 800+ tags. specidying `n` did not make any difference :)

CHANGES/7974.feature
```diff
@@ -0,0 +1 @@
+Added pagination to the _catalog and the tags endpoint in the registry API.
```
ipanova, line 1: s/tags/tags/list

pulp_container/app/registry_api.py
ipanova: i would not know what's the reasonable max page_size, probably 1000 is ok if having 100 as defaults unchanged

pulp_container/app/registry_api.py
```diff
@@ -378,36 +383,136 @@ def get(self, request):
...
+            parsed_url = list(urlparse(self.url))
+            query_params = parse_qs(parsed_url[4])
+            query_params["n"] = str(self.n)
+            query_params["last"] = tag_names[-1]
+            parsed_url[4] = urlencode(query_params)
+            url = urlunparse(parsed_url)
+            headers["Link"] = f'<{url}>; rel="next"'
+        return Response(headers=headers, data={"name": self.path, "tags": tag_names})
+
+    page_size = api_settings.PAGE_SIZE
```
ipanova, line 498: any reason this is defined twice?

pulp_container/app/registry_api.py
```diff
@@ -378,36 +383,136 @@ def get(self, request):
...
+        headers = {}
+        tag_names = [tag["name"] for tag in data]
+        if tag_names:
+            parsed_url = list(urlparse(self.url))
+            query_params = parse_qs(parsed_url[4])
+            query_params["n"] = str(self.n)
+            query_params["last"] = tag_names[-1]
+            parsed_url[4] = urlencode(query_params)
+            url = urlunparse(parsed_url)
+            headers["Link"] = f'<{url}>; rel="next"'
```
ipanova, line 495: This will still send the link header even when whole result is displayed.

there are 2 options:

1. check in len of the tags_names. 
```
 def get_paginated_response(self, data):
        """
        Prepare the paginated container _catalog response.
        """
        headers = {}
        tag_names = [tag["name"] for tag in data]
        if tag_names:
            parsed_url = list(urlparse(self.url))
            query_params = parse_qs(parsed_url[4])
            query_params["n"] = str(self.n)
            query_params["last"] = tag_names[-1]
            parsed_url[4] = urlencode(query_params)
            url = urlunparse(parsed_url)
            if len(tag_names) == self.n:
                headers["Link"] = f'<{url}>; rel="next"'
        return Response(headers=headers, data={"name": self.path, "tags": tag_names})
```
if there have been requests 120 tags out of 120 we will still produce one unnecessary link header which will lead to to empty list. We can live with that alas.
2. If we want to eliminate even this conrner case we could store a result of the rest of the qs self.count_rest_of_qs = queryset.order_by("name")[self.n: ].count()` and compare the len of tag_names with value of this var, if 0 means we have returned full data. Might be performance not optimal :)

ipanova:
@mdellweg the pr looks good :octocat:  please take a look at this comment https://github.com/pulp/pulp_container/pull/220#discussion_r578650103
I would really love to see this pagination code for tags and catalog consolidated into one code path, but let's merge this as is for now.
I would also clean up registry_api.py and leave only viewes/viewsets related stuff and for example extract serializers into a separate file. But we can do this as well in separate PR.

CHANGES/7974.feature
mdellweg:
```suggestion
Added pagination to the _catalog and the tags/list endpoint in the registry API.
```
Ha, i needed to read you sed-ish twice...

pulp_container/app/registry_api.py
mdellweg: I like option one. Even in option 2, there could be one more repo that gets deleted between the two calls and we'd return the empty list. So the client must be prepared for this event (getting an empty list) anyway.

ipanova: concur :) i decided to list 2nd option for the sake of completeness.

mdellweg:
I think consolidating the code paths would be possible, once we add renderers to the mix.
ATM the pagination class takes care of the rendering, which is not ideal.

ipanova (approved):
thank you! :clinking_glasses:

PR #221, opened 2021-02-04
Handle HEAD requests as if they are GET requests

dkliban:
The Content-Type header is stripped by DRF when there is no body in the response.

Django and aiohttp both strip the body of HEAD requests. It is safe to handle
HEAD requests the same as GET requests and rely on the web server to strip the
body of the response.

This patch also adds the Content-Type header to the ManifestResponse objects
which are used when responding to requests of Manifest creation.

fixes: #8208
https://pulp.plan.io/issues/8208

### DIFF
--- a/CHANGES/8208.bugfix
+++ b/CHANGES/8208.bugfix
@@ -1,0 +1,1 @@
+Fixed bug experienced when pulling using docker 20.10 client.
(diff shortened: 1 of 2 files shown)


mdellweg (approved):
I assume, this is not interfering with the type conversion, right?

Should we have a test?

dkliban:
> I assume, this is not interfering with the type conversion, right?
> 
> Should we have a test?

This will not interfere with the type conversion. I am not sure that a test is needed with the latest implementation of the fix.

mdellweg (approved):

PR #225, opened 2021-02-05
Add view_permission to the conditions of other actions on the object.

ipanova:
[noissue]


mdellweg (approved):
I think, we agreed, that this is the more specific version of it, but it does not change behavior.

ipanova:
> I think, we agreed, that this is the more specific version of it, but it does not change behavior.

yes, it does not change the behaviour, just makes the policy and rules explicit.

dkliban (approved):

PR #226, opened 2021-02-06
Return valid scope for the catalog endpoint

lubosmj:
In this commit, there was removed the 'catalog' policy from 'ContainerDistributionViewSet' and the initial permission checks for the catalog endpoint were restored. It does not make too much sense to couple the catalog endpoint with a specific distribution. Nevertheless, the distribution cannot be determined when accessing the catalog endpoint.

closes #8045


pulp_container/app/authorization.py
```diff
@@ -205,19 +207,10 @@ def has_view_catalog_permissions(self, path):
...
-        request.method = "GET"
-        request.user = self.user
-        request._full_data = {"base_path": path}
-        # Fake the corresponding view
-        view = namedtuple("FakeView", ["action", "get_object"])("catalog", lambda: distribution)
-        return self.access_policy.has_permission(request, view)
+        if path == "catalog" and self.user.is_staff:
+            return True
+        else:
+            return False
```
mdellweg, line 213: This change is reverting to an old behavior, where we had the permission hard coded.
I think we want to keep this going through the access_policy.

pulp_container/app/token_verification.py
```diff
@@ -99,23 +110,41 @@ def authenticate(self, request):
...
         realm = settings.TOKEN_SERVER
-        repository_path, access_action = _access_scope(request)
-
         authenticate_string = f'{self.keyword} realm="{realm}",service="{request.get_host()}"'
 
-        if repository_path:
-            scope = f"repository:{repository_path}:{access_action}"
-            authenticate_string += f',scope="{scope}"'
+        try:
+            scope = ScopeFactory.from_request(request)
```
mdellweg, line 123: Why the trouble of introducing three new classes just to juggle three strings?

I know this is about JWT tokens, but does that mean we need to code like it was Java? ;)

mdellweg (changes requested):
I do not see, where the single problem described in the ticket is handled?
I'd expect something to go into `authenticate_header` of the `TokenAuthentication` class, where the scope is only generated for repositories atm.

Also the rest of this PR looks a bit over engineered to me.

mdellweg:
Sorry for the mass close by removing the `rbac` branch. Please reopen against master.

PR #232, opened 2021-02-08
Update branch versions

ipanova:


.github/workflows/scripts/install.sh
```diff
@@ -44,7 +44,7 @@ image:
   tag: "${TAG}"
 plugins:
   - name: pulpcore
-    source: pulpcore
```
goosemania, line 47: I was not aware of this one

.github/workflows/scripts/install.sh
```diff
@@ -44,7 +44,7 @@ image:
...
+    source: pulpcore>=3.10,<3.12
```
goosemania, line 47: I was not aware of this one

goosemania (approved):

mdellweg (approved):
LGTM

PR #234, opened 2021-02-08
Fall back to BasicAuth if token is disabled

mdellweg:
fixes #8074
https://pulp.plan.io/issues/8074

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -27,6 +27,7 @@ from rest_framework.exceptions import (
     ParseError,
     ValidationError,
 )
+from rest_framework.permissions import IsAuthenticated
 from rest_framework.renderers import BaseRenderer, JSONRenderer
 from rest_framework.response import Response
 from rest_framework.viewsets import ViewSet
@@ -35,7 +36,12 @@ from rest_framework.views import APIView
 from pulp_container.app import models, serializers
 from pulp_container.app.authorization import AuthorizationService
 from pulp_container.app.redirects import FileStorageRedirects, S3StorageRedirects
-from pulp_container.app.token_verification import TokenAuthentication, TokenPermission
+from pulp_container.app.token_verification import (
+    RegistryAuthentication,
+    TokenAuthentication,
+    RegistryPermission,
+    TokenPermission,
+)
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


pulp_container/app/registry_api.py
```diff
@@ -211,10 +212,21 @@ class ContainerRegistryApiMixin:
...
     """
 
-    authentication_classes = [TokenAuthentication]
     permission_classes = [TokenPermission]
     schema = None
 
+    @property
+    def authentication_classes(self):
+        if settings.get("TOKEN_AUTH_DISABLED", False):
+            return api_settings.DEFAULT_AUTHENTICATION_CLASSES
```
mdellweg, line 221: I decided that this is the better way, as we can fallback to the _list_ of configured auth-classes.
Also it does not mess with the class creation process as `__new__` does.

pulp_container/app/token_verification.py
```diff
@@ -116,6 +113,26 @@ def authenticate_header(self, request):
...
+    """
+
+    message = "Access to the requested resource is not authorized."
+
+    def has_permission(self, request, view):
+        """
+        Decide upon permission based on user.
+        """
+        # TODO RBAC goes here
+        if request.user.is_staff:
```
ipanova, line 128: I have tested this pr, user here is anonymours no matter what i am doing
1. login into the registry permits to login even with wrong credentials
2. if i'm accessing api directly this codepath is niot executed and i can perform get/post/put call without any auth provided.

pulp_container/app/registry_api.py
ipanova: this changed fixed the direct api calls but still not for client push
`return [api_settings.DEFAULT_AUTHENTICATION_CLASSES[1]]`

ipanova: i get this `"Attempting next endpoint for push after error: unauthorized: Authentication credentials were not provided."` 
it feels like the client does not send the basic auth creds in the calls?

mdellweg: I used podman, and push with logged in as admin worked for me.

ipanova: podman-2.2.1-1.fc32.x86_64

pulp_container/app/token_verification.py
mdellweg: You are right. As long as we don't force "/v2/" to need credentials, podman will never try to provide them.
But we break anonymous pull if we do, because he will still try to find that and fallback to /v1ping/ instead.
This needs more investigation.

ipanova (approved):
with this pr i was able to push with admin credentials and to pull whether with a valid logged in user or not being logged in( anonymous pull). I can confirm that push is possible only with admin credentials.
This PR however does does not take into account whether a distribution is private or not during pull operation because policy is defined on the distribution/namespace viewsets which are not used whenever token_auth is disabled.

PR #235, opened 2021-02-12
Use DistributionFilter directly

lubosmj:
[noissue]


mdellweg (approved):
This is in the plugin api of 3.10. right?

lubosmj:
> This is in the plugin api of 3.10. right?

Yes, it is.

PR #236, opened 2021-02-12
Use get_user_model instead of auth.models.user

newswangerd:
Importing and using `django.contrib.auth.models.User` breaks any apps that use custom user models defined in `AUTH_USER_MODEL`. To prevent this from happening, the `User` model should be loaded using django's `get_user_model()` function.

See https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#referencing-the-user-model for details

### DIFF
--- a/CHANGES/8260.bugfix
+++ b/CHANGES/8260.bugfix
@@ -1,0 +1,1 @@
+Use ``get_user_model()`` to prevent pulp_container from crashing when running alongside other pulp plugins that override the default user authentication models. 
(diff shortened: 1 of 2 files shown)


lubosmj (commented):
Thanks for opening this PR. Would you mind adding `[noissue]` to the commit message in order to make the tests pass?

newswangerd:
> Thanks for opening this PR. Would you mind adding `[noissue]` to the commit message in order to make the tests pass?

Done. For some reason I thought the tag was `[no-issue]`

ipanova (approved):

mdellweg (approved):

PR #237, opened 2021-02-12
Return error code in the WWW-Authenticate header

lubosmj:
closes #8046

### DIFF
--- a/pulp_container/app/token_verification.py
+++ b/pulp_container/app/token_verification.py
@@ -6,7 +6,7 @@ from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth import get_user_model
 
 from rest_framework.authentication import BaseAuthentication, BasicAuthentication
-from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.exceptions import AuthenticationFailed, NotAuthenticated
 from rest_framework.permissions import BasePermission, SAFE_METHODS
 
 User = get_user_model()
@@ -106,6 +106,7 @@ class TokenAuthentication(BaseAuthentication):
             raise AuthenticationFailed(
                 detail="Access to the requested resource is not authorized. "
                 "The provided Bearer token is invalid.",
+                code="invalid_token",
             )
 
         username = decoded_token.get("sub")
@@ -169,6 +170,9 @@ class TokenPermission(BasePermission):
         """
         try:
             decoded_token = request.auth
-            return _contains_accessible_actions(decoded_token, *_access_scope(request))
+            if _contains_accessible_actions(decoded_token, *_access_scope(request)):
+                return True
         except Exception:
-            return False
+            raise NotAuthenticated()
+
+        raise AuthenticationFailed(detail="Insufficient permissions", code="insufficient_scope")
--- a/CHANGES/8046.bugfix
+++ b/CHANGES/8046.bugfix
@@ -1,0 +1,1 @@
+Added missing error code that should be returned in the WWW-Authenticate header.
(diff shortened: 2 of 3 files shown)


pulp_container/app/token_verification.py
```diff
@@ -127,10 +128,14 @@ def has_permission(self, request, view):
...
         try:
-            decoded_token = request.auth
-            return decoded_token is True or _contains_accessible_actions(
-                decoded_token, *_access_scope(request)
-            )
+            if _contains_accessible_actions(decoded_token, *_access_scope(request)):
+                return True
         except Exception:
-            return False
+            raise NotAuthenticated()
```
mdellweg, line 139: No code here?

Also the examples in https://www.django-rest-framework.org/api-guide/permissions/ all return boolean.
And the docs talk about a `PermissionDenied` exception. (It has a `code` attribute too.)

lubosmj: No error code means that a user did not provide any authentication credentials. In this case, dockerhub returns an "UNAUTHORIZED" response without the error code. I tried to employ the same pattern in Pulp.

Speaking of `PermissionDenied`, all raised exceptions are handled by the method `handle_exception()`:https://github.com/pulp/pulp_container/blob/ca86159b39f110087abc7eac9f24ff5233460f49/pulp_container/app/registry_api.py#L237

Here, the authenticate header is built after invoking `super().handle_exception(exc)`: https://github.com/encode/django-rest-framework/blob/d635bc9c71b6cf84d137a68610ae2e628f8b62b3/rest_framework/views.py#L456. Since we already construct the authenticate header by ourselves, with my changes we just append the error code to it only when the exception `AuthenticationFailed` with a code is raised.

On the other hand, the exception `PermissionDenied` does not result into calling the method `authenticate_header()`: https://github.com/pulp/pulp_container/blob/ca86159b39f110087abc7eac9f24ff5233460f49/pulp_container/app/token_verification.py#L100

We would be required to parse the request manually, as they do in `django-rest-framework` and make the method `authenticate_header()` to be accessible globally. But then we will return "DENIED" in the response message, as opposed to "UNAUTHORIZED" in dockerhub.

pulp_container/app/token_verification.py
```diff
@@ -127,10 +128,14 @@ def has_permission(self, request, view):
...
+
+        raise AuthenticationFailed(code="insufficient_scope")
```
ipanova, line 141: if you don't add a `detail` here the user will get ` incorrect username/password` which is misleading
can you add `detail=insufficient permissions` or something along the lines

pulp_container/app/token_verification.py
ipanova: i agree we should return 401 and not 403

ipanova:
i did some tests and i do see errors in the auth-header

lubosmj:
I decided to return an HTTP 401 response in every single case (no token provided, invalid token, insufficient scope) of failure since we did not reach a consensus.

pulp_container/app/token_verification.py
mdellweg: What needs to happen anyway that we come to this point and raise this exception?
Is this a case of "invalid-token" or "no-token"?

lubosmj: `NotAuthenticated` means that a user did not provide any authentication credentials.

mdellweg: In that case, `request.auth` should be None, right? That should work.

mdellweg (approved):

ipanova (approved):
thank you!

PR #238, opened 2021-02-12
Return valid scope for the catalog endpoint

lubosmj:
closes #8045

### DIFF
--- a/pulp_container/app/authorization.py
+++ b/pulp_container/app/authorization.py
@@ -20,6 +20,8 @@ from pulp_container.app.access_policy import RegistryAccessPolicy
 
 TOKEN_EXPIRATION_TIME = settings.get("TOKEN_EXPIRATION_TIME", 300)
 
+FakeView = namedtuple("FakeView", ["action", "get_object"])
+
 
 class AuthorizationService:
     """
@@ -28,9 +30,6 @@ class AuthorizationService:
     This class represents a token server which manages and grants permissions
     according to a user's scope.
     """
-
-    ANONYMOUS_USER = "AnonymousUser"
-    VALID_ACTIONS = ["pull", "push", "*"]
 
     def __init__(self, user, service, scope):
         """
@@ -75,7 +74,7 @@ class AuthorizationService:
         access = self.determine_access()
         token_server = getattr(settings, "TOKEN_SERVER", "")
         claim_set = self.generate_claim_set(
-            access=[access],
+            access=access,
             audience=self.service,
             issued_at=int(current_datetime.timestamp()),
             issuer=token_server,
@@ -129,9 +128,14 @@ class AuthorizationService:
         The determination is based on role based access control.
 
         Returns:
-            list: An intersected set of the requested and the allowed access.
+            list: An intersected set of the requested and the allowed access if the scope was
+                specified. Otherwise, returns an empty list indicating permission for the root
+                endpoint.
 
         """
+        if not self.scope:
+            return []
+
         typ, name, actions = self.scope.split(":")
         actions = set(actions.split(","))
 
@@ -149,7 +153,7 @@ class AuthorizationService:
             if has_permission:
                 permitted_actions.add(action)
 
-        return {"type": typ, "name": name, "actions": list(permitted_actions)}
+        return [{"type": typ, "name": name, "actions": list(permitted_actions)}]
 
     def has_permission(self, obj, method, action, data):
         """Check if user has permission to perform action."""
@@ -160,7 +164,7 @@ class AuthorizationService:
         request.user = self.user
         request._full_data = data
         # Fake the corresponding view
-        view = namedtuple("FakeView", ["action", "get_object"])(action, lambda: obj)
+        view = FakeView(action, lambda: obj)
         return self.access_policy.has_permission(request, view)
 
     def has_pull_permissions(self, path):
@@ -203,20 +207,17 @@ class AuthorizationService:
 
     def has_view_catalog_permissions(self, path):
         """
-        Check if the authenticated user is an administrator and is requesting the catalog endpoint.
+        Check if the authenticated user has permission to access the catalog endpoint.
         """
-        try:
-            distribution = ContainerDistribution.objects.get(base_path=path)
-        except ContainerDistribution.DoesNotExist:
-            distribution = None
+        if path != "catalog":
+            return False
 
         # Fake the request
         request = Request(HttpRequest())
         request.method = "GET"
         request.user = self.user
-        request._full_data = {"base_path": path}
-        # Fake the corresponding view
-        view = namedtuple("FakeView", ["action", "get_object"])("catalog", lambda: distribution)
+        # Fake the view
+        view = FakeView("catalog", lambda: ContainerDistribution())
         return self.access_policy.has_permission(request, view)
 
     @staticmethod
--- a/pulp_container/app/token_verification.py
+++ b/pulp_container/app/token_verification.py
@@ -1,6 +1,8 @@
 import jwt
 import logging
 
+from collections import namedtuple
+
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth import get_user_model
@@ -9,6 +11,7 @@ from rest_framework.authentication import BaseAuthentication, BasicAuthenticatio
 from rest_framework.exceptions import AuthenticationFailed, NotAuthenticated
 from rest_framework.permissions import BasePermission, SAFE_METHODS
 
+Scope = namedtuple("Scope", "resource_type, name, action")
 User = get_user_model()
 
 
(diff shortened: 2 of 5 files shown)


pulp_container/app/authorization.py
```diff
@@ -203,20 +205,17 @@ def has_push_permissions(self, path):
 
     def has_view_catalog_permissions(self, path):
```
mdellweg, line 206:
```suggestion
    def has_view_catalog_permissions(self):
```
Would this work too? The catalog has no `path`.

lubosmj: All the methods which check for permissions have the same input parameters and are called in the loop: https://github.com/pulp/pulp_container/blob/0de33d42bf5f86b11d29a99757b5f6a53cca8d40/pulp_container/app/authorization.py#L148

Furthermore, we do not check anywhere whether the path is truly "catalog" or not. That is why I left the path parameter there.

pulp_container/app/authorization.py
```diff
@@ -20,6 +20,8 @@
 
 TOKEN_EXPIRATION_TIME = settings.get("TOKEN_EXPIRATION_TIME", 300)
 
+FakeView = namedtuple("FakeView", ["action", "get_object"])
```
mdellweg, line 23: :+1:

pulp_container/app/authorization.py
mdellweg: Isn't the path `_catalog` then?

https://docs.docker.com/registry/spec/api/#listing-repositories

pulp_container/tests/functional/api/test_repositories_list.py
```diff
@@ -93,8 +93,10 @@ def test_listing_repositories(self):
         authenticate_header = content_response.headers["Www-Authenticate"]
 
         queries = AuthenticationHeaderQueries(authenticate_header)
-        self.assertFalse(hasattr(queries, "scope"))
-        content_response = requests.get(queries.realm, params={"service": queries.service})
+        self.assertEqual(queries.scope, "registry:catalog:*")
```
mdellweg, line 96: :+1:

pulp_container/app/token_verification.py
```diff
@@ -8,49 +8,60 @@
 from rest_framework.permissions import BasePermission, SAFE_METHODS
 
 
-def _decode_token(encoded_token, request):
```
mdellweg, line 11: My comments on the old PR for this file are still valid.
I think the factory pattern, to compare three strings here, is adding an overwhelming amount of complexity.

ipanova:
I did not expect that much of complexity in this PR.  My thought was to add where needed logic for the catalog endpoint.
```
diff --git a/pulp_container/app/authorization.py b/pulp_container/app/authorization.py
index d6790e5..1d58a7f 100644
--- a/pulp_container/app/authorization.py
+++ b/pulp_container/app/authorization.py
@@ -205,18 +205,18 @@ class AuthorizationService:
         """
         Check if the authenticated user is an administrator and is requesting the catalog endpoint.
         """
-        try:
-            distribution = ContainerDistribution.objects.get(base_path=path)
-        except ContainerDistribution.DoesNotExist:
-            distribution = None
+        # try:
+        #    distribution = ContainerDistribution.objects.get(base_path=path)
+        # except ContainerDistribution.DoesNotExist:
+        #    distribution = None

         # Fake the request
         request = Request(HttpRequest())
         request.method = "GET"
         request.user = self.user
-        request._full_data = {"base_path": path}
+        # request._full_data = {"base_path": path}
         # Fake the corresponding view
-        view = namedtuple("FakeView", ["action", "get_object"])("catalog", lambda: distribution)
+        view = namedtuple("FakeView", ["action", "get_object"])("catalog", lambda: ContainerDistribution())
         return self.access_policy.has_permission(request, view)

     @staticmethod
diff --git a/pulp_container/app/registry_api.py b/pulp_container/app/registry_api.py
index a383785..96ef4e2 100644
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -385,6 +385,9 @@ class CatalogView(ContainerRegistryApiMixin, APIView):

     def get(self, request):
         """Handles GET requests for the /v2/_catalog endpoint."""
+        # TODO check request.user and display distribution only
+        # he has perms on
+        # Not part of this PR
         repositories_names = models.ContainerDistribution.objects.values_list(
             "base_path", flat=True
         )
diff --git a/pulp_container/app/token_verification.py b/pulp_container/app/token_verification.py
index a153850..bcb2171 100644
--- a/pulp_container/app/token_verification.py
+++ b/pulp_container/app/token_verification.py
@@ -34,16 +34,19 @@ def _decode_token(encoded_token, request):

 def _access_scope(request):
     repository_path = request.resolver_match.kwargs.get("path", "")
+    catalog_path = ""

     if request.method in SAFE_METHODS:
         access_action = "pull"
     else:
         access_action = "push"
+    if 'catalog' in request.path:
+        catalog_path = "catalog"
+        access_action = '*'
+    return repository_path, catalog_path, access_action

-    return repository_path, access_action

-
-def _contains_accessible_actions(decoded_token, repository_path, access_action):
+def _contains_accessible_actions(decoded_token, repository_path, catalog_path, access_action):
     """
     Check if a client has an access permission to execute the pull/push operation.

@@ -57,6 +60,8 @@ def _contains_accessible_actions(decoded_token, repository_path, access_action):
                 return True
             if not repository_path:
                 return True
+
+    # TODO add catalog logic
     return False

@@ -127,13 +132,16 @@ class TokenAuthentication(BaseAuthentication):
         and will be used in the "WWW-Authenticate" header.
         """
         realm = settings.TOKEN_SERVER
-        repository_path, access_action = _access_scope(request)
+        repository_path, catalog_path, access_action = _access_scope(request)

         authenticate_string = f'{self.keyword} realm="{realm}",service="{request.get_host()}"'

         if repository_path:
             scope = f"repository:{repository_path}:{access_action}"
             authenticate_string += f',scope="{scope}"'
+        elif catalog_path:
+            scope = f"registry:{catalog_path}:{access_action}"
+            authenticate_string += f',scope="{scope}"'
         return authenticate_string
```

Thoughts?

pulp_container/app/authorization.py
lubosmj: No, it is not. Because we are parsing the scope string (scope="registry:catalog:*") and not the endpoint itself during the authorization.

lubosmj:
I got rid of the factory pattern and additional java-like classes. Now, there is only one function `init_scope()` that has the same functionality. Other changes are mostly related to the methods' rearranging.

pulp_container/app/token_verification.py
```diff
@@ -167,8 +163,32 @@ def has_permission(self, request, view):
...
+            return is_requesting_root_endpoint(decoded_token)
+        else:
+            return contains_accessible_actions(decoded_token, scope)
+
+
+def is_requesting_root_endpoint(decoded_token):
+    """
+    Returns True if no access type was detected.
+    """
+    return len(decoded_token["access"]) == 0
```
mdellweg, line 182: I don't see the value in this extra function. You can just insert the expression in the place where you are using it.
And since it's purpose is not so obvious, that seems like the right place for one line of comment (the function name would do just well).

pulp_container/app/token_verification.py
```diff
@@ -167,8 +163,32 @@ def has_permission(self, request, view):
...
+def contains_accessible_actions(decoded_token, scope):
+    """
+    Check if a client has permission to perform operations within the current scope
+    """
+    for access in decoded_token["access"]:
+        if scope.resource_type == access["type"] and scope.name == access["name"]:
+            if scope.action in access["actions"]:
+                return True
+
+    return False
```
mdellweg, line 194: I think, i have seen this function before. Maybe you just moved it around, maybe you changed it on the way. Can you move it back, so it's easier to review?

pulp_container/app/token_verification.py
```diff
@@ -101,7 +59,7 @@ def authenticate(self, request):
             return None
         token = authorization_header[len(self.keyword) + 1 :]
         try:
-            decoded_token = _decode_token(token, request)
+            decoded_token = decode_token(token, request)
```
mdellweg, line 62: What's wrong about hiding that function?

mdellweg (changes requested):
This pr is still touching a lot more lines than needed.
If you think this file is particularly bad structured, try splitting feature and refactoring PR's.

pulp_container/app/token_verification.py
goosemania: Alternatively, one can just name a variable that way.
Comments tend to get out of date, naming a boolean condition makes it self-explanatory but also doesn't require creating a function.

```
is_requesting_root_endpoint = len(decoded_token["access"]) == 0
```

ipanova (approved):

pulp_container/app/token_verification.py
```diff
@@ -9,6 +11,7 @@
 from rest_framework.exceptions import AuthenticationFailed, NotAuthenticated
 from rest_framework.permissions import BasePermission, SAFE_METHODS
 
+Scope = namedtuple("Scope", "resource_type, name, action")
```
mdellweg, line 14:
```suggestion
Scope = namedtuple("Scope", ["resource_type", "name", "action"])
```

Not blocking this, if it works, but it looks weird to me.

mdellweg (approved):

pulp_container/app/token_verification.py
lubosmj: @mdellweg, I was inspired by https://github.com/pulp/pulp_container/pull/217#discussion_r571015874. And I liked it.

PR #244, opened 2021-02-17
Allow user to add a namespace matching their login

mdellweg:
This is a change to the default access policy accompanied by a helper
function `namespace_is_username` for  access policies.

fixes #8197
https://pulp.plan.io/issues/8197

### DIFF
--- a/CHANGES/8197.feature
+++ b/CHANGES/8197.feature
@@ -1,0 +1,2 @@
+Added a `namespace_is_username` helper to decide whether the namespace matches the username of the requests user.
+Changed the namespace access_policy to allow users without permissions to create the namespace that matches their username.
--- a/pulp_container/app/access_policy.py
+++ b/pulp_container/app/access_policy.py
@@ -6,7 +6,7 @@ from pulpcore.plugin.access_policy import AccessPolicyFromDB
 from pulp_container.app import models
 
 
-class NamespaceAccessPolicyMixin:
+class NamespacedAccessPolicyMixin:
     """
     Access policy mixin for ContainerDistributionViewSet and ContainerPushRepositoryViewSet which
     handles namespace permissions.
@@ -21,7 +21,7 @@ class NamespaceAccessPolicyMixin:
         if type(obj) == models.ContainerDistribution:
             namespace = obj.namespace
             return request.user.has_perm(permission, namespace)
-        else:
+        elif type(obj) == models.ContainerPushRepository:
             dists_qs = models.ContainerDistribution.objects.filter(repository=obj)
             for dist in dists_qs:
                 if request.user.has_perm(permission, dist.namespace):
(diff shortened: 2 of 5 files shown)


pulp_container/app/viewsets.py
```diff
@@ -1160,6 +1160,12 @@ class ContainerNamespaceViewSet(
                 "effect": "allow",
                 "condition": "has_model_perms:container.add_containernamespace",
             },
+            {
+                "action": ["create"],
```
ipanova, line 1164: we need to talk through what needs to be done whenever access policy changes, probably nothing for additional changes?

pulp_container/app/access_policy.py
```diff
@@ -51,14 +54,25 @@ def is_private(self, request, view, action):
...
     namespace permissions.
     """
 
+class NamespaceAccessPolicy(AccessPolicyFromDB):
+    """
+    Access policy for ContainerNamespaceViewSet.
+    """
 
-class RegistryAccessPolicy(AccessPolicy, NamespaceAccessPolicyMixin):
+    def namespace_is_username(self, request, view, action):
```
ipanova, line 68: have not fully loaded this into my head, buy can't we have this part of the `NamespaceAccessPolicyMixin` and NamespaceViewset would just use the `NamespaceAccessPolicyFromDB`?

pulp_container/app/viewsets.py
mdellweg: Well, i believe, as long as the access policy is marked "unchanged by the user" we would just overwrite it. But yes the code for that is not yet in place.

pulp_container/app/access_policy.py
mdellweg: Good question. It's a helper function that only works without exploding when used in an access policy for Namespaces.
And atm NamespaceAccessPolicyFromDB is used only for stuff that is not a namespace. And most of it's helper functions will blow up on namespaces.
That is why i think it should be called "Namespace__d__AccessPolicy".

Having said that, i fail to see, why i introduced:
https://github.com/pulp/pulp_container/pull/244/files#diff-bada7270b35add8957655a936b1bdb701c4dc829033fd87467735943b839c354R29-R31

pulp_container/app/access_policy.py
```diff
@@ -51,14 +51,26 @@ def is_private(self, request, view, action):
...
+class NamespaceAccessPolicy(AccessPolicyFromDB):
+    """
+    Access policy for ContainerNamespaceViewSet.
+    """
+
+    def namespace_is_username(self, request, view, action):
+        """
+        Check if the namespace in the request matches the username.
+        """
+        return request.data["name"] == request.user.username
```
mdellweg, line 70: use get

pulp_container/app/access_policy.py
ipanova: So i have been thinking some and I agree with you, this would have saved some lines of code and we'd have defined `'namespace_is_username` in one place however then  the Mixin would contain all the stuff for repos, distributions and namespaces and this seems to be too entangled. Better to have a separate cleaner policy purely for Namespaces as you did.

CHANGES/8197.feature
```diff
@@ -0,0 +1,2 @@
+Added a `namespace_is_username` helper to decide whether the namespace matches the username of the requests user.
```
ipanova, line 1: please update the  docs this is mentioned there as well https://docs.pulpproject.org/pulp_container/role-based-access-control.html#namespaces

pulp_container/app/access_policy.py
ipanova: do what @mdellweg  said ^^ :D

ipanova:
@mdellweg  i have tested and it all works well. Can you please add a test? 1. push with matcing login without any perms 2. push with not matching login without any perms.

PR #245, opened 2021-02-17
Change progress bar status codes to include the task type/stage

dralley:
Also use convenience functions for pulp types.

[noissue]

### DIFF
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -63,7 +63,7 @@ class ContainerFirstStage(Stage):
         total_blobs = []
 
         with ProgressReport(
-            message="Downloading tag list", code="downloading.tag_list", total=1
+            message="Downloading tag list", code="sync.downloading.tag_list", total=1
         ) as pb:
             repo_name = self.remote.namespaced_upstream_name
             relative_url = "/v2/{name}/tags/list".format(name=repo_name)
@@ -91,7 +91,7 @@ class ContainerFirstStage(Stage):
 
         pb_parsed_tags = ProgressReport(
             message="Processing Tags",
-            code="processing.tag",
+            code="sync.processing.tag",
             state=TASK_STATES.RUNNING,
             total=len(tag_list),
         )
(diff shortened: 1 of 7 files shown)


pulp_container/app/tasks/sync_stages.py
```diff
@@ -63,7 +63,7 @@ async def run(self):
         total_blobs = []
 
         with ProgressReport(
-            message="Downloading tag list", code="downloading.tag_list", total=1
+            message="Downloading tag list", code="sync.downloading.tag_list", total=1
```
dralley, line 66: This helps keep the stages separate in combined sync/publish tasks.  Container plugin doesn't publish but it would be good to follow the same pattern.

pulp_container/app/tasks/synchronize.py
```diff
@@ -41,7 +41,7 @@ def synchronize(remote_pk, repository_pk, mirror):
     log.info(_("Synchronizing: repository={r} remote={p}").format(r=repository.name, p=remote.name))
     first_stage = ContainerFirstStage(remote)
     dv = ContainerDeclarativeVersion(first_stage, repository, mirror)
-    dv.create()
+    return dv.create()
```
mdellweg, line 44: Is this a bugfix? Is rq expecting us to return something?

dralley: No bugfix, just following the same patterns that we're using elsewhere now due to the auto-publish changes.

pulp_container/app/tasks/sync_stages.py
ipanova: there is one more `code` to update on line 94, can you update it as well?

ipanova: nvm, i am blind :D

pulp_container/app/tasks/untag.py
```diff
@@ -9,7 +9,7 @@ def untag_image(tag, repository_pk):
     repository = Repository.objects.get(pk=repository_pk).cast()
     latest_version = repository.latest_version()
 
-    tags_in_latest_repository = latest_version.content.filter(pulp_type="container.tag")
+    tags_in_latest_repository = latest_version.content.filter(pulp_type=Tag.get_pulp_type())
```
ipanova, line 13: git grep shows that there are some harcoded types in the viewsets.py can you update those also?

dralley: yup

mdellweg (commented):
I'm fine with returning a value from a function, that is there anyway, But deriving functions just to call `super`, seems like a waste of resources on all ends.

mdellweg (approved):

PR #247, opened 2021-02-19
Add an endpoint for removing images recursively by digests

lubosmj:
closes #8105

### DIFF
--- a/CHANGES/8105.feature
+++ b/CHANGES/8105.feature
@@ -1,0 +1,1 @@
+Added a new API endpoint that allows users to remove an image by a digest from a push repository.
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -335,7 +335,11 @@ class TagOperationSerializer(serializers.Serializer):
         latest_version = new_data["repository"].latest_version()
         if not latest_version:
             raise serializers.ValidationError(
-                _("The latest repository version of '{}' was not found".format(data["repository"]))
+                _(
+                    "The latest repository version of '{}' was not found".format(
+                        new_data["repository"]
+                    )
+                )
             )
 
         new_data["latest_version"] = latest_version
@@ -503,6 +507,56 @@ class ManifestCopySerializer(CopySerializer):
         required=False,
         help_text="A list of media_types to copy.",
     )
+
+
+class RemoveImageSerializer(serializers.Serializer):
+    """
+    A serializer for parsing and validating data associated with the image removal.
+    """
+
+    digest = serializers.CharField(help_text="sha256 of the Manifest file")
+
+    def validate(self, data):
+        """
+        Validate and extract the latest repository version, manifest, and tags from the passed data.
+        """
+        new_data = {}
+        new_data.update(self.initial_data)
+
+        latest_version = new_data["repository"].latest_version()
+        if not latest_version:
+            raise serializers.ValidationError(
+                _(
+                    "The latest repository version of '{}' was not found".format(
+                        new_data["repository"]
+                    )
+                )
+            )
+
+        new_data["latest_version"] = latest_version
+
+        try:
+            manifest = models.Manifest.objects.get(
+                pk__in=new_data["latest_version"].content.all(), digest=new_data["digest"]
+            )
+        except models.Manifest.DoesNotExist:
+            raise serializers.ValidationError(
+                _(
+                    "A manifest with the digest '{}' does not "
+                    "exist in the latest repository version '{}'".format(
+                        new_data["digest"], new_data["latest_version"]
+                    )
+                )
+            )
+
+        new_data["manifest"] = manifest
+
+        tags_pks = models.Tag.objects.filter(
+            pk__in=new_data["latest_version"].content.all(), tagged_manifest=new_data["manifest"]
+        ).values_list("pk", flat=True)
+        new_data["tags_pks"] = tags_pks
+
+        return new_data
 
 
 class OCIBuildImageSerializer(serializers.Serializer):
--- a/pulp_container/app/tasks/recursive_remove.py
+++ b/pulp_container/app/tasks/recursive_remove.py
@@ -1,9 +1,8 @@
 from django.db.models import Q
-from pulpcore.plugin.models import Content
+from pulpcore.plugin.models import Content, Repository
 
 from pulp_container.app.models import (
     Blob,
-    ContainerRepository,
     Manifest,
     MEDIA_TYPE,
     Tag,
@@ -35,7 +34,7 @@ def recursive_remove_content(repository_pk, content_units):
             should be removed from the Repository.
 
     """
-    repository = ContainerRepository.objects.get(pk=repository_pk)
+    repository = Repository.objects.get(pk=repository_pk).cast()
     latest_version = repository.latest_version()
     latest_content = latest_version.content.all() if latest_version else Content.objects.none()
     if "*" in content_units:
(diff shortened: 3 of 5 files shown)


lubosmj:
@ipanova, @mdellweg, is it sufficient enough to lock just a repository while performing the removal task? Is it always guaranteed that all objects within the repository (i.e., manifests, blobs, tags) are going to be locked as well? If so, has the locking of all the objects any negative impact on the performance?

ipanova:
> @ipanova, @mdellweg, is it sufficient enough to lock just a repository while performing the removal task? Is it always guaranteed that all objects within the repository (i.e., manifests, blobs, tags) are going to be locked as well? If so, has the locking of all the objects any negative impact on the performance?

we can't lock on the content units for the performance reasons. lock on the repo means that no other action can be performed on it until you finish your current task. So if you have locked the repo for a removal operation, another removal/add operation won't be executed in parallel, it will wait until the lock will be released, so tldr no content will disappear from the repo during the removal rask in course.

CHANGES/8105.feature
```diff
@@ -0,0 +1 @@
+Added a new API endpoint that allows users to remove an imange by a digest within a push repository.
```
ipanova, line 1: s/imange/image

ipanova: s/within a push repository/from a push repository

pulp_container/app/tasks/recursive_remove.py
```diff
@@ -1,15 +1,28 @@
...
 from pulp_container.app.models import (
     Blob,
-    ContainerRepository,
     Manifest,
     MEDIA_TYPE,
     Tag,
 )
 
 
+def recursive_remove_content_find_tags(repository_pk, manifest_pk):
```
ipanova, line 12: we need to add blobs here as well. And manifests in case what you are trying to remove by digest happens to be a manifest_list. Now the trick is to remove only those blobs that are not referenced by any other image within that repo; remove only those image manifests that are not referenced in other manifest lists within that repo and those that are not tagged. All this logic is respected in the task `recursive_remove_content`.
I suggest using it, just in the `content_unit` already provide pks of the manifest/manifest-list and pks of all tags you will find referenced by that object.

ipanova: you can find all the pks i was talking about in the viewet and pass it to the task  from the viewset.

pulp_container/app/serializers.py
```diff
@@ -505,6 +508,38 @@ class ManifestCopySerializer(CopySerializer):
...
+    digest = serializers.CharField(help_text="sha256 of the Manifest file")
+
+    def validate(self, data):
+        """
+        Validate and extract the latest repository version, manifest, and tags from the passed data.
+        """
+        new_data = {}
+        new_data.update(self.initial_data)
+
+        latest_version = new_data["repository"].latest_version()
```
ipanova, line 525: if you subclass it from TagOperationSerializer you would not have to have this check.

ipanova: not really, TagOperationSerializer has tag field, disregard this comment.

pulp_container/app/serializers.py
```diff
@@ -358,23 +358,26 @@ def validate(self, data):
...
-            )
-        except models.Manifest.DoesNotExist:
-            raise serializers.ValidationError(
-                _(
-                    "A manifest with the digest '{}' does not "
-                    "exist in the latest repository version '{}'".format(
-                        new_data["digest"], new_data["latest_version"]
-                    )
+
+def get_manifest(data):
```
ipanova, line 365: extracting the repeated code into a separate function and leaving it like this in between all other serializers' code makes it hard to read because it gets 'lost', you need to logically place it.
you can define a mixin that would validate repo_version and manifest or you can just let it be and leave this try/excpet repeated twice. Sometimes it is more readable to have the code repeated rather then adding mixins and have multi-inheritance for the sake of saving several lines of code.

pulp_container/tests/functional/api/test_recursive_remove.py
```diff
@@ -372,3 +382,81 @@ def test_cannot_remove_tagged_manifest(self):
...
+        monitor_task(tag_response.task)
+
+        latest_version_href = cls.repositories_api.read(cls.repo.pulp_href).latest_version_href
+        cls.content_to_remove = cls.versions_api.read(latest_version_href).content_summary.present
+
+    @classmethod
+    def tearDownClass(cls):
+        """Delete the created distribution."""
+        cls.distributions_api.delete(cls.distribution.pulp_href)
+        cls.namespaces_api.delete(cls.namespace.pulp_href)
```
ipanova, line 435: there is no need to delete distribution, it gets removed automatically with the namespace.

pulp_container/app/serializers.py
```diff
@@ -505,6 +508,38 @@ class ManifestCopySerializer(CopySerializer):
...
+    """
+
+    digest = serializers.CharField(help_text="sha256 of the Manifest file")
+
+    def validate(self, data):
+        """
+        Validate and extract the latest repository version, manifest, and tags from the passed data.
+        """
+        new_data = {}
+        new_data.update(self.initial_data)
```
mdellweg, line 523: What is happening here?
Should this not be `new_data = super().validate(data)`?
Does this circumvent the field validations?
Why do we refer to a repository here, and there is no repository field?

pulp_container/tests/functional/api/test_recursive_remove.py
```diff
@@ -372,3 +382,81 @@ def test_cannot_remove_tagged_manifest(self):
...
+        """Remove the 'manifest_a' image along with the related blobs, manifest, and tags."""
+        remove_response = self.repositories_api.remove_image(
+            self.repo.pulp_href, RemoveImage(digest=self.manifest_a.digest)
+        )
+        monitor_task(remove_response.task)
+
+        latest_version_href = self.repositories_api.read(self.repo.pulp_href).latest_version_href
+        content_summary = self.versions_api.read(latest_version_href).content_summary
+
+        self.assertEqual(content_summary.present, {})
```
ipanova, line 448: nice! :sun_with_face:

pulp_container/app/serializers.py
mdellweg: I am wondering if this Manifest lookup can be rephrased with a HiddenField, as well as the latest_version [0]. But that would be refactoring and we should not do that along with a feature PR. I agree with @ipanova that sometimes keeping the things where they happen to not have to scoll up and down to understand makes reading easier.
At the very least i am quite unhappy with the function name, because in django all the "get_<an object>" functions raise `DoesNotExist` and this one will raise `ValidationError`.

[0] https://www.django-rest-framework.org/api-guide/fields/#hiddenfield

pulp_container/app/serializers.py
lubosmj: There is no need to call the base `validate()` method because it does literally nothing. The field for a repository is initialized automatically. We already use such a technique here: https://github.com/pulp/pulp_container/blob/8274e1ed1b83ccfe2c7eec1294582dfb9d8a71ea/pulp_container/app/serializers.py#L324-L335

It does not circumvent the validation because of this:
https://github.com/pulp/pulp_container/pull/247/files#diff-31c44b2812e50f3171248db027f439e8e64248cb84067fc7c305de74cdb9a86fR875-R880

ipanova (approved):

pulp_container/tests/functional/api/test_recursive_remove.py
lubosmj: The last force-push is resolving this comment.

PR #248, opened 2021-02-23
Remove the note stating the outdated behaviour

lubosmj:
[noissue]

### DIFF
--- a/docs/workflows/push.rst
+++ b/docs/workflows/push.rst
@@ -3,16 +3,21 @@
 Push content to a Repository
 =============================
 
-Users can push a container image to the repositories hosted by the Container Registry
+Users can push images to the repositories hosted by the Container Registry. Only the users who are
+logged in to the registry are allowed to perform such an action. Find below a complete example of
+pushing a tagged image.
 
 .. note::
-   If token auth is enabled admin credentials will be required during the push operation.
-   Provide them in the login to the registry or in each API call.
+    Having disabled the token authentication, only users with staff privileges (i.e.,
+    administrators) are allowed to push content to the registry.
+
 
 ::
 
         $ podman tag d21d863f69b5 localhost:24817/test/this:mytag1.8
-        $ push d21d863f69b5 localhost:24817/test/this:mytag1.8
+        $ podman login -u user -p password localhost:24817
+          Login Succeeded!
+        $ podman push d21d863f69b5 localhost:24817/test/this:mytag1.8
           Getting image source signatures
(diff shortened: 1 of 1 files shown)


docs/workflows/push.rst
```diff
@@ -3,16 +3,16 @@
...
 =============================
 
-Users can push a container image to the repositories hosted by the Container Registry
-
-.. note::
-   If token auth is enabled admin credentials will be required during the push operation.
-   Provide them in the login to the registry or in each API call.
+Users can push images to the repositories hosted by the Container Registry. Only the users who are
+logged in to the registry are allowed to perform such an action. Find below a complete example of
+pushing a tagged image.
```
ipanova, line 8: @lubosmj  can you add a note that will say that if you have toke-auth disabled only admin user can push.
i am not sure where we say that in some other place?

docs/workflows/push.rst
```diff
@@ -3,16 +3,16 @@
...
 
 ::
 
         $ podman tag d21d863f69b5 localhost:24817/test/this:mytag1.8
-        $ push d21d863f69b5 localhost:24817/test/this:mytag1.8
+        $ podman login -u admin -p password localhost:24817
```
ipanova, line 13: how about `$ podman login -u user -p password localhost:24817`

ipanova:
@lubosmj since you're updated outdated docs, please update the note at the botton about push repos and removal of content via pulp api https://github.com/pulp/pulp_container/blob/master/docs/workflows/push.rst

docs/workflows/push.rst
lubosmj: Oh, I am sorry. Other users will be allowed to push images after resolving the story https://pulp.plan.io/issues/8271, right?

ipanova: right, for now we have enabled only admin to push if you have disabled auth.

docs/workflows/push.rst
```diff
@@ -3,16 +3,21 @@
...
 
-Users can push a container image to the repositories hosted by the Container Registry
+Users can push images to the repositories hosted by the Container Registry. Only the users who are
+logged in to the registry are allowed to perform such an action. Find below a complete example of
+pushing a tagged image.
 
 .. note::
-   If token auth is enabled admin credentials will be required during the push operation.
-   Provide them in the login to the registry or in each API call.
+    Having disabled the token authentication, only users with staff privileges (i.e.,
```
ipanova, line 11: :+1:

ipanova (approved):

PR #250, opened 2021-02-24
Download config blob regardless of the remote's settings.

ipanova:
[noissue]

### DIFF
--- a/CHANGES/8319.feature
+++ b/CHANGES/8319.feature
@@ -1,0 +1,1 @@
+Config blob is downloaded always, regardless of the remote's settings.
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -195,7 +195,7 @@ class ContainerFirstStage(Stage):
             total_blobs.append(blob_dc)
         layer = content_data.get("config", None)
         if layer:
-            blob_dc = self.create_blob(man, layer)
+            blob_dc = self.create_blob(man, layer, deferred_download=False)
             blob_dc.extra_data["config_relation"] = man
             total_blobs.append(blob_dc)
 
@@ -309,13 +309,15 @@ class ContainerFirstStage(Stage):
         )
         return man_dc
 
-    def create_blob(self, man_dc, blob_data):
+    def create_blob(self, man_dc, blob_data, deferred_download=True):
         """
         Create blob.
 
         Args:
             man_dc (pulpcore.plugin.stages.DeclarativeContent): dc for a ImageManifest
             blob_data (dict): Data about a blob
+            deferred_download (bool): boolean that indicates whether not to download a blob
+                immediatly. Config blob is downloaded regardless of the remote's settings
 
         """
         digest = blob_data.get("digest") or blob_data.get("blobSum")
@@ -330,7 +332,7 @@ class ContainerFirstStage(Stage):
             url=blob_url,
             relative_path=digest,
             remote=self.remote,
-            deferred_download=self.deferred_download,
+            deferred_download=deferred_download and self.deferred_download,
         )
         blob_dc = DeclarativeContent(content=blob, d_artifacts=[da])
 


pulp_container/app/tasks/sync_stages.py
```diff
@@ -190,7 +190,7 @@ def handle_blobs(self, man, content_data, total_blobs):
         for layer in content_data.get("layers") or content_data.get("fsLayers"):
             if not self._include_layer(layer):
                 continue
-            blob_dc = self.create_blob(man, layer)
+            blob_dc = self.create_blob(man, layer, self.deferred_download)
```
mdellweg, line 193: It's kind of weird, that the special handling is performed at the normal blob, and not the special config blob.
How about you specify `deferred_download=False` explicitly for config blobs?
I'll try to outline the rest of this solution in suggestions:
```suggestion
            blob_dc = self.create_blob(man, layer)
```

pulp_container/app/tasks/sync_stages.py
```diff
@@ -309,13 +309,15 @@ def create_manifest(self, list_dc, manifest_data):
         )
         return man_dc
 
-    def create_blob(self, man_dc, blob_data):
+    def create_blob(self, man_dc, blob_data, deferred_download=False):
```
mdellweg, line 312:
```suggestion
    def create_blob(self, man_dc, blob_data, deferred_download=True):
```

pulp_container/app/tasks/sync_stages.py
```diff
@@ -309,13 +309,15 @@ def create_manifest(self, list_dc, manifest_data):
...
         """
         Create blob.
 
         Args:
             man_dc (pulpcore.plugin.stages.DeclarativeContent): dc for a ImageManifest
             blob_data (dict): Data about a blob
+            deferred_download (bool): boolean that indicates whether to download blob.
+                Config blob is downloaded regardless of the remote's settings
```
mdellweg, line 320:
```suggestion
            deferred_download (bool): boolean that indicates whether not to download a blob immediately.
                Config blobs should be downloaded regardless of the remote's settings.
```

pulp_container/app/tasks/sync_stages.py
```diff
@@ -330,7 +332,7 @@ def create_blob(self, man_dc, blob_data):
             url=blob_url,
             relative_path=digest,
             remote=self.remote,
-            deferred_download=self.deferred_download,
+            deferred_download=deferred_download,
```
mdellweg, line 335:
```suggestion
            deferred_download=deferred_download and self.deferred_download,
```

pulp_container/app/tasks/sync_stages.py
mdellweg: I cannot add the corresponding suggestion to line 198

ipanova: special handling is left for normal blobs because the user can actually make a call whether he wants it downloaded or not. He can control that by specifying the download policy in the remote. Whilst config blob he cannot affect in any case because it will always be downloaded.
I can put another the boolean evaluation into the create_blob if that improves readability.

mdellweg: In the Pulp mindset, i consider honoring the download_policy to be 'normal'. Therefore overwriting it for certain content classes would be the special case.

CHANGES/8319.feature
```diff
@@ -0,0 +1 @@
+Config blob is downloaded always, regardless of the remote's settings.
```
mdellweg, line 1:
```suggestion
Config blobs are always downloaded, regardless of the remote's settings.
```

pulp_container/app/tasks/sync_stages.py
```diff
@@ -309,13 +309,15 @@ def create_manifest(self, list_dc, manifest_data):
...
-    def create_blob(self, man_dc, blob_data):
+    def create_blob(self, man_dc, blob_data, deferred_download=True):
         """
         Create blob.
 
         Args:
             man_dc (pulpcore.plugin.stages.DeclarativeContent): dc for a ImageManifest
             blob_data (dict): Data about a blob
+            deferred_download (bool): boolean that indicates whether not to download a blob
+                immediatly. Config blob is downloaded regardless of the remote's settings
```
mdellweg, line 320:
```suggestion
            deferred_download (bool): boolean to indicate whether to honor the remotes
                download_policy (True) or download artifacts immediately anyway (False).
                i.e. config blobs are downloaded regardless of the remote's settings.
```

mdellweg (approved):
Looks good to me.
I just added suggestions that would express the intent better to me personally.
Feel free to ignore them.

PR #252, opened 2021-03-01
Return an empty string when the key does not exist

lubosmj:
In this commit, there was also fixed the output format of the field `Cmd`.

In the past, we had been returning:
\"container_config\":{\"Cmd\":\"/bin/sh -c #(nop) ADD file:ab8940a9b8a62d2f2c4634bf5e63a87ea73736bd0313412e41be2351c4fbdf53 in / \"}.

As of now, the command is wrapped by a list:
\"container_config\":{\"Cmd\":[\"/bin/sh -c #(nop) ADD file:ab8940a9b8a62d2f2c4634bf5e63a87ea73736bd0313412e41be2351c4fbdf53 in / \"]}

closes #8299

### DIFF
--- a/CHANGES/8299.bugfix
+++ b/CHANGES/8299.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug that caused the registry to fail during the schema conversion when there was not
+provided the field ``created_by``.
--- a/pulp_container/app/schema_convert.py
+++ b/pulp_container/app/schema_convert.py
@@ -193,9 +193,12 @@ class Schema2toSchema1Converter:
             config.pop("history", None)
             config.pop("rootfs", None)
         else:
+            # both `created` and `created_by` are optional according to the OCI specs
+            container_config = dict(Cmd=[fs_layer.history.get("created_by", "")])
+            created = fs_layer.history.get("created", "")
             config = dict(
-                created=fs_layer.history["created"],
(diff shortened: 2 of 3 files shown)


lubosmj:
"v1Compatibility": "{\"container_config\":{\"Cmd\":[\"\"]},\"created\":\"2019-04-16T15:35:01.957134Z\",\"id\":\"91dcdb7a26b021728e2c3f959e3bffd8c4063b939d37f8e140fe16443e03c2f2\",\"parent\":\"be037f28143aedd8e7b3554c65fd3ea791e9f97383260d5d98d03f54cf968ca9\"}"

ipanova:
@lubosmj why have you decided to leave an empty string in the array of Cmd? can we just do `Cmd:[]` ?
Besides created_by fix i have noticed that `history` and `created` are also optional and we are accessing them by key. Please fix the rest of occurences when we access optional config by key and switch it to .get()

lubosmj:
@ipanova, dockerhub returns an empty string in an array when there is no `created_by` command present in a manifest. Also, according to the go code here https://github.com/distribution/distribution/blob/22c074842eaad74036164f95b2182a9d0fe484b5/manifest/schema1/config_builder.go#L150, the value of `created_by` defaults to an empty string during the assignment regardless `omitempty`, which is considered only when parsing JSON data.

ipanova (approved):
thanks!

PR #253, opened 2021-03-02
Sanitize page size in pagination classes

mdellweg:
This will prevent the request to result in a 500 response when a negative
page size is requested.

fixes #8318
https://pulp.plan.io/issues/8318

### DIFF
--- a/CHANGES/8318.bugfix
+++ b/CHANGES/8318.bugfix
@@ -1,0 +1,1 @@
+Prevent the registry pagination classes to fail if a negative page size is requested.
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -416,6 +416,8 @@ class ContainerCatalogPagination(BasePagination):
         else:
             if self.n > 10 * api_settings.PAGE_SIZE:
                 self.n = 10 * api_settings.PAGE_SIZE
+            if self.n < 0:
+                self.n = 0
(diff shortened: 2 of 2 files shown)


ipanova:
I am not sure this will work, have you tested these changes?
It will most likely fail on index out of range here `repositories_names[-1]` because `len(repositories_names) == self.n:` will match the condition.
```
def get_paginated_response(self, data):
        """
        Prepare the paginated container _catalog response.
        """
        headers = {}
        repositories_names = [repo["base_path"] for repo in data]
        if len(repositories_names) == self.n:
            # There's a high chance we haven't gotten all entries here.
            parsed_url = list(urlparse(self.url))
            query_params = parse_qs(parsed_url[4])
            query_params["n"] = str(self.n)
            query_params["last"] = repositories_names[-1]
            parsed_url[4] = urlencode(query_params)
            url = urlunparse(parsed_url)
            headers["Link"] = f'<{url}>; rel="next"'
        return Response(headers=headers, data={"repositories": repositories_names})
```

mdellweg:
Right. I fixed it.

ipanova:
I would suggest to return a empty list in case there was specified and a value of n <=0 if possible.
Maybe add a check in `get_paginated_response` on value of self.n and if not a positive integer return the `Response`?

ipanova (approved):

PR #259, opened 2021-03-05
Add 429 polling to BlobUpload commit

mdellweg:
Opening this to have a place to discuss the current attempt to the solution.

Thinks to keep in mind / discuss:
- I believe this works as designed for blob upload. There we are in the lucky position, that the Upload pk is a unique id that can only be used once.
- With tagging manifests, this will be more complicated.
  - It would be nice to be able to calculate some identifier of a request that will be preserved after a 429 retry. This would solve the problem at hand in a more general way (It would basically be task polling written in HTTP language).
- Immediately waiting for the task <=1s before sending the first 429 is not yet implemented (mostly to show that the 429 mechanism actually works).
- Is it OK to insert an arbitrary string as a task_reserved_resource here?

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -9,6 +9,8 @@ import logging
 import hashlib
 import re
 from collections import namedtuple
+import time
+
 from urllib.parse import urlparse, urlunparse, parse_qs, urlencode
 from tempfile import NamedTemporaryFile
 
@@ -21,14 +23,16 @@ from django.conf import settings
 
 from guardian.shortcuts import get_objects_for_user
 
-from pulpcore.plugin.models import Artifact, ContentArtifact, UploadChunk
+from pulpcore.plugin.models import Artifact, ContentArtifact, Task, UploadChunk
 from pulpcore.plugin.files import PulpTemporaryUploadedFile
+from pulpcore.plugin.tasking import add_and_remove, enqueue_with_reservation
 from rest_framework.exceptions import (
     AuthenticationFailed,
     NotAuthenticated,
     PermissionDenied,
     NotFound,
     ParseError,
+    Throttled,
     ValidationError,
 )
 from rest_framework.generics import ListAPIView
@@ -205,7 +209,6 @@ class BlobResponse(Response):
         artifact = blob._artifacts.get()
         size = artifact.size
 
-        log.info("digest: {digest}".format(digest=blob.digest))
         headers = {
             "Docker-Content-Digest": blob.digest,
             "Location": "/v2/{path}/blobs/{digest}".format(path=path, digest=blob.digest),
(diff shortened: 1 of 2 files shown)


pulp_container/app/registry_api.py
```diff
@@ -609,12 +636,21 @@ def put(self, request, path, pk=None):
...
 
-            with repository.new_version() as new_version:
-                new_version.add_content(models.Blob.objects.filter(pk=blob.pk))
-
             upload.delete()
 
-            return BlobResponse(blob, path, 201, request)
+            task = enqueue_with_reservation(
+                add_and_remove,
+                [f"upload:{pk}", repository, blob],
```
mdellweg, line 643: Is it appropriate to include blob here? Or do we never lock on Content units?

ipanova: imo, we should never lock on content units. although in the 'tag' task we are locking on manifest and we should not.

mdellweg:
cc @bmbouter

pulp_container/app/registry_api.py
```diff
@@ -743,14 +787,34 @@ def put(self, request, path, pk=None):
...
-            )
-        return ManifestResponse(manifest, path, request, status=201)
+            tag = models.Tag.objects.get(name=tag.name, tagged_manifest=manifest)
+
+        tags_to_remove = models.Tag.objects.filter(
+            pk__in=repository.latest_version().content.all(), name=tag
+        ).exclude(tagged_manifest=manifest)
+        task = enqueue_with_reservation(
+            add_and_remove,
+            [repository, blob],
```
ipanova, line 797: the lock should happen only on the repo.

pulp_container/app/registry_api.py
```diff
@@ -743,14 +787,34 @@ def put(self, request, path, pk=None):
...
+            time.sleep(1)
+            task = Task.objects.get(pk=task.id)
+            if task.state == "completed":
+                task.delete()
+                return ManifestResponse(manifest, path, request, status=201)
+            elif task.state in ["waiting", "running"]:
+                continue
+            else:
+                task.delete()
+                raise Exception("Failed.")
```
ipanova, line 816: I think we should delete the task before returning 429 regardless of it's state. Neither podman or docker retry during manifest put. Also they don't issue head before put so I don't see the reason of why to keep the task?

pulp_container/app/registry_api.py
```diff
@@ -584,7 +587,29 @@ def put(self, request, path, pk=None):
         _, repository = self.get_dr_push(request, path)
 
         digest = request.query_params["digest"]
-        upload = models.Upload.objects.get(pk=pk, repository=repository)
+        try:
```
ipanova, line 590: please add some comments around what's it happening here. this will help us whenever we would look into this code in X months

ipanova (commented):
the code looks good! i am going to hand test a bit

pulp_container/app/registry_api.py
mdellweg: OK. I will add an attempt to cancel it and then do another finished check and delete it.

pulp_container/app/registry_api.py
```diff
@@ -613,12 +645,31 @@ def put(self, request, path, pk=None):
             except IntegrityError:
                 pass
 
-            with repository.new_version() as new_version:
-                new_version.add_content(models.Blob.objects.filter(pk=blob.pk))
-
             upload.delete()
 
-            return BlobResponse(blob, path, 201, request)
+            task = enqueue_with_reservation(
```
ipanova, line 650: i get occasionally this error `AttributeError: 'Task' object has no attribute 'id'`
we should not use `task` as variable here, but rather `job` because on line 663 `task` variable holds a task object and not Job, and so if hitting continue statement on line 668 will make us looping through the cycle and we'd fail on line 663

mdellweg: Ha! So you are saying, if the thing takes longer than the first second, we start to hit this?

ipanova (approved):
i am going to approve these changes I have tested them and they do what supposed to. Thank you for working on this!

PR #261, opened 2021-03-08
Update CI from template

mdellweg:
Using plugin_template tag 2021.02.04

[noissue]


ipanova:
@mdellweg based on the ci failure i believe this is the commit that would fix it https://github.com/pulp/pulp_container/pull/221/files

PR #263, opened 2021-03-09
Pre release 2.1.1

mdellweg:


mdellweg:
Looks like we need to merge this first before considering the changelog generation commit.

PR #265, opened 2021-03-09
Filter repositories based on assigned permissions

lubosmj:


lubosmj:
I am not sure why the tests are failing here because they are passing on my local machine...

pulp_container/app/registry_api.py
```diff
@@ -449,6 +452,19 @@ class CatalogView(ContainerRegistryApiMixin, ListAPIView):
...
+        """Filter the queryset based on assigned permissions."""
+        queryset = super().get_queryset()
+
+        accessible_repo_pks = []
+        for obj in queryset:
+            view = FakeView("pull", lambda: obj)
+            if self.access_policy_class.has_permission(self.request, view):
+                accessible_repo_pks.append(obj.pk)
+
+        return queryset.filter(pk__in=accessible_repo_pks)
```
ipanova, line 467: you probably want to order the results alphabetically or by date, so your results are deterministic. This is also can be the reason of the failing tests, when 2 lists are compared items seem to have different sequence

ipanova: results get ordered in the pagination class, no need to order

pulp_container/app/registry_api.py
```diff
@@ -449,6 +452,19 @@ class CatalogView(ContainerRegistryApiMixin, ListAPIView):
...
     serializer_class = ContainerCatalogSerializer
     pagination_class = ContainerCatalogPagination
+    access_policy_class = RegistryAccessPolicy()
+
+    def get_queryset(self, *args, **kwargs):
+        """Filter the queryset based on assigned permissions."""
+        queryset = super().get_queryset()
+
+        accessible_repo_pks = []
+        for obj in queryset:
```
mdellweg, line 462: This is looping over the complete queryset. We should not do that, but extend the queryset with rules that can be translated into SQL.

pulp_container/tests/functional/utils.py
```diff
@@ -52,18 +52,20 @@
...
-    "        perm = Permission.objects.get(codename=codename, content_type__app_label=app_label)",
-    "    else:",
-    "        perm = Permission.objects.get(codename=permission)",
-    "    user.user_permissions.add(perm)",
+    "",
+    "for permission in {model_permissions!r}:",
+    "    assign_perm(permission, user)",
+    "",
+    "for permission, obj_name in {object_permissions!r}:",
+    "    obj = ContainerDistribution.objects.filter(name=obj_name)",
```
mdellweg, line 67: Can't we have a more general way to address objects?
I tried to design this piece of code in a way, that we can move it to pulp-smash at some time, but the dependence on specific container models would prevent that.

Now that I am writing this, i realize, that we also do not follow that goal with the way we provide the Api's here. This might need more thought.

pulp_container/app/registry_api.py
mdellweg: Discussion of the container meeting:
We should use `get_objects_for_user` and filter for `public=True`.
This will hard code the queryset, but that is at least in line with what we do elsewhere. But it will not digest the whole db-table for every single call.

pulp_container/app/registry_api.py
```diff
@@ -451,6 +456,15 @@ class CatalogView(ContainerRegistryApiMixin, ListAPIView):
...
     pagination_class = ContainerCatalogPagination
+    access_policy_class = RegistryAccessPolicy()
+
+    def get_queryset(self, *args, **kwargs):
+        """Filter the queryset based on public repositories and assigned permissions."""
+        queryset = super().get_queryset()
+        permission = "container.pull_containerdistribution"
+        public_repositories = queryset.filter(private=False)
+        accessible_repositories = get_objects_for_user(self.request.user, permission, queryset)
+        return (public_repositories | accessible_repositories).distinct()
```
mdellweg, line 467: :+1:

pulp_container/tests/functional/api/test_repositories_list.py
```diff
@@ -30,8 +32,24 @@
...
+    authorization headers will be removed if you get redirected off-host. In order to override such
+    behaviour, it is necessary to redefine the method which is responsible for changing the
+    authentication header.
+
+    This 'hack' was required for running the tests in the containerized environment, where redirects
+    happen. Credits to https://stackoverflow.com/a/60451840/3907906.
+    """
+
+    def rebuild_auth(self, prepared_request, response):
+        """Do not change the authentication header."""
```
mdellweg, line 48: I think i need more info here. Who is redirecting where?
This "container setup" is a possible real world deployment, so this must not be needed.

pulp_container/app/registry_api.py
```diff
@@ -451,6 +456,15 @@ class CatalogView(ContainerRegistryApiMixin, ListAPIView):
     queryset = models.ContainerDistribution.objects.all().only("base_path")
     serializer_class = ContainerCatalogSerializer
     pagination_class = ContainerCatalogPagination
+    access_policy_class = RegistryAccessPolicy()
+
+    def get_queryset(self, *args, **kwargs):
+        """Filter the queryset based on public repositories and assigned permissions."""
+        queryset = super().get_queryset()
+        permission = "container.pull_containerdistribution"
```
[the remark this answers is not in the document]
mdellweg: Right, and it must be checked against namespaces.

ipanova:
i think we should add model level perms checks here as well https://github.com/pulp/pulp_container/pull/265/files#diff-bada7270b35add8957655a936b1bdb701c4dc829033fd87467735943b839c354R39
this is not related to this pr but i have just noticed that if i have a model perm for pull_distribution i cannot pull on private dists.

pulp_container/tests/functional/api/test_repositories_list.py
lubosmj: So, it behaved like this: When I removed the `.netrc` file, the tests were passing. When I disabled redirects by passing `allow_redirects=False` to the get method, it did not reach the token server. When I used a custom session object (i.e., `session.get`) instead of `requests` (`requests.get`), the tests were passing even when the `.netrc` file was defined.

I thought that this behaviour could be caused either because of "Authorization headers will be removed if you get redirected off-host" or "Authorization headers set with headers= will be overridden if credentials are specified in .netrc, which in turn will be overridden by the auth= parameter. Requests will search for the netrc file at ~/.netrc, ~/_netrc, or at the path specified by the NETRC environment variable." (see Section 2.2.7 https://buildmedia.readthedocs.org/media/pdf/requests/master/requests.pdf). I decided to consider just the first statement.

mdellweg: My question still: Why is there a redirect at all? What is different in the other tests where we don't need such hackery?
Does the first question mean we are deploying pulp wrongly in the ci?

lubosmj: So, `pulp:80/token` is redirected to `pulp/token/` because the Authorization header contains the following value:

`Bearer realm="http://pulp:80/token",service="pulp",scope="registry:catalog:*"`

Could be this discrepancy caused by the configuration declared in `pulp_container/app/webserver_snippets/*.conf`? Would changing `/token http://${pulp-api}/token` to `/token/ http://${pulp-api}/token/` resolve the problem?

lubosmj: Actually, changing the `TOKEN_SERVER` setting will resolve the problem.

mdellweg: Wait, this is redirection by the reverse proxy. The client shouldn't even see this. This is no explanation that we need to mess with the requests library. And still we are calling to the token api in other tests too.
Or are you saying that the client visible redirect is from "http://pulp:80/token" to "http://pulp:80/token/"?

lubosmj: Yes, the client is redirected from `/token` to `/token/`. We are calling the token API without providing authentication headers. That is why it worked in other cases.

mdellweg: Can you add the / in the authentication header?

pulp_container/tests/functional/utils.py
```diff
@@ -73,12 +75,19 @@
 ]
 
 
-def gen_user(permissions):
+def gen_user(model_permissions=None, object_permissions=None):
```
ipanova, line 78: model_permissions = []
this will eliminate the need for the lines 80-85

ipanova (approved):

pulp_container/tests/functional/utils.py
lubosmj: I am not going to use `list` as a default argument here. It could possibly create some issues in the future when calling the method more than once (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

lubosmj:
I dropped the commit that added the check for model-level permission since it is all going to be resolved in a separate PR.

ipanova:
@mdellweg will you review? i am ready to merge otherwise

mdellweg (approved):

PR #267, opened 2021-03-17
Backport 8208

mdellweg:


docs/conf.py
```diff
@@ -57,9 +57,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = "2.2.0"
+version = "2.2.1.dev0"
```
ipanova, line 60: should this be just `.dev`?

mdellweg: It should be the same. And i used bumpversion to get there...

PR #274, opened 2021-03-26
Add checks for model level permissions.

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/access_policy.py
+++ b/pulp_container/app/access_policy.py
@@ -17,6 +17,8 @@ class NamespacedAccessPolicyMixin:
         Check if a user has object-level perms on the namespace associated with the distribution
         or repository.
         """
+        if request.user.has_perm(permission):
+            return True
         obj = view.get_object()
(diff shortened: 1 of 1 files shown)


pulp_container/app/access_policy.py
```diff
@@ -20,11 +20,13 @@ def has_namespace_obj_perms(self, request, view, action, permission):
         obj = view.get_object()
         if type(obj) == models.ContainerDistribution:
             namespace = obj.namespace
-            return request.user.has_perm(permission, namespace)
+            return request.user.has_perm(permission) or request.user.has_perm(permission, namespace)
```
lubosmj, line 23: I think we can call `request.user.has_perm(permission)` just once, right before entering the if-else statement. Then it will look like this:
```
if request.user.has_perm(permission):
    return True

obj = view.get_object()
if type(obj) == models.ContainerDistribution:
    ...
```

mdellweg (approved):

PR #276, opened 2021-03-30
Explain why blob digest can change during push operation.

ipanova:
[noissue]

### DIFF
--- a/docs/workflows/push.rst
+++ b/docs/workflows/push.rst
@@ -50,3 +50,10 @@ pushing a tagged image.
 
 .. note::
    Rollback to the previous repository versions is not possible with a push repository. Its latest version will always be served.
+
+.. warning::
+   Image that has been pulled from a registry and then subsequently pushed to another registy can lead to the blobs digest change.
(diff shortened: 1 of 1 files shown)


mdellweg (approved):
Oh wow! This is interesting. Does this lead to differing Manifests? Or are they referencing checksums to the extracted files?

ipanova:
> Oh wow! This is interesting. Does this lead to differing Manifests? Or are they referencing checksums to the extracted files?

manifests digests changes respectively since its contents change.

ipanova:
> > @mdellweg i have pulled python:latest from docker hub with (1) docker (2) podman and pushed respectively to repo1 and repo2. These repos ended up having blobs and manifests with different digests. So in the nutshell you have 2 identical python images but because of different implementation of recompression on the client side it tuns into blob/manifests having different digests.
> 
> Did they ever think about data integrity?

i would not know what were the challenges around  writing a recompression implementation which is not reproducible **shrugs** and in the first place why not re-using same implmentation across multiple clients instead of writing different code each

PR #284, opened 2021-04-12
Updated Range header in the blob upload response so it is inclusive.

ipanova:
closes #8543


mdellweg (approved):

CHANGES/8543.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed a bug where image push of the same tag with docker client ended up in the different manifest upload.
```
mdellweg, line 1: :+1:

PR #289, opened 2021-04-15
Fix a race while creating push repositories

mdellweg:
This bug is highly reproducable when pushing multilayer images to a not
yet existing repository.

fixes #8565
https://pulp.plan.io/issues/8565

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -336,6 +336,14 @@ class ContainerRegistryApiMixin:
                         distribution = dist_serializer.create(dist_serializer.validated_data)
                 except ValidationError:
                     raise RepositoryInvalid(name=path)
+                except IntegrityError:
+                    # Seems like another process created our stuff already. Retry fetching it.
+                    distribution = models.ContainerDistribution.objects.get(base_path=path)
+                    repository = distribution.repository
+                    if repository:
+                        repository = repository.cast()
+                        if not repository.PUSH_ENABLED:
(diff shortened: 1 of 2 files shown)


Spredzy (approved):
Functionnaly working for me

pulp_container/app/registry_api.py
```diff
@@ -336,6 +336,9 @@ def get_dr_push(self, request, path, create=False):
                         distribution = dist_serializer.create(dist_serializer.validated_data)
                 except ValidationError:
                     raise RepositoryInvalid(name=path)
+                except IntegrityError:
+                    # Seems like another process created our stuff already. Retry fetching it.
+                    distribution = models.ContainerDistribution.objects.get(base_path=path)
```
ipanova, line 341: `repository = distribution.repository` needs to be added to overcome `variable repository was reference before assignment` error

ipanova:
@mdellweg we might need to revise the throttling strategy. Turns out if you specify `wait` on the Throttle() the exponential backoff is not working, meaning that client will always come back after 5 seconds of wait time. Backoff is properly working though if waiting time is not specified -client retries in 4,8,16 seconds.

pulp_container/app/registry_api.py
ipanova: i would love to refactor get_rd_push so it does not look that crazy, maybe not part of this pr.

ipanova (approved):

PR #296, opened 2021-04-20
Change phrasing for repoversion

melcorr:
[noissue]

Problem: 
The way the Versioning is listed in the features, it reads like there are calls to create a repo version specifically. 
There is no workflow for this, and from what I understand it is automatic. 
I would like to introduce this example of a GET response better also: 
https://docs.pulpproject.org/pulp_container/workflows/sync.html#versioned-repo-created
In my opinion, the user lands here and is given no context. 
Could we add some introductory here? Any suggestions? I'd happily do it as part of this PR.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,7 +12,7 @@ Features
 
 * :ref:`Synchronize <sync-workflow>` container image repositories hosted on Docker-hub, Google Container Registry,
   Quay.io, etc., in mirror or additive mode
-* :ref:`Create Versioned Repositories <versioned-repo-created>` so every operation is a restorable snapshot
+* Automatically :ref:`Creates Versioned Repositories <versioned-repo-created>` so every operation is a restorable snapshot
 * :ref:`Download content on-demand <create-remote>` when requested by clients to reduce disk space
 * :ref:`Perform docker/podman pull <host>` from a container distribution served by Pulp
 * :ref:`Perform docker/podman push <push-workflow>` to the Pulp Registry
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

ipanova:
> [noissue]
> 
> Problem:
> The way the Versioning is listed in the features, it reads like there are calls to create a repo version specifically.
> There is no workflow for this, and from what I understand it is automatic.
> I would like to introduce this example of a GET response better also:
> https://docs.pulpproject.org/pulp_container/workflows/sync.html#versioned-repo-created
> In my opinion, the user lands here and is given no context.
> Could we add some introductory here? Any suggestions? I'd happily do it as part of this PR.

+1 let's provide an example for a repository-versions GET call.
To make a bridge between sync a repo-version inspection we can say  'As a result of a successful sync operation a repo version is produced:
<here is the repository-versions GET call>
<and here is repo-version GET Response>

melcorr:
In the container docs, all the example calls are done in a bash script but I don't find an example of a GET call. 
How do you think it would be best to give the example for consistency? 
In my previous job it would have been something like this for a GET call that wasn't so complicated:
`curl -v http://$BASEURL/versions/`

What do you think?

mdellweg:
We usually use `httpie` as in
`http $BASE_URL/pulp/api/v3/repositories/container/<uuid>/versions/`
But maybe it's worth exploring if the cli is ready to cover the workflow examples.

ipanova:
> We usually use `httpie` as in
> `http $BASE_URL/pulp/api/v3/repositories/container/<uuid>/versions/`

This will suffice imo.

> But maybe it's worth exploring if the cli is ready to cover the workflow examples.
I would prefer if switching to the cli examples then doing it consistently across all the workflows. The question is whether cli for pulp-container has all the needed support already

ipanova (approved):

PR #298, opened 2021-04-22
Allow users with namespace permissions to update distributions.

newswangerd:
@ipanova I don't know if using `has_model_or_obj_perms` for `container.change_containerdistribution` was intentional or not, but in case it wasn't, here's a fix.

### DIFF
--- a/CHANGES/8618.bugfix
+++ b/CHANGES/8618.bugfix
@@ -1,0 +1,1 @@
+Fix a bug where users with container.namespace_change_containerdistribution couldn't change distributions.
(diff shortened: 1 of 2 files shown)


ipanova:
whooops, that's an oversight! Thanks for catching it.  Mind opening an issue and create a changelog entry?

mdellweg (approved):

mdellweg:
> @ipanova I also noticed that the [ContainerRepisitoryViewset](https://github.com/pulp/pulp_container/blob/master/pulp_container/app/viewsets.py#L453) exclusively uses `has_model_or_obj_perms`. Should those also be updated?

Good observation! But that would be part of a bigger task:
https://pulp.plan.io/issues/8244

We should not hold this PR for it.

PR #299, opened 2021-04-23
Fix dispatch

ipanova:


dralley (approved):

dralley:
@ipanova Since this one contains my commit, I went ahead and closed my PR.

pulp_container/app/registry_api.py
```diff
@@ -693,7 +693,7 @@ def put(self, request, path, pk=None):
 
             upload.delete()
 
-            job = dispatch(
+            dispatched_task = dispatch(
```
mdellweg, line 696: It would now be fine to just call it `task` already. But that is not a necessary requirement.

mdellweg (approved):

PR #302, opened 2021-04-30
Fixed "connection already closed" error in the Registry handler.

ipanova:
closes #8672

### DIFF
--- a/pulp_container/app/registry.py
+++ b/pulp_container/app/registry.py
@@ -97,6 +97,8 @@ class Registry(Handler):
                 streamed back to the client.
 
         """
+        self._reset_db_connection()
+
         path = request.match_info["path"]
         tag_name = request.match_info["tag_name"]
         distribution = self._match_distribution(path)
@@ -203,6 +205,8 @@ class Registry(Handler):
         """
         Return a response to the "GET" action.
         """
+        self._reset_db_connection()
+
         path = request.match_info["path"]
         digest = "sha256:{digest}".format(digest=request.match_info["digest"])
         distribution = self._match_distribution(path)
--- a/CHANGES/8672.bugfix
+++ b/CHANGES/8672.bugfix
@@ -1,0 +1,1 @@
+Fixed "connection already closed" error in the Registry handler.


pulp_container/app/registry.py
```diff
@@ -97,6 +97,8 @@ async def get_tag(self, request):
                 streamed back to the client.
 
         """
+        self._reset_db_connection()
```
ipanova, line 100: I was not sure where to reset the connection - directly in the plugin or core.
What would be the implications if we'd add `self._reset_db_connection()` instead in core directly in each method?
@daviddavis @mdellweg ?

mdellweg:
I'm not sure what is happening here. As far as i have understood, django orm is recreating the connection whenever it seems to need a new one.
OTOH, i have seen a similar error message when forking the process and then closing the connection in one process while using it in the other. My suspicion is, that we may be hitting some async related bad sharing of a psycopg connection.
Do we know, where the connection is closed in the first place?

pulp_container/app/registry.py
daviddavis: I think what you have here is correct (at least according to my original intent, which was that it should be called once at each entry point and not in helper functions such as `_match_distribution`). 

However, I was very leery about introducing this function originally and still not sure if it's a good idea. Having someone else who's familiar with the content app to review and weigh in would be good.

daviddavis:
Here is the original issue: https://pulp.plan.io/issues/6045

And commit: https://github.com/pulp/pulpcore/commit/5aabb202dd59cbe2d30ef5ad91f01932c9ca041b

Maybe there's some more insight there.

mdellweg:
I also found this: https://www.psycopg.org/docs/advanced.html#support-for-coroutine-libraries

pulp_container/app/registry.py
mdellweg: I'd prefer to have that call being the first thing when a new request is handled. But i don't even see where in `Handler` that would be. But maybe it's safe and not a bit performance impact if we add it in more places.

ipanova:
@dralley @bmbouter  can you provide any input on this PR please? I am going to backport and include it  into an older release needed for katello tomorrow

pulp_container/app/registry.py
bmbouter: I see the core's content app has these two primary entry points to the code https://github.com/pulp/pulpcore/blob/master/pulpcore/content/__init__.py#L54-L55 and I see both of them already call the `self._reset_db_connection` so I think core's already got it. Agreed that I'm also a little unclear on why it's needed in the first place, but core has it.

I'm not very familiar with pulp_container's registry, but it seems like it adds these top-level new routes: https://github.com/pulp/pulp_container/blob/master/pulp_container/app/content.py#L8-L14

ipanova: pulp_container does not call into core's `stream_content` hence I needed to add `self._reset_db_connection` before `_math_distribution` is called.

ipanova:
@mdellweg if you don't have any objections please give me an ack so i can merge and backport this

pulp_container/app/registry.py
dralley: I think in general though, if pulpcore is already calling it in these handlers, then adding the calls here is unlikely to break anything.

mdellweg (approved):

PR #303, opened 2021-04-30
Create and return empty blob on the fly.

ipanova:
closes #8631

### DIFF
--- a/pulp_container/app/registry.py
+++ b/pulp_container/app/registry.py
@@ -11,7 +11,7 @@ from pulpcore.plugin.models import ContentArtifact
 from pulp_container.app.models import ContainerDistribution, Tag
 from pulp_container.app.schema_convert import Schema2toSchema1ConverterWrapper
 from pulp_container.app.utils import get_accepted_media_types
-from pulp_container.constants import MEDIA_TYPE
+from pulp_container.constants import EMPTY_BLOB, MEDIA_TYPE
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


pulp_container/app/registry.py
```diff
@@ -225,3 +228,17 @@ async def get_by_digest(self, request):
                 return await Registry._dispatch(artifact.file, headers)
             else:
                 return await self._stream_content_artifact(request, web.StreamResponse(), ca)
+
+    @staticmethod
+    async def _empty_blob():
+        empty_tar = [
+            31, 139, 8, 0, 0, 9, 110, 136, 0, 255, 98, 24, 5, 163, 96, 20, 140, 88, 0, 8, 0, 0, 255,
+            255, 46, 175, 181, 239, 0, 4, 0, 0,
+        ]
```
mdellweg, line 237: I don't really like, what black does to these lines. Do you think a `b"\x1f..."` string would look better?

mdellweg (approved):
This all looks good.
So you think it's worth adding a small download test? This empty blob should be "available" on all repositories.

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -188,6 +188,37 @@ def test_api_performes_schema_conversion(self):
             computed_digest, header_digest.split(":")[1], "The manifest digests are not equal"
         )
 
+    def test_create_empty_blob_on_the_fly(self):
```
mdellweg, line 191: :+1:

pulp_container/app/registry_api.py
```diff
@@ -760,6 +761,10 @@ def get(self, request, path, pk=None):
         try:
             blob = models.Blob.objects.get(digest=pk, pk__in=repository_version.content)
         except models.Blob.DoesNotExist:
+            if pk == EMPTY_BLOB:
+                return redirects.distribution.redirect_to_content_app(
+                    f"{settings.CONTENT_ORIGIN}/pulp/container/{path}/blobs/{pk}"
+                )
```
mdellweg, line 767: Would this work too?
```suggestion
                return redirects.redirect_to_content_app("blobs", pk)
```

mdellweg (approved):
Just a last suggestion.
Thanks for the test!

pulp_container/app/registry_api.py
ipanova: yes it should, i was wrong copy paste >.>
thanks for pointing this out!

mdellweg:
:shipit:

PR #318, opened 2021-05-19
Fix links to container API

melcorr:
The API links in these workflows are broken. 
I addressed what is mentioned in the issue. 
I tested all the links just there, and they seem correct now, if I'm pointing to the correct place.
I can go through all the workflows as part of this but I wanted to check first: 
Is this likely to break again? Is some better solution necessary?

fixes #8125

### DIFF
--- a/docs/workflows/manage-content.rst
+++ b/docs/workflows/manage-content.rst
@@ -40,7 +40,7 @@ a user is trying to tag an image with a same name but with a different
 digest, the tag associated with the old manifest is going to be
 eliminated in a new repository version.
 
-Reference: `Container Tagging Usage <../restapi.html#tag/container:-tag>`_
+Reference: `Container Tagging Usage <../restapi.html#tag/Content:-Tags>`_
 
 .. _untagging-workflow:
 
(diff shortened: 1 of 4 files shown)


docs/workflows/manage-content.rst
```diff
@@ -40,7 +40,7 @@ a user is trying to tag an image with a same name but with a different
 digest, the tag associated with the old manifest is going to be
 eliminated in a new repository version.
 
-Reference: `Container Tagging Usage <../restapi.html#tag/container:-tag>`_
+Reference: `Container Tagging Usage <../restapi.html#tag/Content:-Tags`_
```
mdellweg, line 43:
```suggestion
Reference: `Container Tagging Usage <../restapi.html#tag/Content:-Tags>`_
```

docs/workflows/sync.rst
```diff
@@ -67,7 +67,7 @@ Remote GET Response::
    matches defined criteria by leveraging wildcards.
 
 
-Reference: `Container Remote Usage <../restapi.html#tag/remotes>`_
+Reference: `Container Remote Usage <../restapi.html#tag/Remotes:-Container`_
```
mdellweg, line 70:
```suggestion
Reference: `Container Remote Usage <../restapi.html#tag/Remotes:-Container>`_
```

mdellweg (commented):
I haven't tested the links, but it looks reasonable to me.

docs/workflows/sync.rst
melcorr: Hah, I am an idiot. I see what I did. Thanks

ipanova (approved):

PR #335, opened 2021-06-01
Pie

mdellweg:


mdellweg:
@ggainey Please take a look as you are an expert here.
I was wondering, why the fk and m2m relations did not automatically used the upstream_id.

ggainey:
> @ggainey Please take a look as you are an expert here.

Will need to resurrect a few neurons, but it looks good at first glance!

> I was wondering, why the fk and m2m relations did not automatically used the upstream_id.

Relying on upstream_id has an (unavoidable) problem: if an imported entity has been deleted, recreated, and then re-export/imported, there is **no way** for the downstream to know that it's the "same" entity; you end up with duplicates.

I wrote up a doc that has a lot of notes on How This All Works, and why relying on an entity's "natural key" is the way to go. See https://hackmd.io/kvXjWYNxQbqCqZAqke25pA for all the (very, very gory) details.

pulp_container/app/modelresource.py
```diff
@@ -0,0 +1,99 @@
...
+        :return: Blobs specific to a specified repo-version.
+        """
+        return Blob.objects.filter(pk__in=self.repo_version.content)
+
+    class Meta:
+        model = Blob
+        import_id_fields = model.natural_key_fields()
+
+
+class ManifestResource(BaseContentResource):
```
ipanova, line 23: Were not you planning to also add a separate ManifestListResource?

mdellweg: I had it, but then with the extra fields on the join table, we need a separate resource for that one, and the need to serialize the ManifestLists separately vanished.

ipanova: nice, I have not realized that `` the need to serialize the ManifestLists separately vanished``

pulp_container/app/modelresource.py
```diff
@@ -0,0 +1,99 @@
...
+class BlobResource(BaseContentResource):
+    """
+    Resource for import/export of blob entities
+    """
+
+    def set_up_queryset(self):
+        """
+        :return: Blobs specific to a specified repo-version.
+        """
+        return Blob.objects.filter(pk__in=self.repo_version.content)
```
ggainey, line 16: Add an explicit order-by to your filters. Without it, it's possible to end up with deadlocks at import-time - see https://github.com/pulp/pulp_rpm/pull/1975 for where we fixed pulp_rpm

pulp_container/tests/functional/api/test_pulpimportexport.py
```diff
@@ -0,0 +1,107 @@
...
+            "repositories": [repository.pulp_href],
+        }
+        exporter = exporters_api.create(body)
+        self.addCleanup(exporters_api.delete, exporter.pulp_href)
+
+        export_response = exports_api.create(exporter.pulp_href, {})
+        export_href = monitor_task(export_response.task).created_resources[0]
+        export = exports_api.read(export_href)
+
+        # Clean the old repository out
```
ggainey, line 76: :+1:

ggainey (changes requested):

pulp_container/app/modelresource.py
```diff
@@ -0,0 +1,101 @@
...
+class BlobResource(BaseContentResource):
+    """
+    Resource for import/export of blob entities
+    """
+
+    def set_up_queryset(self):
+        """
+        :return: Blobs specific to a specified repo-version.
+        """
+        return Blob.objects.filter(pk__in=self.repo_version.content).order_by("content_ptr_id")
```
ggainey, line 16: :+1:

ggainey (approved):
With the addition of order-bys, this looks good to me!

pulp_container/app/modelresource.py
```diff
@@ -0,0 +1,101 @@
...
+        widget=widgets.ForeignKeyWidget(Manifest, field="digest"),
+    )
+
+    def set_up_queryset(self):
+        """
+        :return: Manifests specific to a specified repo-version.
+        """
+        return ManifestListManifest.objects.filter(
+            manifest_list__pk__in=self.repo_version.content
+        ).order_by("id")
```
ipanova, line 73: pulp_id

CHANGES/6636.feature
```diff
@@ -0,0 +1 @@
+Add model resources to allow pulp import export handle pulp_container content units.
```
ipanova, line 1: I suggest capturing this feature in this feature list https://docs.pulpproject.org/pulp_container/#features  e.g. "Support disconnected and air-gaped environments with pulp import/export facility'

ipanova:
I have tested he following workflow:
1. create and sync repo A that has blobs/manifests/listed manifests/tags
2. Export it
3. On a clean env I have created container repo with the same name A
4. Performed import.

ipanova:
while import/export for synced container repos works fine, additional changes needs to be made for container-push repo import/export. Container-push repo cannot be created ahead of import because it is always being created as part of `podman push` operation alongside with the distribution.

pulp_container/app/modelresource.py
ipanova: you're right :+1

docs/index.rst
```diff
@@ -27,6 +27,7 @@ Features
 * :ref:`Build an OCI format image from a Containerfile <build-Containerfile-workflow>` and make it available from the Pulp Registry
 * Host content either `locally or on S3 <https://docs.pulpproject.org/installation/storage.html>`_
 * De-duplication of all saved content
+* Support disconnected and air-gaped environments with pulp import/export facility
```
ipanova, line 30: @mdellweg  can you please update this line so it captures that import/export would work only for synced repos?And the changelog entry as well. We will add separately support for pushed repos, once we complete this change in core https://pulp.plan.io/issues/7297
My last suggestion would be to add a note that it is in the tech preview https://docs.pulpproject.org/pulp_container/tech-preview.html

ipanova (approved):

docs/index.rst
```diff
@@ -27,6 +27,7 @@ Features
 * :ref:`Build an OCI format image from a Containerfile <build-Containerfile-workflow>` and make it available from the Pulp Registry
 * Host content either `locally or on S3 <https://docs.pulpproject.org/installation/storage.html>`_
 * De-duplication of all saved content
+* Support disconnected and air-gaped environments with pulp import/export facility for synched container repositories
```
ipanova, line 30: i think, it is `synced` or at least i was always writing it this way :D

PR #337, opened 2021-06-11
Fixed access policy for the continaer repository ``repair`` endpoint.

ipanova:
closes #8884

### DIFF
--- a/CHANGES/8884.bugfix
+++ b/CHANGES/8884.bugfix
@@ -1,0 +1,1 @@
+Fixed access policy for the container repository ``repair`` endpoint.
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -796,6 +796,15 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionQuerySetMixin, Reposito
                     "has_repo_attr_model_or_obj_perms:container.view_containerrepository",
                 ],
             },
+            {
+                "action": ["repair"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_repo_attr_model_or_obj_perms:container.sync_containerrepository",
+                    "has_repo_attr_model_or_obj_perms:container.view_containerrepository",
+                ],
+            },
         ],
     }
 


pulp_container/app/viewsets.py
```diff
@@ -796,6 +796,16 @@ class ContainerRepositoryVersionViewSet(RepositoryVersionQuerySetMixin, Reposito
                     "has_repo_attr_model_or_obj_perms:container.view_containerrepository",
                 ],
             },
+            {
+                "action": ["repair"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_repo_attr_model_or_obj_perms:container.sync_containerrepository",
```
ipanova, line 804: @mdellweg @goosemania @dkliban can you weigh in here?
I thought to use `modify_content_containerrepository` but none of  the actions assigned to that permission bring in RA.
I can introduce a new permission `container.repair_content_containerrepository` but i am trying to avoid the migration

mdellweg: looks like a reasonable choice to me.

ipanova: also, i have on purpose non added `repair` for the container-push repo because all of the content that comes in it comes through the upload.

mdellweg (approved):

CHANGES/8884.bugfix
```diff
@@ -0,0 +1 @@
+Fixed access policy for the continaer repository ``repair`` endpoint.
```
goosemania, line 1:
```suggestion
Fixed access policy for the container repository ``repair`` endpoint.
```

pulp_container/app/viewsets.py
goosemania: I think your choice is reasonable, though it brings me back to the topic of how to make it understandable for users which perms they need for what if they work with pulp's default policies.
Should the default policies become a part of the generated documentation?

goosemania (approved):

pulp_container/app/viewsets.py
ipanova: I agree that this change is rather implicit than explicit.
The only place where the remissions are 'documented' in some way is here https://github.com/pulp/pulp_container/blob/master/pulp_container/app/models.py#L372
It would be great if we can have default policy documented

PR #346, opened 2021-06-18
Remove Django pin.

ipanova:
[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ ecdsa~=0.13.2
 pyjwkest~=1.4.0
 pyjwt[crypto]~=1.7.1
 url-normalize~=1.4.2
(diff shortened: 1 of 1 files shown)


goosemania (approved):
:bouquet:

mdellweg (approved):

mdellweg:
The error here looks like a mismatch of click and rq.

mdellweg:
Maybe we need to add a pin for click<8.

ipanova:
> Maybe we need to add a pin for click<8.

there is a pin in the pulp-smash, i am still not sure what installs click 0.8

mdellweg:
Pulp-smash is outside of the container. This is failing rq inside it.

mdellweg (approved):

PR #347, opened 2021-06-18
Ensure banner gets added when docs are built

gerrod3:
[noissue]

When the CI goes to build the docs it installs a really outdated version of sphinx that doesn't have the feature required to add the js needed for the inserting the survey banner.


ipanova:
the test failure is unrelated, thank you!

ipanova (approved):

PR #352, opened 2021-06-30
Add pulpcore upper bound.

ipanova:
[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pulpcore>=3.12.1
+pulpcore>=3.14,<3.15
 ecdsa~=0.13.2
 pyjwkest~=1.4.0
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-pulpcore>=3.12.1
+pulpcore>=3.14,<3.15
```
dkliban, line 1: This means that you can only release after pulpcore 3.14 is released.

dkliban (approved):
Thank you!

dkliban (changes requested):
There are a couple of small changes that got added to plugin_template yesterday. Let's include them in this PR.

dkliban (approved):
I pushed a commit to your branch. So i am approving the PR now.

PR #355, opened 2021-07-02
Rename NewDistributionFilter to DistributionFilter

goosemania:
To address pulpcore 3.15 compatibility.
NewDistributionFilter was deprecated in 3.14.
DistributionFilter was added in 3.14

[noissue]


goosemania:
NewDistributionFilter has just been removed from the pulpcore.
Fixing it because it affects pulp-2to3-migration CI.

ipanova (approved):

PR #358, opened 2021-07-15
Comply with breaking changes drf-access-policy.

ipanova:
[noissue]


fao89:
I believe this is the issue galaxy is facing, I think they are using RHEL8
```
[ec2-user@ip-172-16-8-22 ~]$ podman push --tls-verify=false 3.238.62.78/demo
Checking if image destination supports signatures
Error: error copying image to the remote destination: Copying this image requires changing layer representation, which is not possible (image is signed or the destination specifies a digest)
```

daviddavis (approved):

dkliban:
Should we add a CHANGELOG entry for the change that was needed because of drf-access-policy version bump?

ipanova:
> I believe this is the issue galaxy is facing, I think they are using RHEL8
> 
> ```
> [ec2-user@ip-172-16-8-22 ~]$ podman push --tls-verify=false 3.238.62.78/demo
> Getting image source signatures
> Checking if image destination supports signatures
> Error: error copying image to the remote destination: Copying this image requires changing layer representation, which is not possible (image is signed or the destination specifies a digest)
> ```

this does not seem to be related, as per slack discussion there has been revealed some weirdness of podman versions on rhel

pulp_container/app/viewsets.py
```diff
@@ -1046,7 +1046,7 @@ class ContainerDistributionViewSet(DistributionViewSet):
                 "action": ["pull"],
                 "principal": "*",
                 "effect": "allow",
-                "condition": [
+                "condition_expression": [
```
mdellweg, line 1049: The update of that package will happen in pulpcore, right? Would that need a migration?

mdellweg (approved):

pulp_container/app/viewsets.py
ipanova: yes, the new version of drf-access-policy was already merged in pulpcore. I don't think however it'd need a migration, since the change happened on the viewset, the policy will be properly updated here https://github.com/pulp/pulpcore/blob/master/pulpcore/app/apps.py#L197

ipanova: to five some background, tldr they have split the condition element into condition and condition_expression, where the latter one should be use during the boolean logic evaluation

mdellweg: I was thinking about the non default policies (modified by users). But when you say, that it's not as easy as moving "condition" to "condition_expression" we should not even think about touching it in a migration.

ipanova: hm, we're not supposed to touch customized policies, that's why we're exposing this change in the changelog entry so users can take action if needed

ipanova: actually @newswangerd  fyi

newswangerd: wow, this is annoying. Thanks for the heads up @ipanova

PR #362, opened 2021-07-27
Touch artifacts and content to prevent cleanup

mdellweg:
This change prepares for the async orphan cleanup that is expected to
come with pulpcore 3.15.

fixes #9134

### DIFF
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -372,6 +372,7 @@ class TagImageSerializer(TagOperationSerializer):
             manifest = models.Manifest.objects.get(
                 pk__in=new_data["latest_version"].content.all(), digest=new_data["digest"]
             )
+            manifest.touch()
         except models.Manifest.DoesNotExist:
             raise serializers.ValidationError(
                 _(
@@ -608,7 +609,9 @@ class OCIBuildImageSerializer(serializers.Serializer):
                     _("Only one of 'containerfile' and 'containerfile_artifact' may be specified.")
                 )
             data["containerfile_artifact"] = Artifact.init_and_validate(data.pop("containerfile"))
-        elif "containerfile_artifact" not in data:
+        elif "containerfile_artifact" in data:
+            data["containerfile_artifact"].touch()
+        else:
             raise serializers.ValidationError(
                 _("'containerfile' or 'containerfile_artifact' must " "be specified.")
             )
@@ -627,6 +630,7 @@ class OCIBuildImageSerializer(serializers.Serializer):
                 )
                 try:
                     artifact = artifactfield.run_validation(data=url)
+                    artifact.touch()
                     artifacts[artifact.pk] = relative_path
                 except serializers.ValidationError as e:
                     # Append the URL of missing Artifact to the error message
--- a/pulp_container/app/tasks/builder.py
+++ b/pulp_container/app/tasks/builder.py
@@ -31,6 +31,7 @@ def get_or_create_blob(layer_json, manifest, path):
     """
     try:
         blob = Blob.objects.get(digest=layer_json["digest"])
+        blob.touch()
     except Blob.DoesNotExist:
         layer_file_name = "{}{}".format(path, layer_json["digest"][7:])
         layer_artifact = Artifact.init_and_validate(layer_file_name)
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -107,6 +107,7 @@ class ContainerFirstStage(Stage):
             except IntegrityError:
                 del tag.artifact_attributes["file"]
                 saved_artifact = Artifact.objects.get(**tag.artifact_attributes)
+                saved_artifact.touch()
 
             tag_name = tag.url.split("/")[-1]
             tag_dc = DeclarativeContent(Tag(name=tag_name))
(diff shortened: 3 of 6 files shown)


pulp_container/app/serializers.py
```diff
@@ -372,6 +372,10 @@ def validate(self, data):
             manifest = models.Manifest.objects.get(
                 pk__in=new_data["latest_version"].content.all(), digest=new_data["digest"]
             )
+            # TODO Evaluate, whether we need touch here.
+            # The manifest is in a repository version, so not orphaned,
+            # but can the version be deleted between this and using it?
+            manifest.touch()
```
mdellweg, line 378: Please discuss

pulp_container/app/viewsets.py
```diff
@@ -668,7 +669,7 @@ def copy_tags(self, request, pk):
 
         result = dispatch(
             tasks.recursive_add_content,
-            [repository],
+            [repository, source_latest.repository],
```
mdellweg, line 672: I think, this was actually a bug.

pulp_container/app/viewsets.py
```diff
@@ -705,7 +706,7 @@ def copy_manifests(self, request, pk):
         manifests_to_add = manifests_in_repo.filter(**filters)
         result = dispatch(
             tasks.recursive_add_content,
-            [repository],
+            [repository, source_latest.repository],
```
mdellweg, line 709: same here

pulp_container/app/serializers.py
ipanova: there is a small window for such case, so i probably would keep it.

pulp_container/app/serializers.py
```diff
@@ -608,7 +612,9 @@ def validate(self, data):
                     _("Only one of 'containerfile' and 'containerfile_artifact' may be specified.")
                 )
             data["containerfile_artifact"] = Artifact.init_and_validate(data.pop("containerfile"))
-        elif "containerfile_artifact" not in data:
+        elif "containerfile_artifact" in data:
```
ipanova, line 615: i don't think this should be called here, at this point the artifact was just created in memory but not saved. This should suffice https://github.com/pulp/pulp_container/pull/362/files#diff-31c44b2812e50f3171248db027f439e8e64248cb84067fc7c305de74cdb9a86fR744

mdellweg: Actually i believe this is the codepath where it was supplied as an artifact_href by the user.

pulp_container/app/tasks/builder.py
```diff
@@ -111,12 +112,14 @@ def build_image_from_containerfile(
 
     """
     containerfile = Artifact.objects.get(pk=containerfile_pk)
+    containerfile.touch()
```
ipanova, line 115: i think, here, the line is not needed as well since we touched it in the viewset

pulp_container/app/tasks/sync_stages.py
```diff
@@ -107,6 +107,7 @@ async def run(self):
             except IntegrityError:
                 del tag.artifact_attributes["file"]
                 saved_artifact = Artifact.objects.get(**tag.artifact_attributes)
```
ipanova, line 109: :+1

pulp_container/app/serializers.py
```diff
@@ -627,6 +633,7 @@ def validate(self, data):
                 )
                 try:
                     artifact = artifactfield.run_validation(data=url)
```
ipanova, line 635: i can't grasp entirely when these artifacts are being saved since i don't see any further save call. I assume it is the `artifactfield.run_validation(data=url)` that creates and saves the artifacts? in that case we probably don't need that touch later.

pulp_container/app/tasks/builder.py
```diff
@@ -111,12 +112,14 @@ def build_image_from_containerfile(
...
     repository = ContainerRepository.objects.get(pk=repository_pk)
     name = str(uuid4())
     with tempfile.TemporaryDirectory(".") as working_directory:
         path = "{}/".format(working_directory.path)
         for key, val in artifacts.items():
             artifact = Artifact.objects.get(pk=key)
+            artifact.touch()
```
ipanova, line 122: i'd move this to the viewset

pulp_container/app/serializers.py
ipanova: ooooh, you're right!

pulp_container/app/serializers.py
ipanova: disregard, i have read the helptext for the `artifacts`

ipanova (approved):
thank you for this pr, it looks good to me - i would rather move touch outside of the task into the viewset https://github.com/pulp/pulp_container/pull/362#discussion_r677387183 and this one seems to be redundant pulp_container

pulp_container/app/tasks/builder.py
mdellweg: I had the same thougt. But then i decided to go ahead anyway, because there could be a considerate a amount of time between the request and the task, and it would be just so inconvenient if the artifact vanished between here and being used after reading all the other artifacts from cloud storage.

ipanova: with this logic then we should double that use of touch in viewset and inside of the task..
the default is set to 1day, i'd still insist on not doubling this.

PR #363, opened 2021-07-27
Test nginx snippet

fao89:
[noissue]


fao89:
Weirdly pulp-container breaks with this change, but pulp-ansible not

PR #364, opened 2021-07-27
Enables reclaim disk space feature.

ipanova:
[noissue]

### DIFF
--- a/CHANGES/9169.feature
+++ b/CHANGES/9169.feature
@@ -1,0 +1,1 @@
+Enable reclaim disk space feature for blobs and manifests.This feature is available with pulpcore 3.15+
--- a/pulp_container/app/models.py
+++ b/pulp_container/app/models.py
@@ -55,6 +55,8 @@ class Blob(Content):
         manifest (models.ForeignKey): Many-to-one relationship with Manifest.
     """
 
+    PROTECTED_FROM_RECLAIM = False
+
     TYPE = "blob"
 
     BLOB_CHOICES = (
@@ -90,6 +92,8 @@ class Manifest(Content):
         listed_manifests (models.ManyToManyField): Many-to-many relationship with Manifest. This
             field is used only for a manifest-list type Manifests.
     """
+
+    PROTECTED_FROM_RECLAIM = False
 
     TYPE = "manifest"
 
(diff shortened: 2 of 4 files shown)


pulp_container/app/registry.py
```diff
@@ -163,7 +156,8 @@ async def dispatch_tag(tag, response_headers):
         try:
             artifact = tag.tagged_manifest._artifacts.get()
         except ObjectDoesNotExist:
-            raise ArtifactNotFound(tag.name)
+            ca = tag.tagged_manifest.contentartifact_set.all()[0]
```
ipanova, line 159: this change is actually redundant if we 'protect' the manifests from being reclaimed.

pulp_container/app/models.py
```diff
@@ -91,6 +93,8 @@ class Manifest(Content):
             field is used only for a manifest-list type Manifests.
     """
 
+    protected_from_reclaim = True
```
mdellweg, line 96: I would introduce this class constant defaulting to to True in pulpcore and only overwrite if reclaim is supported.
This way plugins can opt in to this feature, and don't need to opt out.
Also it's supposed to be a constant, so it should be all caps.

ipanova: it is set to true in core, it is just mentioned here in a redundant way so i will remove it

pulp_container/app/registry.py
ipanova: I have decided to enable reclaim for manifests as well, it required to tweak the logic in schema_conversion.

mdellweg: What if the manifest is broken and has not artifact nor remote artifact?

mdellweg (approved):

pulp_container/app/registry.py
ipanova: actually handler handles this just fine - 404will be raised https://github.com/pulp/pulpcore/blob/master/pulpcore/content/handler.py#L681

CHANGES/9169.feature
```diff
@@ -0,0 +1 @@
+Enable reclaim disk space feature for blobs and manifests.
```
mdellweg, line 1: We should note, it's only working with pulpcore 3.15.

PR #366, opened 2021-07-29
Add one more case for touch()

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/tasks/tag.py
+++ b/pulp_container/app/tasks/tag.py
@@ -1,4 +1,4 @@
-from pulpcore.plugin.models import ContentArtifact, CreatedResource, Repository
+from pulpcore.plugin.models import CreatedResource, Repository
 from pulp_container.app.models import Manifest, Tag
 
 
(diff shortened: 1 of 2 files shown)


pulp_container/app/tasks/tag.py
```diff
@@ -27,10 +27,8 @@ def tag_image(manifest_pk, tag, repository_pk):
     if created:
         resource = CreatedResource(content_object=manifest_tag)
         resource.save()
-
```
ipanova, line 30: tag is artifactless

mdellweg (approved):

pulp_container/app/tasks/tag.py
```diff
@@ -27,10 +26,8 @@ def tag_image(manifest_pk, tag, repository_pk):
...
-    ContentArtifact.objects.get_or_create(
-        artifact=artifact, content=manifest_tag, relative_path=tag
-    )
+    else:
+        manifest_tag.touch()
```
mdellweg, line 30: good find!
Should we link this to #9134?

ipanova: i did reference 9134 so the revision is created however i think ci will still fail because of lack of changelong entry, i suggest to merge as admin

mdellweg: Well, the changelog is there; just not in the same commit. XD

ipanova: hm ok, seems like ci logic is smart enough to handle this case, merging :)

mdellweg: I'd say, not smart enough to _see_ it.

PR #382, opened 2021-08-30
Pulpcore 3.15 adjustments

mdellweg:
* async ProgressReport

[noissue]

### DIFF
--- a/CHANGES/9310.misc
+++ b/CHANGES/9310.misc
@@ -1,0 +1,1 @@
+Refactored sync pipeline for pulpcore 3.15 compatibility.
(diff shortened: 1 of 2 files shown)


mdellweg:
:man_facepalming: "Update plugin ci for release branch" should never have landed on the main branch.

mdellweg:
This will need to be rebased on #383 , because we might need to backport that one.

mdellweg:
That's wrong for sure. WDYT? Why doesn't it fail the usual sync tests?
https://github.com/pulp/pulp_container/pull/382/checks?check_run_id=3495986687#step:11:929

dkliban (approved):

PR #383, opened 2021-08-30
Sync race

mdellweg:


pulp_container/app/tasks/sync_stages.py
```diff
@@ -498,23 +501,6 @@ def relate_blob(self, dc):
...
-        Relate an ImageManifest to a Tag.
-
-        Args:
-            dc (pulpcore.plugin.stages.DeclarativeContent): dc for a Tag
-
-        """
-        related_dc = dc.extra_data.get("man_relation")
-        dc.content.tagged_manifest = related_dc.content
-        try:
-            dc.content.save()
```
mdellweg, line 512: This is particularly bad, as content should be immutable. And i believe it is the root cause of https://pulp.plan.io/issues/9292 .

pulp_container/app/tasks/sync_stages.py
```diff
@@ -498,23 +501,6 @@ def relate_blob(self, dc):
...
-        except IntegrityError:
-            existing_tag = Tag.objects.get(name=dc.content.name, tagged_manifest=related_dc.content)
-            dc.content.delete()
```
mdellweg, line 515: And deletion should never happen outside of orphan_cleanup.

mdellweg:
@dkliban this is the almost minimal patch for the attached issue. Maybe we should get this out soon. But i see more need to refactor the interrelate process.
@jlsherrill can you have a trial run with this?

dkliban (approved):

jlsherrill:
this appears to have resolved the issue for me!

PR #384, opened 2021-08-31
Update CI files

pulpbot:
[noissue]


.github/workflows/scripts/script.sh
```diff
@@ -145,7 +145,7 @@ if [[ "$TEST" == "upgrade" ]]; then
 
   # Running post upgrade tests
   git checkout ci_upgrade_test -- pulp_container/tests/
-  pytest -vv -r sx --color=yes --pyargs --capture=no --durations=0 pulp_container.tests.upgrade.post
+  pytest -v -r sx --color=yes --pyargs -capture=no pulp_container.tests.upgrade.post
```
mdellweg, line 148: Is this correct?
@fao89

fao89: yep, I forgot to change back, I added these to debug

mdellweg: And what about the double dash before "capture" ?

fao89: I didn't notice it, I need to change it at the plugin-template

fao89: done - https://github.com/pulp/plugin_template/pull/486

mdellweg (approved):

PR #394, opened 2021-09-08
Adopt exclusive and shared resources in the task dispatch.

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -647,7 +647,7 @@ class BlobUploads(ContainerRegistryApiMixin, ViewSet):
             try:
                 task = Task.objects.filter(
                     name__endswith="add_and_remove",
-                    reserved_resources_record__resource=f"upload:{pk}",
+                    reserved_resources_record=[f"upload:{pk}"],
                 ).last()
             except Task.DoesNotExist:
                 # No upload and no task for it => the upload probably never existed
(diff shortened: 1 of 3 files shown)


dralley (approved):

pulp_container/app/registry_api.py
```diff
@@ -647,7 +647,7 @@ def put(self, request, path, pk=None):
             try:
                 task = Task.objects.filter(
                     name__endswith="add_and_remove",
-                    reserved_resources_record__resource=f"upload:{pk}",
+                    reserved_resources_record=[f"upload:{pk}"],
```
mdellweg, line 650: Good catch! This is actually a bugfix, right?

pulp_container/app/registry_api.py
```diff
@@ -699,7 +699,8 @@ def put(self, request, path, pk=None):
 
             dispatched_task = dispatch(
                 add_and_remove,
-                [f"upload:{pk}", repository],
+                shared_resources=[f"upload:{pk}"],
```
mdellweg, line 702: Shared is ok, since we really only _use_ this to identify the task.
(Reminds me that we wanted to find a way to run tasks in the foreground [not dispatching them].)

pulp_container/app/viewsets.py
```diff
@@ -574,7 +574,8 @@ def sync(self, request, pk):
 
         result = dispatch(
             tasks.synchronize,
-            [repository, remote],
+            shared_resources=[remote],
```
mdellweg, line 577: :+1:

mdellweg (approved):

pulp_container/app/registry_api.py
ipanova: yeah, this was forgotten to be updated to reflect new tasking system changes

PR #396, opened 2021-09-14
sync_async

ipanova:
[nooissue]

### DIFF
--- a/CHANGES/9385.misc
+++ b/CHANGES/9385.misc
@@ -1,0 +1,2 @@
+Use sync_to_async to mitigate DB connection issues and comply with Django 3.2 ORM usage in async
+code.
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -145,7 +145,8 @@ class ContainerFirstStage(Stage):
 
         for manifest_future in future_manifests:
             man = await manifest_future.resolution()
-            with man._artifacts.get().file.open() as content_file:
+            artifact = await sync_to_async(man._artifacts.get)()
+            with artifact.file.open() as content_file:
                 raw = content_file.read()
             content_data = json.loads(raw)
             man_dc = man_dcs[man.digest]


dralley (approved):

dralley (approved):

pulp_container/app/modelresource.py
```diff
@@ -95,7 +97,7 @@ def set_up_queryset(self):
 
     class Meta:
         model = Tag
-        import_id_fields = model.natural_key_fields()
+        import_id_fields = tuple(chain.from_iterable(model._meta.unique_together))
```
dralley, line 100: Leave a comment about why this needs to be done

pulp_container/app/models.py
```diff
@@ -199,6 +199,13 @@ class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
         unique_together = (("name", "tagged_manifest"),)
 
+    @classmethod
+    def natural_key_fields(cls):
+        """
+        Returns a tuple of the natural key fields which usually equates to unique_together fields
+        """
+        return ("name", "tagged_manifest_id")
```
dralley, line 207: Same here

CHANGES/9385.misc
```diff
@@ -0,0 +1 @@
+Use sync_to_async to mitigate DB connection issues.
```
mdellweg, line 1: "And comply with Django 3.2 ORM usage in async code." ?

pulp_container/app/models.py
mdellweg: "natural_key_fields" is a pulp invention. Is there a way we could be more clever on the pulpcore side?
Just thinking. Not blocking here.

dralley: We probably could, by doing something like [this](https://github.com/rsinger86/django-lifecycle/pull/64/commits/eed6c922b8e12a6397b7c037723b366bcc88371c#diff-ccd0462c3332b1460bdeefa901260186d8b9de888e6acd16c6d63a0fbbbe2c27R161-R163). 

This feels OK as a short-term solution though.  This is a pretty rare case.

mdellweg: Looks like this could work. But yes, we should cache the result.

https://github.com/pulp/pulpcore/pull/1625

pulp_container/app/modelresource.py
mdellweg: Actually, this should not be true when relying on the natural_key PR in pulpcore. It purposefully does _not_ change `natural_key_fields`.

ipanova: yeah, sorry i have a typo s/contains/constraints/

ipanova: I'm a mess, got confused with natural_keys()

ipanova:
@mdellweg I'm waiting to rebase/merge this after your PR

mdellweg:
> @mdellweg I'm waiting to rebase/merge this after your PR

Redone it with the plugin_template:
https://github.com/pulp/pulp_container/pull/397

mdellweg:
Also we need to wait for:
https://github.com/pulp/pulpcore/pull/1625

PR #397, opened 2021-09-16
Unset DJANGO_ALLOW_ASYNC_UNSAFE

mdellweg:
[noissue]


mdellweg:
Sadly this approach does not work, because django does not care about the value of this environment variable. You really need to remove it to make it work, it seems.

lubosmj:
@mdellweg, do we also want to change the env value in pulpcore/content/\_\_init\_\_.py: https://github.com/pulp/pulpcore/blob/3e80f006f9afa3a2227637040b2d4b2a1990a308/pulpcore/content/__init__.py#L16 ?

ipanova:
> @mdellweg, do we also want to change the env value in pulpcore/content/__init__.py: https://github.com/pulp/pulpcore/blob/3e80f006f9afa3a2227637040b2d4b2a1990a308/pulpcore/content/__init__.py#L16 ?

There is a separate piece of work that needs to be done for content app https://pulp.plan.io/issues/9354.  For now we can do this only for tasking system

ipanova (approved):

mdellweg:
Actually, i expect this PR to fail CI. We need to combine it with your change one or the other way...

PR #398, opened 2021-09-16
Use bulk_touch when adding content

lubosmj:
[noissue]

### DIFF
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -20,6 +20,7 @@ from guardian.shortcuts import get_objects_for_user
 
 from rest_framework import mixins
 from rest_framework.decorators import action
+from rest_framework.exceptions import ValidationError
 
 from pulpcore.plugin.access_policy import AccessPolicyFromDB
 from pulpcore.plugin.models import RepositoryVersion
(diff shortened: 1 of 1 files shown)


lubosmj:
The CI failure is probably caused by the same problem as stated in the comment https://github.com/pulp/pulp_container/pull/396#pullrequestreview-756279106.

pulp_container/app/viewsets.py
```diff
@@ -594,16 +594,17 @@ def add(self, request, pk):
...
         repository = self.get_object()
         serializer = serializers.RecursiveManageSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
 
         if "content_units" in request.data:
-            for url in request.data["content_units"]:
-                content = NamedModelViewSet.get_resource(url, Content)
-                content.touch()
-                add_content_units.append(str(content.pk))
+            add_content_units = [
```
ipanova, line 602: @lubosmj  tldr; you'd need to make more or less same changes as you did in core for the `modify` endpoint. Where you've also raise error for non existent content in the separate function you've written.

mdellweg (commented):
This all looks good. There is just one subtle difference to notice. The old code did a db lookup on all content units. The new one will only extract the pks and pass them on to the task.

lubosmj:
I am thinking about exposing the method used for touching content from pulpcore (https://pulp.plan.io/issues/9419).

pulp_container/app/viewsets.py
```diff
@@ -595,24 +596,44 @@ def add(self, request, pk):
...
+        existing_content_units_pks = {str(pk) for pk in existing_content_units_pks}
+
+        missing_pks = content_units_pks - existing_content_units_pks
+        if missing_pks:
+            missing_hrefs = [content_units[pk] for pk in missing_pks]
+            raise ValidationError(
+                _("Could not find the following content units: {}").format(missing_hrefs)
+            )
+
+        Content.objects.filter(pk__in=content_units.keys()).touch()
```
mdellweg, line 635: Please make it first touch then verify.
Also keeping the query in a variable may help. Like: `qs = Content.objects.filter(pk__in=content_unit_pks)`.

mdellweg (changes requested):

pulp_container/app/viewsets.py
lubosmj: Could you please provide any specific reason why should I touch content before verifying it?

mdellweg: because touching prevents it from disappearing. If you first check, and then touch, you leave a window open where the content is gone by the time you attempt to touch it.

mdellweg (approved):

PR #399, opened 2021-09-17
Update import test to handle task group response

daviddavis:
[noissue]

Required PR: https://github.com/pulp/pulpcore/pull/1631

### DIFF
--- a/pulp_container/tests/functional/api/test_pulpimportexport.py
+++ b/pulp_container/tests/functional/api/test_pulpimportexport.py
@@ -91,7 +91,10 @@ class PulpImportExportTestCase(unittest.TestCase):
 
         filenames = [f for f in list(export.output_file_info.keys()) if f.endswith("tar.gz")]
         import_response = imports_api.create(importer.pulp_href, {"path": filenames[0]})
-        task_group_href = monitor_task(import_response.task).created_resources[1]
+        if hasattr(import_response, "task_group"):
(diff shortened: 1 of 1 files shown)


pulp_container/tests/functional/api/test_pulpimportexport.py
```diff
@@ -91,8 +91,7 @@ def test_import_export(self):
 
         filenames = [f for f in list(export.output_file_info.keys()) if f.endswith("tar.gz")]
         import_response = imports_api.create(importer.pulp_href, {"path": filenames[0]})
-        task_group_href = monitor_task(import_response.task).created_resources[1]
-        monitor_task_group(task_group_href)
+        monitor_task_group(import_response["task_group"])
```
mdellweg, line 94: We could make it understand both versions of pulpcore:

```suggestion
        if "task_group" in import_response:
            task_group_href = import_response["task_group"]
        else:
            task_group_href = monitor_task(import_response.task).created_resources[1]
        monitor_task_group(task_group_href)
```
also it seems strange, that the task_group is accessed via `[]` while the task seems to be an attribute. Is this bindings weirdness?

daviddavis: That's a good question. I'll take a look on Monday.

mdellweg (approved):

PR #400, opened 2021-09-24
Enable azure tests

fao89:
[noissue]


fao89:
it requires a deeper knowledge of pulp_container:
https://github.com/pulp/pulp_container/blob/master/pulp_container/app/registry_api.py#L729-L745

PR #402, opened 2021-09-27
Add validation for a repository base path

lubosmj:
closes #9403

### DIFF
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -1,5 +1,6 @@
 from gettext import gettext as _
 import os
+import re
 
 from django.db import IntegrityError
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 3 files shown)


pulp_container/app/serializers.py
```diff
@@ -288,6 +293,12 @@ def validate(self, data):
                 )
         return validated_data
 
+    def is_base_path_valid(self, base_path):
+        """Check whether the passed repository base path is valid or not."""
+        component_re = re.compile(r"^[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?$")
```
mdellweg, line 298: I am a bit unsure, how long these processes live. Do you think it's worth storing the compiled regex in a global var?

pulp_container/app/serializers.py
```diff
@@ -288,6 +293,12 @@ def validate(self, data):
...
+        all_match = all(re.match(component_re, piece) for piece in base_path.split("/"))
+        return len(base_path) < 256 and all_match
```
mdellweg, line 300: The `len` test seems to be cheap. Can we have it return before we start the regex operations? Also, i would raise the exceptions from this function instead of returning a bool (renaming it accordingly).
Actually, maybe follow this pattern: https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

pulp_container/app/serializers.py
```diff
@@ -288,6 +290,20 @@ def validate(self, data):
                 )
         return validated_data
 
+    def validate_base_path(self, value):
```
mdellweg, line 293: :+1:

mdellweg (approved):

PR #403, opened 2021-09-28
WIP: Use roles for RBAC

mdellweg:
Required-PR: https://github.com/pulp/pulpcore/pull/1627

[noissue]


pulp_container/app/access_policy.py
```diff
@@ -24,15 +28,14 @@ def has_namespace_obj_perms(self, request, view, action, permission):
             namespace = obj.namespace
             return request.user.has_perm(permission, namespace)
         elif type(obj) == models.ContainerPushRepository:
-            dists_qs = models.ContainerDistribution.objects.filter(repository=obj)
-            for dist in dists_qs:
+            for dist in obj.distributions:
```
ipanova, line 31: nice

pulp_container/app/viewsets.py
```diff
@@ -857,64 +878,32 @@ class ContainerPushRepositoryViewSet(
...
-                },
-                "permissions": [
-                    "container.view_containerpushrepository",
+                    "has_namespace_obj_perms:container.change_containerpushrepository or "
+                    "has_distribution_perms:container.change_containerpushrepository",
                 ],
             },
         ],
     }
+    LOCKED_ROLES = {}
```
ipanova, line 906: this is still TBD right?

pulp_container/app/access_policy.py
```diff
@@ -96,6 +99,21 @@ class NamespacedAccessPolicyFromDB(AccessPolicyFromDB, NamespacedAccessPolicyMix
...
+class PushRepositoryAccessPolicy(NamespacedAccessPolicyFromDB):
+    """
+    Access policy for ContainerPushRepositoryViewSet.
+    """
+
+    def has_distribution_perms(self, request, view, action, permission):
+        """
+        Check if the user has permissions on the corresponding distribution.
+        """
+        distributions = view.get_object().distributions
```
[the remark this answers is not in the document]
mdellweg: Yes, and i think, we support that workflow.

pulp_container/app/viewsets.py
mdellweg: Is this the push-repo? Then no. I want to check all the permissions by the roles on the distribution. Because you cannot assign a role to the repo and the distribution tied together. (as in "will be considered a single role assignment")

pulp_container/app/access_policy.py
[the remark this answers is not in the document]
ipanova: i mean.. we do not tie user's hands not do to that >.>

pulp_container/app/access_policy.py
mdellweg: Except it didn't really work that way...

goosemania:
@mdellweg , is it ready for review?

pulp_container/app/viewsets.py
```diff
@@ -985,26 +946,21 @@ class ContainerPushRepositoryVersionViewSet(
...
-                "principal": "authenticated",
-                "effect": "allow",
-            },
-            {
-                "action": ["retrieve"],
+                "action": ["list", "retrieve"],
                 "principal": "authenticated",
                 "effect": "allow",
-                "condition": "has_repo_attr_model_or_obj_perms:container.view_containerpushrepository",  # noqa
+                "condition": "has_distribution_perms:container.view_containerdistribution",
```
goosemania, line 957: Do we need to check namespace perms as well? 
It does not seem that the distribution perm check will look at the namespace perms.
I'm not sure though if in the current workflows it is possible to have a model or obj level permission for the namespace but not for the distribution of a push repo.

pulp_container/app/access_policy.py
```diff
@@ -24,15 +28,18 @@ def has_namespace_obj_perms(self, request, view, action, permission):
...
         elif type(obj) == models.ContainerPushRepository:
-            dists_qs = models.ContainerDistribution.objects.filter(repository=obj)
-            for dist in dists_qs:
-                if request.user.has_perm(permission, dist.namespace):
+            for dist in obj.distributions.all():
+                if request.user.has_perm(permission, dist.cast().namespace):
+                    return True
+        elif type(obj) == models.ContainerPushRepositoryVersion:
+            for dist in obj.repository.distributions.all():
+                if request.user.has_perm(permission, dist.cast().namespace):
```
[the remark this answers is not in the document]
mdellweg: The data model allows it. And maybe you can associate an existing push-repo with a (second) distribution. I'm not sure there.

pulp_container/app/viewsets.py
mdellweg: You are probably right.

goosemania:
To the question of a dedicated push repo permission vs distribution+namespace perms.
I do not have strong preference. In general, less permission types to deal with, the better.
However potential concerns with removing the dedicated push repo permission:
 -  a developer needs to remember that both distribution and namespace perms need a check
 -  a user who creates their own roles need to know those details as well

What do you think?

pulp_container/app/migrations/0023_data_translate_perms_to_roles.py
```diff
@@ -0,0 +1,241 @@
...
+                UserRole.objects.bulk_create(user_roles)
+                user_roles.clear()
+            if len(group_roles) > 1000:
+                GroupRole.objects.bulk_create(group_roles)
+                group_roles.clear()
+
+    # Find object permissions and transform into user/group roles
+    for permission_names, rolename in OBJECT_PERMISSION_ROLE_TRANSLATOR:
+        perms = [Permission.objects.filter(content_type__app_label=app_label, codename=codename).first() for app_label, codename in permission_names]
+        if all(perms):
```
goosemania, line 193: @mdellweg, is my understanding correct, that it's the best we can do and there can be individual object permissions left not translated to the roles? The case when `all` returns  `False`.

goosemania:
Needs a rebase, plz ping me for re-review

goosemania (approved):
Shiiiiiiip it! :tada:

PR #405, opened 2021-09-29
Bump ecdsa to 0.14 [noissue]

dannysauer:
Update to resolve relatively old timing vulnerability in ecdsa dependency.
https://snyk.io/vuln/SNYK-PYTHON-ECDSA-511942

Not sure if this should have a changelog entry and redmine record or not; lemme know if that needs to happen. :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
+# remember to also update unittest_requirements.txt when updating this file
 pulpcore>=3.15.0
-ecdsa~=0.13.2
+ecdsa~=0.14
 pyjwkest~=1.4.0
 pyjwt[crypto]~=1.7.1
 url-normalize~=1.4.2
--- a/unittest_requirements.txt
+++ b/unittest_requirements.txt
@@ -1,4 +1,7 @@
-ecdsa~=0.13.2
 mock
-pulpcore>=3.4
+# include the top-level requirements.txt contents below
+pulpcore>=3.15.0
+ecdsa~=0.14
 pyjwkest~=1.4.0
+pyjwt[crypto]~=1.7.1
+url-normalize~=1.4.2


mdellweg (approved):
Dependabot also works without changelogs. So i guess, it's ok.

ipanova:
not having a changelog entry for this PR is ok however you should use `[noissue]` in your commit in order to have CI green.

dannysauer:
Cool.  PR and commit titles both updated.

ipanova:
Here is one more place where the dep needs to be updated https://github.com/pulp/pulp_container/blob/master/unittest_requirements.txt
Thank you for the PR!

dannysauer:
> Here is one more place where the dep needs to be updated https://github.com/pulp/pulp_container/blob/master/unittest_requirements.txt Thank you for the PR!

Whoops, I completely missed that.

It looks like unittest should just match requirements.txt with mock added.  To prevent future similar oversights (like how pulpcore is oddly a way older version and a couple other modules are missing), would just `-r requirements.txt` perhaps be an appropriate fix?  I went ahead and did that.  Lemme know if there's a problem with that and I'll undo that / stay in my lane. :)

dannysauer:
Clearly that didn't work so well, what with the whole "files copied to /tmp" code in scripts/script.sh.  It says that test script is generated from a template, so I hate to go fiddling around with the script and break something in the future.

Better to make a low-tech comment in requirements.txt to update unittest_requirements.txt, then?  Or alter the script so requirements.txt is copied to the appropriate place?

mdellweg:
> Clearly that didn't work so well, what with the whole "files copied to /tmp" code in scripts/script.sh. It says that test script is generated from a template, so I hate to go fiddling around with the script and break something in the future.
> 
> Better to make a low-tech comment in requirements.txt to also update unittest_requirements.txt on change, then? Or alter the script so requirements.txt is copied to the appropriate place?

How about doing the easy fix for the actual problem now (bumping in both files), and think about a better solution in another place. Maybe we should use dependabot for this anyway. wdyt @ipanova

ipanova:
> > Clearly that didn't work so well, what with the whole "files copied to /tmp" code in scripts/script.sh. It says that test script is generated from a template, so I hate to go fiddling around with the script and break something in the future.
> > Better to make a low-tech comment in requirements.txt to also update unittest_requirements.txt on change, then? Or alter the script so requirements.txt is copied to the appropriate place?
> 
> How about doing the easy fix for the actual problem now (bumping in both files), and think about a better solution in another place. Maybe we should use dependabot for this anyway. wdyt @ipanova

bumping the version in both files works for me, we can leave a comment in the requirements file as @dannysauer  suggested.

dannysauer:
> Maybe we should use dependabot for this anyway

FWIW, I've been experimenting with Snyk recently.  Dependabot is nice because it automatically rebases PRs when main changes, and stays bleeding-edge with modules.  Snyk's automatic security PRs have much better detail, IMO, and they just bump to the version which fixes the security issue rather than absolute newest.  They don't seem to automatically rebase on main/master when it's updated, though.  Maybe an action would fix that.  Either way, Dependabot missed this being a security issue, so I initially ignored it.

Here are examples of the Snyk PR opened against my Docker Repo for the ecdsa issue v/s the Dependabot PR:
Kong/docker-pulp#227
Kong/docker-pulp#87

Sorry for the semi-relevant diversion.  Back to discussing this PR now. :)

PR #406, opened 2021-09-30
Azure backend

ipanova:


pulp_container/app/redirects.py
```diff
@@ -114,3 +114,16 @@ def redirect_to_s3_storage(self, artifact, return_media_type):
...
+class AzureStorageRedirects(S3StorageRedirects):
+    """
+    A class that implements methods for the direct retrieval of manifest objects.
+    """
+
+    def redirect_to_object_storage(self, artifact, return_media_type):
+        """
+        Redirect to the passed artifact's file stored in the Azure storage.
+        """
+        content_url = artifact.file.storage.url(artifact.file.name)
```
ipanova, line 128: The tests would not pass until we figure out how to set needed headers. There does not seem to be any way how pass/set them in url redirect.
The only option( I've found so far) is to set the params on the object creation and for that we'd need to subclass the backend AzureStorage.get_object_parameters.
https://github.com/jschneier/django-storages/blob/master/storages/backends/azure_storage.py#L289
https://django-storages.readthedocs.io/en/latest/backends/azure.html

ipanova:
I filed this issue to enable azure storage for Container Registry https://pulp.plan.io/issues/9488?

pulp_container/app/redirects.py
benthomasson: @ipanova how did subclassing AzureStorage go?

fao89: after https://github.com/jschneier/django-storages/pull/1071 gets merged:
```python
filename = os.path.basename(artifact.file.name)
parameters = {
    "content_type": return_media_type,
    "content_disposition": f"attachment; filename={filename}",
}
content_url = artifact.file.storage.url(artifact.file.name, parameters=parameters)
```
I tested and it worked:
https://github.com/fao89/pulp_container/runs/3875246508?check_suite_focus=true

PR #409, opened 2021-10-06
Update plugin ci for release branch.

ipanova:
[noissue]


mdellweg:
I think you need to change the target branch

PR #413, opened 2021-10-06
Fixed a failure during distribution update that occured when unsetting repository_version.

ipanova:
…_version.

closes #9434


CHANGES/9434.bugfix
```diff
@@ -0,0 +1 @@
+Updated distribution validation to handle properly usage of repo/repo_version.
```
mdellweg, line 1:
```suggestion
Updated the distribution validation to properly handle the use of ``repository`` / ``repository_version``.
```

PR #414, opened 2021-10-08
Fixed blob media_type during image push.

ipanova:
closes #9229

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -115,6 +115,24 @@ class BlobNotFound(NotFound):
         )
 
 
+class BlobInvalid(ParseError):
+    """Exception to render a 400 with the code 'BLOB_UNKNOWN'"""
+
+    def __init__(self, digest):
+        """Initialize the exception with the blob digest."""
+        super().__init__(
+            detail={
+                "errors": [
+                    {
+                        "code": "BLOB_UNKNOWN",
+                        "message": "blob unknown to registry",
+                        "detail": {"digest": digest},
+                    }
+                ]
+            }
+        )
+
+
 class ManifestNotFound(NotFound):
     """Exception to render a 404 with the code 'MANIFEST_UNKNOWN'"""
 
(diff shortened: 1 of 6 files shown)


pulp_container/app/registry_api.py
```diff
@@ -814,14 +850,36 @@ def put(self, request, path, pk=None):
...
+        except models.Blob.DoesNotExist:
+            raise BlobNotFound(digest=config_digest)
+        config_media_type = config_layer.get("mediaType")
+        if config_media_type not in (
+            models.MEDIA_TYPE.CONFIG_BLOB,
+            models.MEDIA_TYPE.CONFIG_BLOB_OCI,
+        ):
+            raise BlobInvalid(digest=config_blob.digest)
+        config_blob.media_type = config_media_type
+        config_blob.save()
```
mdellweg, line 879: I'd really prefer if we didn't update existing content units.

pulp_container/app/registry_api.py
```diff
@@ -836,15 +894,25 @@ def put(self, request, path, pk=None):
...
+            blob_media_type = blobs[blob.digest]
+            if blob_media_type not in (
+                models.MEDIA_TYPE.REGULAR_BLOB,
+                models.MEDIA_TYPE.REGULAR_BLOB_OCI,
+            ):
+                raise BlobInvalid(digest=blob.digest)
+            blob.media_type = blob_media_type
             thru.append(models.BlobManifest(manifest=manifest, manifest_blob=blob))
         models.BlobManifest.objects.bulk_create(objs=thru, ignore_conflicts=True, batch_size=1000)
+        models.Blob.objects.bulk_update(objs=blobs_qs, fields=["media_type"])
```
mdellweg, line 915: The same here. Is there a way to have the Blob content type available when creating the blob?

CHANGES/9229.bugfix
```diff
@@ -0,0 +1 @@
+Fixed blob media_type during image push.
```
ipanova, line 1: @mdellweg I am not sure what to do with it. I learned that the blob_media_type is not important to client at all, instead it cares about the content_type returned, which I've fixed.
I suggest to remove this changelog entry from the PR and deal with 9229 separately - not sure how, we can probably keep blob.media_type field on the model but treat it as informative only, or deprecate entirely.

mdellweg: If all the bugs are fixed without that, we should remove it and get the rest shipped.
When we realize that we really ignore the media_type field, we can remove it eventually. But if we can keep that migration out of this PR, we get more flexibility for backports. Right?

ipanova: exactly

mdellweg (approved):

PR #416, opened 2021-10-13
Test.

ipanova:
[noissue]


fao89:
I think this is the reason CI is failing
https://github.com/pulp/pulpcore/pull/1679

PR #430, opened 2021-11-15
Fixed codes and headers for upload and blob responses.

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -153,7 +153,7 @@ class UploadResponse(Response):
     This response object provides information about Uploads during 'push' operations.
     """
 
-    def __init__(self, upload, path, request):
+    def __init__(self, upload, path, request, status=202):
         """
         Args:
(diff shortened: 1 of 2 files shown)


pulp_container/app/registry_api.py
```diff
@@ -162,10 +162,14 @@ def __init__(self, upload, path, request):
...
+        else:
+            offset = int(upload.size - 1)
         headers = {
             "Docker-Upload-UUID": upload.pk,
             "Location": f"/v2/{path}/blobs/uploads/{upload.pk}",
-            "Range": "0-{offset}".format(offset=int(upload.size - 1)),
+            "Range": "0-{offset}".format(offset=offset),
             "Content-Length": 0,
         }
         super().__init__(headers=headers, status=202)
```
mdellweg, line 175:
```suggestion
        super().__init__(headers=headers, status=status)
```

mdellweg (approved):

PR #436, opened 2021-11-23
Added an ability to sync image signatures from a provided sigstore.

goosemania:
closes #9507


pulp_container/app/models.py
```diff
@@ -200,6 +200,48 @@ class Meta:
...
+        signed_manifest (models.ForeignKey): An image manifest this signature is relevant to.
+
+    """
+    PROTECTED_FROM_RECLAIM = False
+
+    TYPE = "signature"
+
+    manifest_digest = models.CharField(max_length=255, db_index=True)
+    identity_reference = models.CharField(max_length=255)
+    signature_type = models.CharField(max_length=255)  # maybe make it a choice field with one
```
ipanova, line 229: +1

pulp_container/app/models.py
```diff
@@ -200,6 +200,48 @@ class Meta:
...
+                                                       # choice so far "atomic"
+    signing_key_id = models.CharField(max_length=255, db_index=True)
+    signature_timestamp = models.PositiveIntegerField()
+
+    signed_manifest = models.ForeignKey(
```
ipanova, line 234: can we also add `creator`?

ipanova: I learned more about this field and we propbably will not gain much by exposing it on the model - it identifies the name and version of the software which has created the signature, not the top-level caller so in our case it will aclways be 'atomic' since we use skopeo https://github.com/containers/skopeo/blob/99621f416892746bd8d223fa5c4773267124e9cd/vendor/github.com/containers/image/v5/signature/signature.go#L70

goosemania: maybe it's the case with skopeo signing but other signatures may contain something more useful. E.g. from rh registry I get `{"creator": "Red Hat RCM Pub 0.67.0-final"}`

goosemania: I'm fine with leaving it out as well.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +379,117 @@ def create_blob(self, blob_data, deferred_download=True):
...
+
+        For sigstore, signatures are downloaded from a specified url. The number of signatures is
+        unknown upfront, need to download them in order by incrementing the index until hit 404.
+
+        Args:
+            man_dc: Declarative content instance of a related Manifest
+            signature_source: Source where to get signatures from
+        """
+
+        def _extract_data_from_signature(signature_raw):
```
goosemania, line 396: Needs to go to utils

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +379,117 @@ def create_blob(self, blob_data, deferred_download=True):
...
+            sig_json["signature_timestamp"] = crypt_obj.timestamp
+            return sig_json
+
+        signature_dcs = []
+        if signature_source == SIGNATURE_SOURCE.SIGSTORE:
+            man_digest_reformatted = man_dc.content.digest.replace(":", "=")
+            # should the namespaced_upstream_name be used to grab a signature?
+            sig_relative_baseurl = f"{self.remote.upstream_name}@{man_digest_reformatted}"
+
+            async with aiohttp.ClientSession() as session:
```
goosemania, line 441: There are 2 places/cases where I need to download stuff and I can't use a downloader from a remote.
1. HEAD request to a public server (no auth configured on a remote should be used but I guess proxy configuration needs to be in place).
2. Here. Need to do multiple downloads from a public server (again no auth but proxy is needed). Number of downloads is not clear upfront. URL contains an index which needs to be incremented for each download attempt until you hit HTTP 404, then stop.

Any suggestions to handle this?
The goal is to reuse sessions and have cleaner code.

Custom downloaders for each case?
Do we want to extend plugin API to add some generic downloaders, where one can specify http method and have a session to reuse but without preset configuration from a remote?

@ipanova @mdellweg @lubosmj @dkliban @bmbouter

pulp_container/app/serializers.py
```diff
@@ -214,8 +247,18 @@ class ContainerRemoteSerializer(RemoteSerializer):
         default=Remote.IMMEDIATE,
     )
 
+    # TODO: add url validation
+    sigstore = serializers.CharField(
```
goosemania, line 251: Figure out why it comes out as required field

pulp_container/app/serializers.py
```diff
@@ -125,6 +125,39 @@ class Meta:
...
+    digest = serializers.CharField(help_text="sha256 digest of the signature blob")
+    signature_type = serializers.CharField(help_text="Container signature type, e.g. 'atomic'")
+    signing_key_id = serializers.CharField(help_text="Signing key ID")
+    signature_timestamp = serializers.IntegerField(help_text="Timestamp of a signature")
+    signature_creator = serializers.CharField(help_text="Signature creator")
+    signed_manifest = DetailRelatedField(
+        many=False,
+        help_text="Manifest that is signed",
+        view_name="container-manifests-detail",
+        queryset=models.ImageManifestSignature.objects.all(),
```
ipanova, line 145: Manifest.objects.all()

pulp_container/app/serializers.py
```diff
@@ -125,6 +125,39 @@ class Meta:
...
+        fields = NoArtifactContentSerializer.Meta.fields + (
+            "name",
+            "digest",
+            "signature_type",
+            "signing_key_id",
+            "signature_timestamp",
+            "signature_creator",
+            "signed_manifest",
+        )
+        model = models.Tag
```
ipanova, line 158: typo, it's should signature model

pulp_container/app/models.py
```diff
@@ -200,6 +200,50 @@ class Meta:
         unique_together = (("name", "tagged_manifest"),)
 
 
+class ImageManifestSignature(Content):
```
ipanova, line 203: my suggestion would be to rename thi class to a more generic `ManifestSignature`, i  believe `container/image` library does not distinguish between image and list.

pulp_container/app/models.py
```diff
@@ -200,6 +200,50 @@ class Meta:
...
+
+    PROTECTED_FROM_RECLAIM = False
+
+    TYPE = "signature"
+
+    SIGNATURE_CHOICES = ((SIGNATURE_TYPE.ATOMIC_SHORT, SIGNATURE_TYPE.ATOMIC_SHORT),)
+
+    name = models.CharField(max_length=255)
+    digest = models.CharField(max_length=255)
+    signature_type = models.CharField(max_length=255, choices=SIGNATURE_CHOICES)
```
ipanova, line 230: this is probably matter of preference, i'd not prepend those fields wit `signature_` since they are already on the signature model

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +379,117 @@ def create_blob(self, blob_data, deferred_download=True):
...
+                    _(
+                        "It is not possible to read the signed document, GPG error: {}".format(
+                            crypt_obj.stderr
+                        )
+                    )
+                )
+                return
+
+            try:
+                sig_json = json.loads(crypt_obj.data)
```
ipanova, line 410: we should add signature validation of the loaded json

pulp_container/app/tasks/sync_stages.py
```diff
@@ -140,6 +162,9 @@ async def run(self):
                     await self.put(man_dc)
                     tag_dc.extra_data["tagged_manifest_dc"] = man_dc
                     await self.handle_blobs(man_dc, content_data)
+                    if signature_source is not None:
+                        signature_dcs.extend(await self.create_signatures(man_dc, signature_source))
```
ipanova, line 166: I am not sure whether our rh registry contains any manifest lists to test, but we should pull in signatures for manifest lists as well.
The question that we need to answer is whether to pull in signatures per every image manifest that's in that manifest list or not. Maybe in reality only manifest list is signed, I asked Miroslav on more details.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +379,117 @@ def create_blob(self, blob_data, deferred_download=True):
...
+                return
+
+            sig_json["signing_key_id"] = crypt_obj.key_id
+            sig_json["signature_timestamp"] = crypt_obj.timestamp
+            return sig_json
+
+        signature_dcs = []
+        if signature_source == SIGNATURE_SOURCE.SIGSTORE:
+            man_digest_reformatted = man_dc.content.digest.replace(":", "=")
+            # should the namespaced_upstream_name be used to grab a signature?
```
ipanova, line 438: This was done to follow dockerhub's `library` conventions. We can leave this as you have since docker has docker content trust for signing, which we do not support https://docs.docker.com/engine/security/trust/

pulp_container/app/tasks/sync_stages.py
ipanova: Here is what I learned- the image verification code only cares about signatures on the per-arch images; manifest lists might have signatures but they don’t really matter.
So, we should for sure  mirror signatures for every manifest that is mirrored, incl. the subordinate per-arch ones.

goosemania: Thanks for asking Miroslav. I took the docs as it's only supported for images https://github.com/containers/image/blob/main/docs/containers-signature.5.md#high-level-overview. So I made clear almost everywhere that it for images only, incl. docstrings.

I understand that at tech level, it should be the same. It affects tests and general expectations.

pulp_container/app/tasks/sync_stages.py
goosemania: Please elaborate what kind of validation you expect?

ipanova: we should assert that everything what's in 'critical' is there and refuse when unrecognized or unsupported is encountered https://github.com/containers/image/blob/main/docs/containers-signature.5.md#json-processing-and-forward-compatibility

ipanova: let's capture this in a separate ticket and implement later not in mvp

pulp_container/app/tasks/sync_stages.py
dralley: I'm collecting greivances with the current downloader system on / attached to this issue: https://pulp.plan.io/issues/7352

I agree that the current download infrastructure is not ideal.

bmbouter: I want to help, but I'm a little lost. Why in this case is the remote not the right way to generate the downloader?

lubosmj: @bmbouter, because we are trying to download data, e.g., from an untrusted source. We may either get rid of some of the attributes (secrets) within the remote's downloader and renew them later or build a customized downloader that will have a separate session shared during the run.

bmbouter: I see. Well in terms of helping you get where you need to go, if you don't want the downloader configured from the remote, you'll have to construct it some other way. A shared session would be good also +1 to that.

What about having a customized factory that produces these types of downloaders? That is kind of the intent of that object, to produce configured downloaders. Or do folks think we should use this challenge to guide architectural changes to the downloaders now? I'm inclined to fix this with what we already have in place, but if folks want to redesign now I'm ok with that too.

As an aside, I'm so confused. Why would we trust content from an untrusted source?

goosemania: I'll try to expand on what @lubosmj  said and what I originally posted and seem to fail to explain properly.

To my understanding, if I generate a downloader from a remote:
  - the session will [get all the tls configuration from it](https://github.com/pulp/pulpcore/blob/master/pulpcore/download/factory.py#L107-L126)
  - the session will [use auth and proxy configuration set on a remote](https://github.com/pulp/pulpcore/blob/master/pulpcore/download/http.py#L272)

Use case:
In addition to a usual source of content which remote is configured for, there is a separate resource for signatures for the content which remote refers it. Signatures are on a public server and do not need any auth to be provided.

More specific needs:
 - to make a HEAD HTTP request to the public server without any auth and use proxy if it's configured on a remote. (Not sure if a "downloader" name is suitable for this need since it's not a GET request)
 - make a many GET requests, N per each content, to the same public server with no auth and use proxy if it's configured on a remote.

Is it helpful or still confusing?

mdellweg: > As an aside, I'm so confused. Why would we trust content from an untrusted source?

If it's a digital signature, it brings means of validation for itself, and there's no need to trust the source it was obtained from.

bmbouter: Ah this makes sense re the signature server not needing to be trusted. Thanks.

Just to confirm, these signatures are only fetched at sync-time? Or are they fetched on-demand or streamed via the content app?

bmbouter: After looking this over more with @goosemania I think two things:

1) For now, subclass the DownloaderFactory and use that to generate these downloaders. It will involve copying some code from the `_make_aiohttp_session_from_remote`method until we make that more granular.

2) @dralley what I think really needs improvement is the `DownloaderFactory`. We should find ways to subclass it with little to no code copying. To me that's at least one of the main issues.

pulp_container/app/downloaders.py
```diff
@@ -159,3 +165,77 @@ def auth_header(token, basic_auth):
...
+
+def user_agent():
+    """
+    Produce a User-Agent string to identify Pulp and relevant system info.
+    """
+    pulp_version = get_distribution("pulpcore").version
+    python = "{} {}.{}.{}-{}{}".format(sys.implementation.name, *sys.version_info)
+    uname = platform.uname()
+    system = f"{uname.system} {uname.machine}"
+    return f"pulpcore/{pulp_version} ({python}, {system}) (aiohttp {aiohttp_version})"
```
goosemania, line 178: Copy-pasted from pulpcore. I suggest to add it to plugin API.

pulp_container/app/models.py
```diff
@@ -320,26 +321,28 @@ class ContainerRemote(Remote, AutoAddObjPermsMixin, AutoDeleteObjPermsMixin):
...
+    sigstore = models.TextField(null=True)
 
     TYPE = "container"
     ACCESS_POLICY_VIEWSET_NAME = "remotes/container/container"
 
     @property
     def download_factory(self):
         """
-        Return the DownloaderFactory which can be used to generate asyncio capable downloaders.
+        Downloader Factory that maps to custom downloaders which support registry auth.
```
goosemania, line 342: Due to flake8 nitpick: D403 First word of the first line should be properly capitalized

pulp_container/app/models.py
```diff
@@ -382,6 +405,45 @@ def get_downloader(self, remote_artifact=None, url=None, **kwargs):
         kwargs["remote"] = self
         return super().get_downloader(remote_artifact=remote_artifact, url=url, **kwargs)
 
+    def get_noauth_downloader(self, remote_artifact=None, url=None, **kwargs):
```
goosemania, line 408: The only difference from the pulpcore method is the download factory which is used. I suggest to change pulpcore method to accept an optional parameter `download_factory` and keep self.download_factory` as default.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -68,11 +72,31 @@ async def run(self):
...
+
+        async def get_signature_source(self):
+            """
+            Find out where signatures come from: sigstore, extension API or not available at all.
+            """
+            if self.remote.sigstore:
+                return SIGNATURE_SOURCE.SIGSTORE
+
+            registry_v2_url = urljoin(self.remote.url, "v2")
+            extension_check_downloader = self.remote.get_noauth_downloader(url=registry_v2_url)
```
goosemania, line 84: To simplify our lives and not to create a downloader for HEAD request, I switched to using GET here, instead of HEAD.
HEAD is cheaper but this is called only once during sync.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +385,75 @@ def create_blob(self, blob_data, deferred_download=True):
...
+                        log.info(
+                            _(
+                                "{} is not accessible, can't sync an image signature. "
+                                "Error: {} {}".format(signature_url, exc.status, exc.message)
+                            )
+                        )
+                    # 404 is fine, it means there are no or no more signatures available
+                    break
+
+                with open(signature_download_result.path, "rb") as f:
```
goosemania, line 425: The downside (read "suboptimal behaviour") of using standard downloader is that result is saved to a file while we do not need an artifact. The signatures are small (under 1 KB), we can read it directly from the response.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +385,75 @@ def create_blob(self, blob_data, deferred_download=True):
...
+
+                signature_counter += 1
+                signature_json = extract_data_from_signature(signature_raw, man_dc.content.digest)
+                if signature_json is None:
+                    continue
+
+                sig_digest = hashlib.sha256(signature_raw).hexdigest()
+                signature = ManifestSignature(
+                    name=f"{man_dc.content.digest}@{sig_digest[:32]}",
+                    digest=sig_digest,
```
ipanova, line 436: can you prepend 'sha256'?
```
sig_digest = hashlib.sha256(data).hexdigest()
signature_json["digest"] = f"sha256:{sig_digest}"
```

pulp_container/app/tasks/sync_stages.py
goosemania: https://pulp.plan.io/issues/9597

pulp_container/app/tasks/sync_stages.py
goosemania: I think you meant to prepend the manifest digest and not the signature digest.
Please check the changes if they look ok now.

goosemania: Never mind, I see this format for manifests.

pulp_container/app/serializers.py
```diff
@@ -696,3 +707,17 @@ class Meta:
...
             "artifacts",
         )
+
+
+class ContainerRepositorySyncURLSerializer(RepositorySyncURLSerializer):
+    """
+    Serializer for Container Sync.
+    """
+
+    signed_only = serializers.BooleanField(
```
goosemania, line 717: Should this option affect only the current sync or should it remove content with no signatures from the existing repo version as well?
I added it as a one-time option to a sync operation. Should be a permanent one? on a repo maybe?

ipanova: i think it should affect only current sync. it user would want to remove content with no signature from the existing repo version he'd use mirror=True, wdyt?

goosemania: Sounds reasonable

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +410,75 @@ def create_blob(self, blob_data, deferred_download=True):
...
+                    key_id=signature_json["signing_key_id"],
+                    timestamp=signature_json["signature_timestamp"],
+                    creator=signature_json["optional"].get("creator"),
+                    data=base64.b64encode(signature_raw),
+                )
+                sig_dc = DeclarativeContent(
+                    content=signature,
+                    extra_data={"signed_manifest_dc": man_dc},
+                )
+                signature_dcs.append(sig_dc)
```
lubosmj, line 472: Can we call `resolution()` right here to not wait until all signatures are collected? Maybe, queuing manifest_dc after calling `resolution()` will cause a deadlock (line no 174, 195). We need to take care of that.

lubosmj (commented):
Great job! I left a couple of comments that mainly focus on details. The functionality looks good to me at first sight.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -349,6 +410,75 @@ def create_blob(self, blob_data, deferred_download=True):
...
+        For sigstore, signatures are downloaded from a specified url. The number of signatures is
+        unknown upfront, need to download them in order by incrementing the index until hit 404.
+
+        Args:
+            man_dc: Declarative content instance of a related Manifest
+            signature_source: Source where to get signatures from
+        """
+        signature_dcs = []
+        if signature_source == SIGNATURE_SOURCE.SIGSTORE:
+            man_digest_reformatted = man_dc.content.digest.replace(":", "=")
```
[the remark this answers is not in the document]
lubosmj: Aha, this is not related to the name of a signature.

pulp_container/app/downloaders.py
```diff
@@ -159,3 +161,66 @@ def auth_header(token, basic_auth):
...
+
+        headers = MultiDict({"User-Agent": NoAuthDownloaderFactory.user_agent()})
+        if self._remote.headers is not None:
+            for header_dict in self._remote.headers:
+                user_agent_header = header_dict.pop("User-Agent", None)
+                if user_agent_header:
+                    headers["User-Agent"] = f"{headers['User-Agent']}, {user_agent_header}"
+                headers.extend(header_dict)
+
+        conn = aiohttp.TCPConnector(**tcp_conn_opts)
```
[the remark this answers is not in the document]
goosemania: It's the way it is done in pulpcore, so it helps a bit to quickly see how it differs from pulpcore.
I can pass it directly, just sharing the reason/thinking behind the current way.

pulp_container/app/models.py
```diff
@@ -320,26 +320,28 @@ class ContainerRemote(Remote, AutoAddObjPermsMixin, AutoDeleteObjPermsMixin):
...
+        exclude_tags (fields.ArrayField): List of tags to exclude during sync.
+        sigstore (models.TextField): The URL to a SigStore where signatures of container images
+            should be synced from.
     """
 
     upstream_name = models.CharField(max_length=255, db_index=True)
     include_foreign_layers = models.BooleanField(default=False)
     include_tags = fields.ArrayField(models.CharField(max_length=255, null=True), null=True)
     exclude_tags = fields.ArrayField(models.CharField(max_length=255, null=True), null=True)
+    sigstore = models.TextField(null=True)
```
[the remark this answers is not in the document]
goosemania: I actually wanted to make it the same as in pulpcore, TextFiled. But you bring a good point, I added the question to the open floor agenda.

pulp_container/app/utils.py
```diff
@@ -16,3 +28,74 @@ def get_accepted_media_types(headers):
...
+    except Exception as exc:
+        log.info(
+            _(
+                "Signed document cannot be parsed to create a signature for {}."
+                " Error: {}".format(man_digest, str(exc))
+            )
+        )
+        return
+
+    signature_type = sig_json["critical"].get("type") if "critical" in sig_json else None
```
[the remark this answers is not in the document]
goosemania: There is a separate story to add validation for this spec https://pulp.plan.io/issues/9597

pulp_container/app/utils.py
```diff
@@ -16,3 +28,74 @@ def get_accepted_media_types(headers):
...
+
+    Args:
+        signature_raw(bytes): A signed doc to get data from
+        man_digest (str): A manifest digest for which the signature is for
+
+    Returns:
+        dict: JSON representation of the document and available data about signature
+
+    """
+    gpg = gnupg.GPG()
```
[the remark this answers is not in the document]
goosemania: In a relatively small repo there was 156 signature (26 manifests, 6 signatures per each)
Not sure, how big repos can be and what is the reasonable number of signatures.

1000 manifests with 10 signatures each will add 30 seconds for the whole sync process.

 >>> timeit(gnupg.GPG, number=1)
0.016664049122482538
>>> timeit(gnupg.GPG, number=156)
0.5067231778521091
>>> timeit(gnupg.GPG, number=10**3)
2.937312678899616
>>> timeit(gnupg.GPG, number=10**4)
29.107102069072425

I can make this call a singleton, but I'm not sure it's worth the complexity.
Thoughts?

pulp_container/app/downloaders.py
```diff
@@ -159,3 +161,66 @@ def auth_header(token, basic_auth):
...
+        connect = self._remote.connect_timeout
+
+        timeout = aiohttp.ClientTimeout(
+            total=total, sock_connect=sock_connect, sock_read=sock_read, connect=connect
+        )
+        return aiohttp.ClientSession(connector=conn, timeout=timeout, headers=headers)
+
+    def _http_or_https(self, download_class, url, **kwargs):
+        """
+        Same as DownloaderFactory._http_or_https, excluding using the basic auth credentials.
```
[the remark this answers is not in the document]
goosemania: :+1:

pulp_container/app/tasks/sync_stages.py
```diff
@@ -126,20 +160,42 @@ async def run(self):
...
                     )
-                    await self.put(list_dc)
-                    tag_dc.extra_data["tagged_manifest_dc"] = list_dc
+                    at_least_one_manifest = False
                     for manifest_data in content_data.get("manifests"):
                         man_dc = self.create_manifest(list_dc, manifest_data)
+                        if signature_source is not None:
+                            man_sig_dcs = await self.create_signatures(man_dc, signature_source)
+                            if self.signed_only and not man_sig_dcs:
+                                # do not pass down the pipeline unsigned manifests
```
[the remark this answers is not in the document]
goosemania: It should not cause any problems with association with tags or manifest lists within the pipeline because we are not passing the dc further.
As for admins, it might be friendly to log that certain manifests have been skipped due to absence of a signature. I'll add it at info level but only here, for per-arch images. Is that what you had in mind?

pulp_container/app/utils.py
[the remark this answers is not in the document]
lubosmj: How about adding `gpg = gnupg.GPG()`  to the top of the module?

pulp_container/app/tasks/sync_stages.py
lubosmj: Yes, I meant 175, sorry.

mdellweg: If i understand the data flow in this pipeline correctly, signatures are handled like tags, so sig dc's are collected until their manifest is ready and only the ready content will be injected into the pipeline. This looks right to me.

pulp_container/app/models.py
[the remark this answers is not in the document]
goosemania: Info from @fao89: we had an OpenAPI issue: https://github.com/pulp/pulpcore/pull/919
but we killed the OpenAPI schema view, so we can use it again.

It was discussed at the open floor today that we do not want to change it in pulpcore.
I would keep TextField since I'm not sure which limit to set for the length, and pulpcore has unlim.

If you feel strongly about this, feel free to let me know before we merge the `signing` branch into the main (probably not any time soon :))

pulp_container/app/serializers.py
```diff
@@ -247,8 +248,18 @@ class ContainerRemoteSerializer(RemoteSerializer):
         default=Remote.IMMEDIATE,
     )
 
+    # TODO: add url validation
+    sigstore = serializers.CharField(
+        required=False, help_text=_("A URL to a sigstore to download image signatures from")
```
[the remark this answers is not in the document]
goosemania: I'll use URLValidator to avoid the length limit.

pulp_container/app/models.py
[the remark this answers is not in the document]
lubosmj: We can go with `TextField` for now. Thank you for adding the validation.

lubosmj (approved):

pulp_container/app/tasks/sync_stages.py
```diff
@@ -126,20 +160,54 @@ async def run(self):
...
                         future_manifests.append(man_dc)
                         man_dcs[man_dc.content.digest] = man_dc
                         await self.put(man_dc)
+
+                    if not at_least_one_manifest:
+                        # no manifests for this manifest list and tag were signed, do not pass
+                        # this manifest list and tag down the pipeline
+                        continue
+
+                    tag_dc.extra_data["tagged_manifest_dc"] = list_dc
```
ipanova, line 194: tag_dc is being appended to the tag_dcs list  to go through and resolve its `tagged_manifest_dc`regardless of whether the list_dc( or manifest_dc) was sent through the pipeline or not.

if i read correctly, this line will fail because it expects to have `tagged_manifest_dc` among the extra_data https://github.com/pulp/pulp_container/pull/436/files#diff-b702cbd3ebdff19bf8e4dfc14688aa3fb36e1a47ef6cf84d24e0f75b61247d36R224 which is not there.
we should change that line to `tagged_manifest_dc = tag_dc.extra_data.get("tagged_manifest_dc")`

pulp_container/app/tasks/sync_stages.py
ipanova: I agree

pulp_container/app/tasks/sync_stages.py
goosemania: I do not think tag_dc is being appended if manifest list or manifest do not go through, at least it was my intention not to sync that tag at all.

Line-192 `continue` makes sure that manifest list and its tag are not passed further (including the appending to tag_dcs).
Line-205 `continue` makes sure that that a tag of a manifest is not passed further (including the appending to tag_dcs).
I expect the `tagged_manifest_dc` to always have a manifest in extra_data.

Let me know if I misunderstood your concern.

ipanova: no, my concern was understood correctly. it was difficult to read nested loops, `continue` on lines 192 and 205 match the out loop so all is :+1:

ipanova (approved):

PR #437, opened 2021-11-25
Bugfix for container clients failing to retrieve artefacts from S3 storage

m-bull:
Issue described [here](https://pulp.plan.io/issues/9586).

### DIFF
--- a/CHANGES/9586.bugfix
+++ b/CHANGES/9586.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug that caused container clients to be unable to interact with content stored on S3.
--- a/pulp_container/app/redirects.py
+++ b/pulp_container/app/redirects.py
@@ -112,7 +112,9 @@ class S3StorageRedirects(CommonRedirects):
             "ResponseContentType": return_media_type,
             "ResponseContentDisposition": f"attachment;filename={filename}",
         }
-        content_url = artifact.file.storage.url(artifact.file.name, parameters=parameters)
+        content_url = artifact.file.storage.url(
+            artifact.file.name, parameters=parameters, http_method=self.request.method
+        )
         return redirect(content_url)
 
 


lubosmj:
@m-bull, would you mind squashing those two commits into a single one and adding a commit message in the format described at https://docs.pulpproject.org/pulpcore/contributing/git.html#commit-message (closes #9586)?

CHANGES/9586.bugfix
```diff
@@ -0,0 +1 @@
+Container clients fail to retrieve artefacts from S3 storage
```
lubosmj, line 1: Fixed a bug that caused container clients to be unable to interact with content stored on S3.

pulp_container/app/redirects.py
```diff
@@ -112,7 +112,9 @@ def redirect_to_object_storage(self, artifact, return_media_type):
             "ResponseContentType": return_media_type,
             "ResponseContentDisposition": f"attachment;filename={filename}",
         }
-        content_url = artifact.file.storage.url(artifact.file.name, parameters=parameters)
+        content_url = artifact.file.storage.url(
+            artifact.file.name, parameters=parameters, http_method=self.request.method
+        )
```
mdellweg, line 117: Should we add a check, that method is in `rest_framework.permissions.SAFE_METHODS`?

ipanova:
@m-bull I was puzzled on how our tests which also run on S3 pass, and after closely looking at them I found that we actually never push twice same content to the same repo. We need to add a test that would be pushing same image twice, for example here https://github.com/pulp/pulp_container/blob/main/pulp_container/tests/functional/api/test_push_content.py#L81 adding 2 consecutive pushes should cover the bugfix you've submitted.

pulp_container/app/redirects.py
ipanova: there is no need to check for safe_methods, redirects are happening anyway only on head/get calls https://github.com/pulp/pulp_container/blob/main/pulp_container/app/registry_api.py#L800 and https://github.com/pulp/pulp_container/blob/main/pulp_container/app/registry_api.py#L832

ipanova: @m-bull we should add this to the core codepath as well https://github.com/pulp/pulpcore/blob/master/pulpcore/content/handler.py#L780

ipanova:
I am merging this and will create a follow-up pr for the mentioned test. Thank you for the contribution!
CI failure are unrelated to the PR changes and should already be fixed on main

PR #439, opened 2021-11-29
Add ManifestSignature model and related serializer/filter/viewset.

goosemania:
re #9507
https://pulp.plan.io/issues/9507


pulp_container/app/models.py
```diff
@@ -200,6 +200,50 @@ class Meta:
...
+    """
+
+    PROTECTED_FROM_RECLAIM = False
+
+    TYPE = "signature"
+
+    SIGNATURE_CHOICES = ((SIGNATURE_TYPE.ATOMIC_SHORT, SIGNATURE_TYPE.ATOMIC_SHORT),)
+
+    name = models.CharField(max_length=255, db_index=True)
+    digest = models.CharField(max_length=255)
```
ipanova, line 229: maybe worth adding db_index

goosemania: it's indexed, it's in the uniqueness constraint below, and on its own.

pulp_container/app/viewsets.py
```diff
@@ -100,6 +100,20 @@ class Meta:
         }
 
 
+class ManifestSignatureFilter(ContentFilter):
+    """
+    FilterSet for image signatures.
+    """
+
```
ipanova, line 107: it might be worth adding a filter for the signed manifest digest in a similar manner it's done here https://github.com/pulp/pulp_container/blob/main/pulp_container/app/viewsets.py#L66

ipanova (approved):
thank you!

PR #440, opened 2021-11-30
Fix queryset scoping

mdellweg:
Only ask for permissions matching the object type.

fixes TBA

### DIFF
--- a/CHANGES/9589.bugfix
+++ b/CHANGES/9589.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug, where permissions were checked against the wrong object type.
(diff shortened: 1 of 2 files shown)


lubosmj:
Please, add the reference to the issue https://pulp.plan.io/issues/9589 in the commit message.

lubosmj (approved):

mdellweg:
The question now: Should we consider this a bug in pulp_container all along and fix and backport it? Or should we turn the corresponding check in pulpcore (I was so sure, guardian checked the same thing, but obviously no...) into a deprecation warning and only enforce with 3.20?

PR #444, opened 2021-12-03
Store content in the Redis cache

lubosmj:
ref #496
Required PR: https://github.com/pulp/pulpcore/pull/1751

### DIFF
--- a/pulp_container/app/cache.py
+++ b/pulp_container/app/cache.py
@@ -1,0 +1,74 @@
+from django.core.exceptions import ObjectDoesNotExist
+
+from pulpcore.plugin.cache import CacheKeys, AsyncContentCache, SyncContentCache
+
+from pulp_container.app.models import ContainerDistribution
+from pulp_container.app.exceptions import RepositoryNotFound
+
+ACCEPT_HEADER_KEY = "accept_header"
+
+
+class RegistryCache:
+    """A class that overrides the default key specs."""
+
+    def __init__(self, base_key=None, expires_ttl=None, auth=None):
+        """Initialize the parent class with the plugin's specific keys."""
+        updated_keys = (CacheKeys.path, CacheKeys.method, ACCEPT_HEADER_KEY)
+        super().__init__(base_key=base_key, expires_ttl=expires_ttl, keys=updated_keys, auth=auth)
+
+
+class RegistryContentCache(RegistryCache, AsyncContentCache):
+    """A wrapper around the Redis content cache handler tailored for the content application."""
+
+    def make_key(self, request):
+        """Make a key composed of the request's path, method, host, and accept header."""
+        accept_header = ",".join(sorted(request.headers.getall("accept", [])))
+        all_keys = {
+            CacheKeys.path: request.path,
+            CacheKeys.method: request.method,
+            CacheKeys.host: request.url.host,
+            ACCEPT_HEADER_KEY: accept_header,
+        }
+        key = ":".join(all_keys[k] for k in self.keys)
+        return key
+
+
+class RegistryApiCache(RegistryCache, SyncContentCache):
+    """A wrapper around the Redis content cache handler tailored for the registry API."""
+
+    def make_key(self, request):
+        """Make a key composed of the request's path, method, host, and accept header."""
+        all_keys = {
+            CacheKeys.path: request.path,
+            CacheKeys.method: request.method,
+            CacheKeys.host: request.get_host(),
+            ACCEPT_HEADER_KEY: request.headers.get("accept", ""),
+        }
+        key = ":".join(all_keys[k] for k in self.keys)
+        return key
+
+
+def find_base_path_cached(request, cached):
+    """
+    Returns the base-path to use for the base-key in the cache
+
+    Args:
+        request (:class:`aiohttp.web.request`): The request from the client.
+        cached (:class:`CacheAiohttp`): The Pulp cache
+
+    Returns:
+        str: The base-path associated with this request
+
+    """
+    path = request.resolver_match.kwargs["path"]
+    path_exists = cached.exists(base_key=path)
+    if path_exists:
+        return path
+    else:
+        try:
+            distro = ContainerDistribution.objects.select_related(
+                "repository", "repository_version"
+            ).get(base_path=path)
+        except ObjectDoesNotExist:
+            raise RepositoryNotFound(name=path)
+        return distro.base_path
--- a/pulp_container/app/registry.py
+++ b/pulp_container/app/registry.py
@@ -10,6 +10,8 @@ from multidict import MultiDict
 
 from pulpcore.plugin.content import Handler, PathNotResolved
 from pulpcore.plugin.models import ContentArtifact
+
+from pulp_container.app.cache import RegistryContentCache
 from pulp_container.app.models import ContainerDistribution, Tag, Blob
 from pulp_container.app.schema_convert import Schema2toSchema1ConverterWrapper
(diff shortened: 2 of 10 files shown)


lubosmj:
The PR is once again ready for review. HEAD requests are being now cached.

lubosmj:
@gerrod3, would you mind reviewing this PR from the caching perspective?

pulp_container/tests/functional/api/test_content_cache.py
```diff
@@ -0,0 +1,205 @@
...
+
+        response = requests.get(url, auth=auth, headers=headers)
+        response_metadata = self.fetch_response_metadata(response)
+        self.assertEqual(expected_metadata, response_metadata, url)
+
+    def fetch_response_metadata(self, response):
+        """Retrieve metadata from the passed response and normalize status code for redirects."""
+        print(response.request.url)
+        print(response.status_code)
+        print(response.headers.get("X-PULP-CACHE"))
```
gerrod3, line 183: Some leftover print statements

pulp_container/app/registry.py
```diff
@@ -76,6 +78,10 @@ async def _dispatch(file, headers):
         file_response = web.FileResponse(path, headers=full_headers)
         return file_response
 
+    @RegistryContentCache(
+        base_key=lambda req, cac: Registry.find_base_path_cached(req, cac),
+        auth=lambda req, cac, bk: Registry.auth_cached(req, cac, bk),
+    )
```
gerrod3, line 84: I think these should error since it should try to await the auth and base_key methods which are non-async.

pulp_container/app/cache.py
```diff
@@ -0,0 +1,132 @@
...
+        updated_keys = (CacheKeys.path, CacheKeys.method, ACCEPT_HEADER_KEY)
+        super().__init__(base_key=base_key, expires_ttl=expires_ttl, keys=updated_keys, auth=auth)
+
+
+class RegistryContentCache(RegistryCache, AsyncContentCache):
+    """A wrapper around the Redis content cache handler tailored for the content application."""
+
+    def make_key(self, request):
+        """Make a key composed of the request's path, method, host, and accept header."""
+        accept_header = ",".join(sorted(request.headers.getall("accept", [])))
```
ipanova, line 27: I understand that `getall` will get all of the occurrences of `Accept` header, which can be multiple,however, every Accept header can contain a comma separated list of headers.
see more for the reference https://github.com/pulp/pulp_container/blob/main/pulp_container/app/utils.py#L16

lubosmj: As far as I know, this had to be made because of how `MultiDict` stores its keys. I focused only on Accept headers that looked like this:
```
Accept: application/vnd.docker.distribution.manifest.list.v2+json,application/vnd.docker.distribution.manifest.v2+json
```

The call converts the header back to its original value from:
```
<CIMultiDictProxy( 'Accept': 'application/vnd.docker.distribution.manifest.list.v2+json',  'Accept': 'application/vnd.docker.distribution.manifest.v2+json')>
```

Are you sure a user can pass multiple Accept headers to a request?

pulp_container/app/cache.py
```diff
@@ -0,0 +1,132 @@
...
+class RegistryApiCache(RegistryCache, SyncContentCache):
+    """A wrapper around the Redis content cache handler tailored for the registry API."""
+
+    def make_key(self, request):
+        """Make a key composed of the request's path, method, host, and accept header."""
+        all_keys = {
+            CacheKeys.path: request.path,
+            CacheKeys.method: request.method,
+            CacheKeys.host: request.get_host(),
+            ACCEPT_HEADER_KEY: request.headers.get("accept", ""),
```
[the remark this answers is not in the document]
lubosmj: Could you please elaborate on the question? The value is accessed, as it should be with respect to the synchronous context.

pulp_container/app/cache.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Args:
+        request (:class:`aiohttp.web.request`): The request from the client.
+        cached (:class:`CacheAiohttp`): The Pulp cache
+
+    Returns:
+        str: The base-path associated with this request
+
+    """
+    path = request.resolver_match.kwargs["path"]
```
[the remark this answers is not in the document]
lubosmj: No, the whole for loop can be removed. Thank you for noticing that.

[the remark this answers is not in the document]
gerrod3: Yeah this code that was copied from the content app which was designed for the fact that we don't know where the base_path ends and the plugin's routing begins in the content app. However, pulp_container does know exactly what the base_path is so this loop isn't need and you can just check if the base_path exists in the cache.

pulp_container/app/cache.py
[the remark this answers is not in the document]
ipanova: i was asking why line 47 and 32 are different

pulp_container/app/cache.py
ipanova: based on the http rfc i read time ago - yes, one can pass multiple headers in a request

pulp_container/app/cache.py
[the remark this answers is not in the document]
lubosmj: Because we are fetching the headers from different contexts. One within `AsyncContentCache` and another one within `SyncContentCache` (we have different types of request objects).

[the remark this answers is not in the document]
lubosmj: These are two different types of request objects which depends on the context in which they are used. One is of type `aiohttp.request` and the second one is of type `django.http.request`. To access headers, it is necessary to use different attributes provided by that object type.

I am still required to verify whether these two types are correctly handling multiple accept headers that are provided within a single request. Thank you for the comment.

pulp_container/tests/functional/api/test_repositories_list.py
```diff
@@ -61,22 +59,8 @@ def get_listed_repositories(self, auth=None):
         """Fetch repositories from the catalog endpoint."""
         repositories_list_endpoint = urljoin(self.cfg.get_base_url(), "/v2/_catalog")
 
-        with self.assertRaises(HTTPError) as cm:
-            requests.get(repositories_list_endpoint).raise_for_status()
-        content_response = cm.exception.response
-        authenticate_header = content_response.headers["Www-Authenticate"]
-
-        queries = AuthenticationHeaderQueries(authenticate_header)
-        self.assertEqual(queries.scope, "registry:catalog:*")
```
ipanova, line 70: with your refactor we're losing this assertion

pulp_container/tests/functional/utils.py
```diff
@@ -264,3 +265,25 @@ def gen_artifact(url=CONTAINER_IMAGE_URL):
         temp_file.write(response.content)
         artifact = ArtifactsApi(core_client).create(file=temp_file.name)
         return artifact.to_dict()
+
+
+def get_auth_for_url(latest_image_url, auth=None):
```
ipanova, line 270: make it more generic s/latest_image_url/registry_endpoint_url/ or similar

ipanova (approved):
I don't want to hold back this PR. please address last 2 comments of mine and merge! thank you for working on this :seal:

lubosmj:
@ipanova, this PR can be merged after merging https://github.com/pulp/pulpcore/pull/1751. Do we need to bump the pulpcore requirement to publish this change? I am not sure if it will cause some problems in the future when we do not update the version (since we wanted to move the pulpcore requirement by every deprecation change only (3.20 now), is that right?).

PR #446, opened 2021-12-06
Enabled rate_limit option on the remote.

ipanova:
closes #9607

### DIFF
--- a/pulp_container/app/downloaders.py
+++ b/pulp_container/app/downloaders.py
@@ -57,6 +57,8 @@ class RegistryAuthHttpDownloader(HttpDownloader):
         headers.update(auth_headers)
         # aiohttps does not allow to send auth argument and auth header together
         self.session._default_auth = None
+        if self.download_throttler:
+            await self.download_throttler.acquire()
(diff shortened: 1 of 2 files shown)


bmbouter (approved):

pulp_container/app/downloaders.py
```diff
@@ -57,6 +57,8 @@ async def _run(self, handle_401=True, extra_data=None):
         headers.update(auth_headers)
         # aiohttps does not allow to send auth argument and auth header together
         self.session._default_auth = None
+        if self.download_throttler:
+            await self.download_throttler.acquire()
```
mdellweg, line 61: I'm not an expert in this area. The interface used here looks like download_throttler was a lock or semaphore, which kind of makes sense.

mdellweg (approved):

pulp_container/app/downloaders.py
ipanova: i took it from core https://github.com/pulp/pulpcore/blob/main/pulpcore/download/http.py#L255

PR #520, opened 2021-12-27
Add support for importing/exporting image signatures

### LINKED ISSUE #506: As a user I can export import/export image signarure altogether with the images

Redmine Issue: 9558, https://pulp.plan.io/issues/9558

---

This feature will be enabled only for sync repo type.

lubosmj:
closes #506


pulp_container/app/modelresource.py
```diff
@@ -76,6 +76,51 @@ class Meta:
...
+        Return signatures specific to a specified repo-version.
+        """
+        return ManifestSignature.objects.filter(pk__in=self.repo_version.content).order_by(
+            "content_ptr_id"
+        )
+
+    class Meta:
+        model = ManifestSignature
+        import_id_fields = model.natural_key_fields()
+        skip_diff = True
```
goosemania, line 121: What does it imply? I'm not sure I understand what diffs bring, why are they enabled by default.
This https://django-import-export.readthedocs.io/en/latest/api_resources.html?highlight=skip_diff#import_export.resources.ResourceOptions.skip_diff does not help me.

lubosmj: This implies that the content imported to Pulp will not be compared with the original content before the import. If I understand it correctly, the following [test](https://github.com/django-import-export/django-import-export/blob/0d4340f02582e86d4515159becb045cad1d53ee0/import_export/resources.py#L705) will not be executed:
```
parsed row before insert == parsed row after insert
```

Signatures will not be valid if the content will be corrupted after the import.

goosemania: @ipanova , @lubosmj , I realized that the signature is stored in the binary field because it was originally implemented to store the signature in a binary form but then we started storing it in the base64 encoded way.
Does it make sense to switch to the TextField then? It makes more sense now I think, and it should make import/export more robust it seems.
Thoughts?

lubosmj: Yes, if we are storing everything in base64, we can convert it to `TextField` and I can therefore get rid of `skip_diff`. Either way, I initially planned to convert `BinaryField` data to its hexadecimal representation before export and then convert it back to the binary format during the import.

goosemania: Cool, let's do it in this PR then, so the import/export works as normal with diffs. Thanks!

goosemania (approved):

PR #521, opened 2022-01-03
Only sync manifest list if all its images are signed.

goosemania:
A prep for the stages refactor where supporting
at-least-one-image-signed approach is not easily achievable.
No one asked for this functionality, so removing it for now.

[noissue]


mdellweg (approved):
Thank you!
Technically, i think, this is more than "just for the refactor", but also for a better definition of the immutable manifest list being ready.

ipanova (approved):
Thank you, that looks right to me. :seal:

PR #527, opened 2022-01-10
Add an extensions API endpoint for downloading signatures

### LINKED ISSUE #504: As a user I can view container image signatures

Redmine Issue: 9513, https://pulp.plan.io/issues/9513

---

Provide an extension of the docker/distribution API. 

Assuming a signed image has already been pushed into the Pulp Container Registry, one can read the signature(s)( any user with  access to the image) using the following calls:

* `GET /extensions/v2/<namespace>/<name>/signatures/`
* `GET /extensions/v2/<namespace>/<name>/signatures/<imagesha256digest>`

See https://docs.openshift.com/container-platform/3.10/admin_guide/image_signatures.html#reading-image-signatures-via-registry-api and https://github.com/openshift/origin/pull/12504/files

lubosmj:
closes #504

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -59,7 +59,7 @@ from pulp_container.app.token_verification import (
     RegistryPermission,
     TokenPermission,
 )
-from pulp_container.constants import EMPTY_BLOB
+from pulp_container.constants import EMPTY_BLOB, SIGNATURE_HEADER
 
 FakeView = namedtuple("FakeView", ["action", "get_object"])
 
@@ -292,7 +292,7 @@ class ContainerRegistryApiMixin:
         Provide common headers to all responses.
         """
         headers = super().default_response_headers
-        headers.update({"Docker-Distribution-Api-Version": "registry/2.0"})
+        headers.update({"Docker-Distribution-Api-Version": "registry/2.0", SIGNATURE_HEADER: "1"})
         return headers
 
     def get_exception_handler_context(self):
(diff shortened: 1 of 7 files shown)


pulp_container/app/registry_api.py
```diff
@@ -970,3 +975,42 @@ def receive_artifact(self, chunk):
                 artifact = Artifact.objects.get(sha256=artifact.sha256)
                 artifact.touch()
             return artifact
+
+
+class Signatures(ContainerRegistryApiMixin, ViewSet):
+    """A ViewSet for image signatures."""
+
+    lookup_value_regex = "sha256:[0-9a-f]+"
```
mdellweg, line 983: Let's be specific:
```suggestion
    lookup_value_regex = "sha256:[0-9a-f]{64}"
```

pulp_container/app/registry_api.py
```diff
@@ -292,7 +292,12 @@ def default_response_headers(self):
         Provide common headers to all responses.
         """
         headers = super().default_response_headers
-        headers.update({"Docker-Distribution-Api-Version": "registry/2.0"})
+        headers.update(
+            {
+                "Docker-Distribution-Api-Version": "registry/2.0",
+                "X-Registry-Supports-Signatures": "1",
```
goosemania, line 298: This header name is already defined in stages file https://github.com/pulp/pulp_container/blob/signing/pulp_container/app/tasks/sync_stages.py#L43, how about putting it into the constants and using it here and there?

pulp_container/app/registry_api.py
```diff
@@ -970,3 +970,42 @@ def receive_artifact(self, chunk):
...
+
+        return Response(self.get_response_data(signatures))
+
+    @staticmethod
+    def get_response_data(signatures):
+        """Extract version, type, name, and content from the passed signature data."""
+        data = []
+        for signature in signatures:
+            signature = {
+                "schemaVersion": 2,
```
ipanova, line 1005: that's not correct, version was correct https://github.com/containers/image/blob/b55fb86c28b7d743cf59701332cd78d4294c7c54/docker/docker_image_src.go#L559
https://github.com/containers/image/blob/f2c99a9b1a0cd6673aab0d538b48d68562702b04/docker/docker_client.go#L76

goosemania: Good find, "schemaVersion" worked much better!

@ipanova , https://github.com/containers/image/blob/f2c99a9b1a0cd6673aab0d538b48d68562702b04/docker/docker_client.go#L76 shows in the comment how json is being serialized, and that it's this sting and not the attribute name is used. But it is confusing, why not to make them the same :woman_shrugging:

ipanova: ok that's confusing as hell

lubosmj: `Version` stands for a parsed value. A raw JSON value is stored as "schemaVersion".

goosemania (approved):

PR #529, opened 2022-01-11
Remove signatures from a repo if their manifest is removed

### LINKED ISSUE #511: As a user when I remove manifest from a repo its signatures are removed as well

Redmine Issue: 9582, https://pulp.plan.io/issues/9582

---

None

goosemania:
closes #511

### DIFF
--- a/pulp_container/app/tasks/recursive_remove.py
+++ b/pulp_container/app/tasks/recursive_remove.py
@@ -4,6 +4,7 @@ from pulpcore.plugin.models import Content, Repository
 from pulp_container.app.models import (
     Blob,
     Manifest,
+    ManifestSignature,
     MEDIA_TYPE,
     Tag,
 )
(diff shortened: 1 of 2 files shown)


pulp_container/app/tasks/recursive_remove.py
```diff
@@ -111,8 +112,14 @@ def recursive_remove_content(repository_pk, content_units):
             .exclude(listed_blobs_must_remain)
         )
 
+        # signatures can't be shared, so no need to calculate which ones to remain
+        signatures_to_remove = ManifestSignature.objects.filter(
+            signed_manifest__in=manifests_to_remove
+        )
```
goosemania, line 118: `manifests_to_remove` which is defined on line 90 above, is used on line 105-106 and here.
Will it be executed once and then cached results are used, or is it executed every time, do you know? @mdellweg @dralley

ipanova (approved):

pulp_container/app/tasks/recursive_remove.py
mdellweg: I think, it is not even executed, but translated into SQL.
Something like:
```
select * from 'ManifestSignature' where 'signed_manifest_id' in select 'id' from 'Manifest' where ...
```

mdellweg (approved):

pulp_container/app/tasks/recursive_remove.py
goosemania: Right, but when the SQL is being executed?
It's a pain if it's done twice and there is no caching or other optimization in place. I hope the postgres and django are smart enough not to perform the same calculations N times.

mdellweg: To my understanding it's executed in Postgres. And Postgres is built for that kind of calls. Just a feeling, but i would not worry about it.

PR #535, opened 2022-01-13
Test

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -56,7 +56,6 @@ class ContainerFirstStage(Stage):
         """
         ContainerFirstStage.
         """
-        future_manifests = []
         tag_list = []
         tag_dcs = []
         to_download = []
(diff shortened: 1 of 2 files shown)


mdellweg:
Hmm, seems like this made things worse. Can it be that the same Blob is part of multiple Manifests? Still i don't know how this would result in that bad situation.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -131,6 +131,7 @@ async def run(self):
                     )
                     await self.put(man_dc)
                     tag_dc.extra_data["tagged_manifest_dc"] = man_dc
+                    await man_dc.resolution()
```
mdellweg, line 134:
```suggestion
                    await man_dc.resolution()
                    assert man_dc.content.pk is not Null
```
Can you add debug statements like this?

mdellweg:
> > Hmm, seems like this made things worse. Can it be that the same Blob is part of multiple Manifests? Still i don't know how this would result in that bad situation.
> 
> it can, i am also puzzled

Well this means that the deduplication logic for blobs in the pipeline will loose half of the association information (a bug on its own).
The fact that we have a missing manifest_id looks more like the deduplication killed a duplicated manifest_dc. Can Manifests be both part of a manifest list _and_ tagged directly?

mdellweg (approved):

PR #533, opened 2022-01-12
Add support to sync signatures using docker API extension

goosemania:
closes #528

PR #543, opened 2022-01-18
Handle PUT requests for pushing image signatures

### LINKED ISSUE #502: As a user I can publish container signature to the server

Redmine Issue: 9511, https://pulp.plan.io/issues/9511

---

To add a single signature, PUT a new json object with `version` set to 2( manifest schema version) , `type` set to atomic, and `content` set to the base64 encoded signature data (usually a GPG signed data).
Name should be set to an unique name with the form digest@per-image-name, where digest is an image manifest digest (also used in the URL), and per-image-name is any unique identifier.
```
$ curl -X PUT --data @signature.json http://<registry_endpoint>:24817/extensions/v2/<namespace>/<name>/signatures/<imagesha256digest>

cat signature.json

{
    "version": 2,
    "type":    "atomic",
    "name":    "sha256:4028782c08eae4a8c9a28bf661c0a8d1c2fc8e19dbaae2b018b21011197e1484@cddeb7006d914716e2728000746a0b23",
    "content": "<base64 encoded signature>",
  }
```

See https://github.com/openshift/openshift-docs/pull/3556/files and https://docs.openshift.com/container-platform/3.10/admin_guide/image_signatures.html#writing-image-signatures-using-registry-api

**Note** skopeo will upload signature given that regitry has present `X-Registry-Supports-Signatures ` header  https://github.com/mtrmac/image/commit/6c17ca34793b19accc3d278fc93ce68e9943fcb4

lubosmj:
closes #502

### DIFF
--- a/CHANGES/502.feature
+++ b/CHANGES/502.feature
@@ -1,0 +1,2 @@
+Added support for pushing image signatures to the Pulp Registry. The signatures can be pushed by
+utilizing the extensions API.
--- a/docs/workflows/sign-images.rst
+++ b/docs/workflows/sign-images.rst
@@ -298,3 +298,42 @@ It is possible to specify a single manifest identified by tag or a list of manif
 by proviging ``tags_list`` option to the call.
 Note that ``manifest lists`` are not signed, instead all the image manifests that manifest lists
 contain, are signed.
+
+Managing signatures via the Extensions API
+==========================================
+
+This API exposes an endpoint for reading and writing image signatures. Users should configure the
+sigstore section in the `registries.d file <https://github.com/containers/image/blob/main/docs/containers-registries.d.5.md>`_
+accordingly to benefit from the API.
+
+Reading image signatures
+------------------------
+
+To read existing signatures, issue the following GET request::
+
+    $ http GET http://localhost:24817/extensions/v2/<namespace>/<name>/signatures/sha256:<manifest-digest>
+
+Signatures are retrieved by container clients automatically if the policy requires so. The policy is
+defined in the file ``/etc/containers/policy.json``.
+
+Writing image signatures
+------------------------
+
+To add a new signature to an image, execute the following PUT request::
+
+    $ http PUT http://localhost:24817/extensions/v2/<namespace>/<name>/signatures/sha256:<manifest-digest> < signature.json
+
+The JSON payload has the same structure as described in the `container signature specs <https://github.com/containers/image/blob/main/docs/containers-signature.5.md>`_::
+
+    {
+      "schemaVersion": 2,
+      "type":    "atomic",
+      "name":    "sha256:4028782c08eae4a8c9a28bf661c0a8d1c2fc8e19dbaae2b018b21011197e1484@cddeb7006d914716e2728000746a0b23",
+      "content": "<cryptographic_signature>"
+    }
+
+This step can be also done via podman or skopeo. After configuring a GPG keyring, it is possible to
+issue the following command to push a tagged image altogether with its signature to the Pulp
+Registry::
+
+    $ podman push --tls-verify=false --sign-by username@email.com localhost:24817/<namespace>/<name>
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -4,6 +4,8 @@ Check `Plugin Writer's Guide`_ for more details.
 . _Plugin Writer's Guide:
     http://docs.pulpproject.org/plugins/plugin-writer/index.html
 """
+import base64
+import binascii
 import json
 import logging
 import hashlib
@@ -59,7 +61,13 @@ from pulp_container.app.token_verification import (
     RegistryPermission,
     TokenPermission,
 )
-from pulp_container.constants import EMPTY_BLOB, SIGNATURE_HEADER
+from pulp_container.app.utils import extract_data_from_signature
+from pulp_container.constants import (
+    EMPTY_BLOB,
+    SIGNATURE_HEADER,
+    SIGNATURE_PAYLOAD_MAX_SIZE,
+    SIGNATURE_TYPE,
+)
 
 FakeView = namedtuple("FakeView", ["action", "get_object"])
 
@@ -169,6 +177,24 @@ class ManifestInvalid(ParseError):
         )
 
 
+class ManifestSignatureInvalid(ParseError):
+    """An exception to render an HTTP 400 response with the code 'SIGNATURE_INVALID'."""
+
+    def __init__(self, digest):
+        """Initialize the exception with the digest of a signed manifest."""
+        super().__init__(
+            detail={
+                "errors": [
+                    {
+                        "code": "SIGNATURE_INVALID",
+                        "message": "signature invalid",
+                        "detail": {"manifest_digest": digest},
+                    }
+                ]
+            }
+        )
+
+
 class ContentRenderer(BaseRenderer):
     """
     Rendered class for rendering Manifest and Blob responses.
@@ -232,6 +258,22 @@ class ManifestResponse(Response):
             "Content-Length": 0,
         }
         super().__init__(headers=headers, status=status, content_type=manifest.media_type)
+
+
+class ManifestSignatureResponse(Response):
+    """
+    An HTTP response class after creating an image signature.
+    """
+
+    def __init__(self, signature, path, status=201):
+        """Initialize the headers with the path to the repository and corresponding digests."""
+        headers = {
+            "Location": "/extensions/v2/{path}/signatures/{digest}".format(
+                path=path, digest=signature.signed_manifest.digest
+            ),
+            "Content-Length": 0,
+        }
+        super().__init__(headers=headers, status=status)
 
 
 class BlobResponse(Response):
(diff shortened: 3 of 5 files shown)


CHANGES/502.feature
```diff
@@ -0,0 +1,2 @@
+Added support for pushing image signatures to the Pulp Registry. The signatures can be pushed by
```
ipanova, line 1: can you extend existing docs? for example add another section to the signing workflow and mention another way how get signature into pulp registry.

pulp_container/app/registry_api.py
```diff
@@ -234,6 +259,21 @@ def __init__(self, manifest, path, request, status=200):
...
+class ManifestSignatureResponse(Response):
+    """
+    An HTTP response class after creating an image signature.
+    """
+
+    def __init__(self, signature, path, status=201):
+        """Initialize the headers with the path to the repository and corresponding digests."""
+        headers = {
+            "Docker-Content-Digest": signature.digest,
+            "Location": f"/extensions/v2/{path}/signatures/{signature.signed_manifest.digest}",
```
ipanova, line 271: we have a mixture of string formatting, it'd be great to have one style :/

pulp_container/app/registry_api.py
```diff
@@ -234,6 +259,21 @@ def __init__(self, manifest, path, request, status=200):
...
 
+class ManifestSignatureResponse(Response):
+    """
+    An HTTP response class after creating an image signature.
+    """
+
+    def __init__(self, signature, path, status=201):
+        """Initialize the headers with the path to the repository and corresponding digests."""
+        headers = {
+            "Docker-Content-Digest": signature.digest,
```
ipanova, line 270: it cannot be signature.digest. PUT can contain a list of signature for 1 image, if anything then it should be digest of the image but in case of docker-content-digest header - usually this header contains digest of the object that has been just created/fetched.
tldr; remove this header

pulp_container/app/registry_api.py
```diff
@@ -1009,3 +1049,61 @@ def get_response_data(signatures):
...
+        """Create a new signature from the received data."""
+        _, repository = self.get_dr_push(request, path)
+
+        try:
+            manifest = models.Manifest.objects.get(digest=pk)
+        except models.Manifest.DoesNotExist:
+            raise ManifestNotFound(reference=pk)
+
+        signature_payload = request.META["wsgi.input"].read(SIGNATURE_PAYLOAD_MAX_SIZE)
+        signature_raw = base64.b64decode(json.loads(signature_payload)["content"])
```
ipanova, line 1063: what if the signature's payload will exceed the limit, will json be able to load and access the `content`?

pulp_container/app/registry_api.py
```diff
@@ -1009,3 +1049,61 @@ def get_response_data(signatures):
...
+        signature_json = extract_data_from_signature(signature_raw, manifest.digest)
```
ipanova, line 1064: this utility accounts only for one signature extraction, PUT can contain 1+ signatures.

pulp_container/app/registry_api.py
ipanova: we need to assert that body of PUT request is correct, namely that each signature instance contains schemaversion ==2 to and type == atomic. Name also should be required and i think it has some rules

pulp_container/app/registry_api.py
```diff
@@ -1009,3 +1049,61 @@ def get_response_data(signatures):
...
+            type=SIGNATURE_TYPE.ATOMIC_SHORT,
+            key_id=signature_json["signing_key_id"],
+            timestamp=signature_json["signature_timestamp"],
+            creator=signature_json["optional"].get("creator"),
+            data=base64.b64encode(signature_raw).decode(),
+            signed_manifest=manifest,
+        )
+        try:
+            signature.save()
+        except IntegrityError:
```
ipanova, line 1081: reading this code it feels like integrity error should not happen because existing signatures are not getting (re)uploaded, but let's play safe https://github.com/containers/image/blob/main/docker/docker_image_dest.go#L654

pulp_container/app/registry_api.py
```diff
@@ -1009,3 +1049,61 @@ def get_response_data(signatures):
...
+            signature = models.ManifestSignature.objects.get(name=signature.name)
```
ipanova, line 1082: i suggest get by digest and not name

pulp_container/app/registry_api.py
```diff
@@ -1009,3 +1049,61 @@ def get_response_data(signatures):
             }
             data.append(signature)
         return {"signatures": data}
+
+    def put(self, request, path, pk):
+        """Create a new signature from the received data."""
+        _, repository = self.get_dr_push(request, path)
+
+        try:
+            manifest = models.Manifest.objects.get(digest=pk)
```
ipanova, line 1058: please also check that the manifest is in the repo

pulp_container/app/registry_api.py
lubosmj: I will keep the try-except block.

pulp_container/app/registry_api.py
lubosmj: Could you please provide a reference for this statement? I was looking into https://github.com/openshift/origin/pull/12504/files#diff-a215f9ef5344e2e492dbb2e3b82acf1d51f330e76e2c1dcc51b10a896e2d22fcR76 and https://docs.openshift.com/container-platform/3.11/admin_guide/image_signatures.html#writing-image-signatures-using-registry-api.

ipanova: I've been looking here https://github.com/mtrmac/image/commit/6c17ca34793b19accc3d278fc93ce68e9943fcb4#diff-6c682511ebec73d3ee49b3ed3d33708b9c598f9dd59c4f0685e793cc16baed8fR376
I do see discrepancies in different sources, yes.  let's discuss more on monday

ipanova: apparently the client parses the list of signatures and then sends N number of PUT requests  with one signature https://github.com/mtrmac/image/commit/6c17ca34793b19accc3d278fc93ce68e9943fcb4#diff-6c682511ebec73d3ee49b3ed3d33708b9c598f9dd59c4f0685e793cc16baed8fR408

lubosmj: Exactly. So, I will not update the code to handle this situation.

docs/workflows/sign-images.rst
```diff
@@ -298,3 +298,27 @@ It is possible to specify a single manifest identified by tag or a list of manif
 by proviging ``tags_list`` option to the call.
 Note that ``manifest lists`` are not signed, instead all the image manifests that manifest lists
 contain, are signed.
+
+Push signed images to the Pulp Registry
```
ipanova, line 302: push image signature to the pulp registry

docs/workflows/sign-images.rst
```diff
@@ -298,3 +298,27 @@ It is possible to specify a single manifest identified by tag or a list of manif
...
+=======================================
+
+Signatures can be pushed to the Pulp Registry via the extensions API. This API exposes an endpoint
+for reading and writing image signatures.
+
+To add a new signature to an image, issue the following PUT request::
+
+    $ http PUT http://localhost:24817/extensions/v2/<namespace>/<name>/signatures/sha256:<manifest-digest> < signature.json
+
+The JSON payload should have the following structure::
```
ipanova, line 312: how about referencing  for more info/specs to this source https://github.com/containers/image/blob/main/docs/signature-protocols.md#openshift-dockerdistribution-api-extension

docs/workflows/sign-images.rst
```diff
@@ -298,3 +298,27 @@ It is possible to specify a single manifest identified by tag or a list of manif
...
+
+    {
+      "schemaVersion": 2,
+      "type":    "atomic",
+      "name":    "sha256:4028782c08eae4a8c9a28bf661c0a8d1c2fc8e19dbaae2b018b21011197e1484@cddeb7006d914716e2728000746a0b23",
+      "content": "<cryptographic_signature>"
+    }
+
+This step can be also done via podman. After configuring a GPG keyring, it is possible to issue
+the following command to push a tagged and signed manifest to the Pulp Registry::
```
ipanova, line 322: ... to push a tagged image altogether with its signature.

docs/workflows/sign-images.rst
```diff
@@ -298,3 +298,27 @@ It is possible to specify a single manifest identified by tag or a list of manif
...
+The JSON payload should have the following structure::
+
+    {
+      "schemaVersion": 2,
+      "type":    "atomic",
+      "name":    "sha256:4028782c08eae4a8c9a28bf661c0a8d1c2fc8e19dbaae2b018b21011197e1484@cddeb7006d914716e2728000746a0b23",
+      "content": "<cryptographic_signature>"
+    }
+
+This step can be also done via podman. After configuring a GPG keyring, it is possible to issue
```
ipanova, line 321: ... via podman or skopeo, or any another client that supports registry extensions API for signatures.

docs/workflows/sign-images.rst
```diff
@@ -298,3 +298,27 @@ It is possible to specify a single manifest identified by tag or a list of manif
 by proviging ``tags_list`` option to the call.
 Note that ``manifest lists`` are not signed, instead all the image manifests that manifest lists
 contain, are signed.
+
+Push signed images to the Pulp Registry
+=======================================
+
+Signatures can be pushed to the Pulp Registry via the extensions API. This API exposes an endpoint
```
ipanova, line 305: can you mention that containers registries.d file sigstore section needs to be configured accordingly? or point to the specs https://github.com/containers/image/blob/main/docs/containers-registries.d.5.md

ipanova: on the client side.

docs/workflows/sign-images.rst
ipanova: let me know what you think -not exactly sure this is the best place but it's convenient to add another section right below since it's about extensions api - 'serve signed images/image signature from the pulp registry.'
i'd cover in this section that  synced/pushed/or signed from within pulp signatures can be served by these extensions api and show an example of a GET call

ipanova: and also a note about containers registries.d file sigstore configuration on the client side

pulp_container/app/registry_api.py
```diff
@@ -169,6 +177,24 @@ def __init__(self, digest):
...
+
+    def __init__(self, digest):
+        """Initialize the exception with the digest of a signed manifest."""
+        super().__init__(
+            detail={
+                "errors": [
+                    {
+                        "code": "SIGNATURE_INVALID",
+                        "message": "signature invalid",
+                        "detail": {"digest": digest},
```
ipanova, line 191: how about manifest_digest, at first i thought this is signature digest until i looked further into the code.

pulp_container/app/registry_api.py
```diff
@@ -1009,3 +1051,75 @@ def get_response_data(signatures):
...
+        except binascii.Error:
+            raise ManifestSignatureInvalid(digest=pk)
+
+        signature_json = extract_data_from_signature(signature_raw, manifest.digest)
+        if signature_json is None:
+            raise ManifestSignatureInvalid(digest=pk)
+
+        sig_digest = hashlib.sha256(signature_raw).hexdigest()
+        signature = models.ManifestSignature(
+            name=f"{manifest.digest}@{sig_digest[:32]}",
```
ipanova, line 1086: please do like here https://github.com/pulp/pulp_container/pull/533/files#diff-b702cbd3ebdff19bf8e4dfc14688aa3fb36e1a47ef6cf84d24e0f75b61247d36R365

pulp_container/app/registry_api.py
```diff
@@ -1009,3 +1051,75 @@ def get_response_data(signatures):
...
+            digest=f"sha256:{sig_digest}",
+            type=SIGNATURE_TYPE.ATOMIC_SHORT,
+            key_id=signature_json["signing_key_id"],
+            timestamp=signature_json["signature_timestamp"],
+            creator=signature_json["optional"].get("creator"),
+            data=base64.b64encode(signature_raw).decode(),
```
ipanova, line 1092: you can probably use directly here `signature_dict["content"]`

pulp_container/app/serializers.py
```diff
@@ -160,6 +161,22 @@ class Meta:
...
 
 
+class ManifestSignaturePutSerializer(serializers.Serializer):
+    """
+    A serializer for image signatures provided in a PUT request.
+    """
+
+    name = serializers.RegexField(regex=VALID_SIGNATURE_NAME_REGEX)
+    schemaVersion = serializers.IntegerField(max_value=2, min_value=2)
+    type = serializers.CharField()
```
ipanova, line 171: you can make it a choice field with1 choice of atomic type

pulp_container/app/registry_api.py
ipanova: i have not seen the serializer for this yet, so feel free to ignore my previous comment

pulp_container/app/serializers.py
```diff
@@ -160,6 +161,17 @@ class Meta:
...
 
 
+class ManifestSignaturePutSerializer(serializers.Serializer):
+    """
+    A serializer for image signatures provided in a PUT request.
+    """
+
+    name = serializers.RegexField(regex=VALID_SIGNATURE_NAME_REGEX)
+    schemaVersion = serializers.IntegerField(max_value=2, min_value=2)
+    type = serializers.ChoiceField(["atomic"])
```
ipanova, line 171: we have a constant defined in constans.py you can use

ipanova (approved):
thank you!

PR #544, opened 2022-01-18
Update CI files

pulpbot:
[noissue]


mdellweg (approved):

.github/workflows/scripts/script.sh
```diff
@@ -162,7 +162,10 @@ fi
 if [ -f $FUNC_TEST_SCRIPT ]; then
   source $FUNC_TEST_SCRIPT
 else
-    pytest -v -r sx --color=yes --pyargs pulp_container.tests.functional
+
+    pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs pulp_container.tests.functional -m parallel -n 8
+    pytest -v -r sx --color=yes --pyargs pulp_container.tests.functional -m "not parallel"
```
lubosmj, line 167: What is the point of running the same test suite twice? In pulp_container, we have a bunch of tests that depend on each other within one class: https://github.com/pulp/pulp_container/blob/main/pulp_container/tests/functional/api/test_tagging_images.py#L65. Does this parallelization happens only on the file level? Similarly, could not our tests end up in an inconsistent state when we are depending on the number of added content and stuff like that?

lubosmj: Aha, now I see this: https://discourse.pulpproject.org/t/parallel-test-runs-added-to-ci/320. It is necessary to annotate tests to have them run in parallel. So, my points are not valid.

lubosmj (approved):

PR #556, opened 2022-01-27
Return a concise message exception on 500 in case file is missing on the FS.

### LINKED ISSUE #555: Fix 500 error when registry content is missing on the filesystem
```
Jan 26 16:39:55 rn000052275 gunicorn[238574]: pulp [-]:  - - [26/Jan/2022:22:39:55 +0000] "HEAD /v2/ee-supported-rhel8/blobs/sha256:1e6a9467750ba441584f9678e1537abd9e3b028f063da3994befd760452196f3 HTTP/1.0" 302 0 "-" "skopeo/1.4.2-dev"
Jan 26 16:39:55 rn000052275 gunicorn[238573]: [2022-01-26 22:39:55 +0000] [238656] [ERROR] Error handling request
Jan 26 16:39:55 rn000052275 gunicorn[238573]: Traceback (most recent call last):
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib64/python3.8/site-packages/aiohttp/web_protocol.py", line 422, in _handle_request
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    resp = await self._request_handler(request)
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib64/python3.8/site-packages/aiohttp/web_app.py", line 499, in _handle
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    resp = await handler(request)
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib64/python3.8/site-packages/aiohttp/web_middlewares.py", line 119, in impl
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    return await handler(request)
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib/python3.8/site-packages/pulpcore/content/authentication.py", line 35, in authenticate
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    return await handler(request)
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib/python3.8/site-packages/pulp_container/app/registry.py", line 269, in get_by_digest
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    return await Registry._dispatch(artifact.file, headers)
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib/python3.8/site-packages/pulp_container/app/registry.py", line 78, in _dispatch
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    full_headers["Content-Length"] = str(file.size)
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib/python3.8/site-packages/django/db/models/fields/files.py", line 71, in size
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    return self.storage.size(self.name)
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib/python3.8/site-packages/django/core/files/storage.py", line 334, in size
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    return os.path.getsize(self.path(name))
Jan 26 16:39:55 rn000052275 gunicorn[238573]:  File "/usr/lib64/python3.8/genericpath.py", line 50, in getsize
Jan 26 16:39:55 rn000052275 gunicorn[238573]:    return os.stat(filename).st_size
Jan 26 16:39:55 rn000052275 gunicorn[238573]: FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/pulp/media/artifact/1e/6a9467750ba441584f9678e1537abd9e3b028f063da3994befd760452196f3'
Jan 26 16:39:55 rn000052275 gunicorn[238573]: [26/Jan/2022:22:39:55 +0000] "HEAD /pulp/container/ee-supported-rhel8/blobs/sha256:1e6a9467750ba441584f9678e1537abd9e3b028f063da3994befd760452196f3?
```

This happened when a new pulp server was provisioned, pointed at an existing database. I was able to reproduce the error by just nuking things under `/var/lib/pulp/media/artifact/`.

ipanova:
closes #555

### DIFF
--- a/pulp_container/app/registry.py
+++ b/pulp_container/app/registry.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from gettext import gettext as _
 
 from asgiref.sync import sync_to_async
 
@@ -64,17 +65,15 @@ class Registry(Handler):
             The :class:`aiohttp.web.FileResponse` for the file.
 
         """
+        path = os.path.join(settings.MEDIA_ROOT, file.name)
+        if not os.path.exists(path):
+            raise Exception(_("Expected path '{}' is not found").format(path))
+
         full_headers = MultiDict()
 
         full_headers["Content-Type"] = headers["Content-Type"]
         full_headers["Docker-Content-Digest"] = headers["Docker-Content-Digest"]
         full_headers["Docker-Distribution-API-Version"] = "registry/2.0"
-        full_headers["Content-Length"] = str(file.size)
-        full_headers["Content-Disposition"] = "attachment; filename={n}".format(
-            n=os.path.basename(file.name)
-        )
-
-        path = os.path.join(settings.MEDIA_ROOT, file.name)
         file_response = web.FileResponse(path, headers=full_headers)
         return file_response
 
@@ -123,7 +122,7 @@ class Registry(Handler):
             and tag.tagged_manifest.media_type not in accepted_media_types
         ):
             log.warn(
-                "OCI format found, but the client only accepts {accepted_media_types}.".format(
+                _("OCI format found, but the client only accepts {accepted_media_types}.").format(
                     accepted_media_types=accepted_media_types
                 )
             )
--- a/pulp_container/app/schema_convert.py
+++ b/pulp_container/app/schema_convert.py
@@ -6,6 +6,9 @@ import hashlib
 import itertools
 import json
 import logging
+
+from gettext import gettext as _
+
 from collections import namedtuple
 from jwkest import jws, jwk, ecc
 
@@ -323,4 +326,11 @@ def _get_manifest_dict(manifest):
 
 
 def _get_dict(artifact):
-    return json.load(artifact.file)
+    try:
+        return json.load(artifact.file)
+    except FileNotFoundError:
+        raise Exception(
+            _(
+                "Expected manifest file 'sha256:{}' needed for schema conversion is not found"
+            ).format(artifact.sha256)
+        )
--- a/CHANGES/555.bugfix
+++ b/CHANGES/555.bugfix
@@ -1,0 +1,1 @@
+Return a more concise message exception on 500 during image pull when content is missing on the FS.


pulp_container/app/registry.py
```diff
@@ -62,6 +62,10 @@ async def _dispatch(file, headers):
             The :class:`aiohttp.web.FileResponse` for the file.
 
         """
+        path = os.path.join(settings.MEDIA_ROOT, file.name)
+        if not os.path.exists(path):
+            raise PathNotResolved(path)
```
mdellweg, line 67: Should this be guarded to the case where we use the default file storage?
os.path.exists will never succeed if the data is in s3, no?

mdellweg:
I think, having an artifact in db that is (no longer) backed up by a real file is to be considered broken in pulp (this is purely from memory and may be wrong.). Can we apply the same rules as for the reclaim feature here? I think this would involve deleting the artifact while keeping the remote artifact(s). And as long as there are remote artifacts, pulp would self-heal.

pulp_container/app/registry.py
ipanova: in case s3 is used than we will use s3redirects, this code path is not executed.

ipanova:
> I think, having an artifact in db that is (no longer) backed up by a real file is to be considered broken in pulp (this is purely from memory and may be wrong.). 
I agree, however we cannot issue an unhandled exception if form of 500 to the client.

Can we apply the same rules as for the reclaim feature here? I think this would involve deleting the artifact while keeping the remote artifact(s). And as long as there are remote artifacts, pulp would self-heal.

it does self heal when it has RA but in case image was pushed there are no RA created.

mdellweg:
> I agree, however we cannot issue an unhandled exception if form of 500 to the client.

I know, we don't want to. But if we consider this an error on the server side, 5xx errors are exactly what we should send. So to me it is mainly a question of how broken do we consider Pulp to be when it is somehow (for an unknown reason) unable to access its files in storage.
The same exception would happen if storage were on a network file system that is temporarily unavailable.

ipanova:
> > I agree, however we cannot issue an unhandled exception if form of 500 to the client.
> 
> I know, we don't want to. But if we consider this an error on the server side, 5xx errors are exactly what we should send. So to me it is mainly a question of how broken do we consider Pulp to be when it is somehow (for an unknown reason) unable to access its files in storage. The same exception would happen if storage were on a network file system that is temporarily unavailable.

temporarily unavailable is different from permanently.
going through 5xx codes list nothing would match this situation https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

mdellweg:
> going through 5xx codes list nothing would match this situation https://en.wikipedia.org/wiki/List_of_HTTP_status_codes thoughts?

If no other one matches better, a plain 500 is the way to go. Turning it into any 4xx is finger pointing at the client who did nothing wrong.

mdellweg (approved):
So we are going to throw 404, right. Maybe this is like a repo on a simple file server where the metadata points to missing files. I stated my case but am going to accept this solution.

pulp_container/app/schema_convert.py
```diff
@@ -323,4 +323,7 @@ def _get_manifest_dict(manifest):
 
 
 def _get_dict(artifact):
-    return json.load(artifact.file)
+    try:
+        return json.load(artifact.file)
+    except FileNotFoundError:
+        raise RuntimeError()
```
mdellweg, line 329: Is there a way to describe this error state in a concise way?

pulp_container/app/registry.py
```diff
@@ -64,17 +65,15 @@ async def _dispatch(file, headers):
             The :class:`aiohttp.web.FileResponse` for the file.
 
         """
+        path = os.path.join(settings.MEDIA_ROOT, file.name)
+        if not os.path.exists(path):
+            raise Exception(_("Expected path '{}' is not found").format(path))
```
lubosmj, line 70:
```suggestion
            raise Exception(_("Expected path '{path}' is not found").format(path=path))
```

pulp_container/app/schema_convert.py
```diff
@@ -323,4 +326,11 @@ def _get_manifest_dict(manifest):
...
 
 def _get_dict(artifact):
-    return json.load(artifact.file)
+    try:
+        return json.load(artifact.file)
+    except FileNotFoundError:
+        raise Exception(
+            _(
+                "Expected manifest file 'sha256:{}' needed for schema conversion is not found"
+            ).format(artifact.sha256)
```
lubosmj, line 335:
```suggestion
                "Expected manifest file 'sha256:{digest}' needed for schema conversion is not found"
            ).format(digest=artifact.sha256)
```

lubosmj (approved):
It might be convenient to have the names of formatting arguments specified inside the brackets to make the life of a translator easier. What do you think?

pulp_container/app/schema_convert.py
mdellweg: "... is not found."

pulp_container/app/registry.py
mdellweg: Is this the message that will be sent to the http client?
Should we say something like "The requested artifact '{path}' is missing in the backend storage."

Also, from a security viewpoint, it may be undesired to expose any file system information to the client.
@ggainey ?

lubosmj: Nginx does not forward detailed error messages when an internal server error occurs. So, the path should be visible only for administrators, I believe.

ipanova: that's correct @lubosmj

mdellweg: Nice.

mdellweg (approved):

PR #562, opened 2022-02-01
Update CI files

pulpbot:
[noissue]


fao89 (approved):

mdellweg:
Maybe it would be wise if we could switch off the need for the `Triage-Needed` label on a per plugin basis.

mdellweg (approved):

fao89:
> Maybe it would be wise if we could switch off the need for the `Triage-Needed` label on a per plugin basis.

Please comment on https://github.com/pulp/plugin_template/pull/565
After it is merged I'll trigger the update job and it will fix this PR

mdellweg (approved):

PR #563, opened 2022-02-02
Pull fixtures from GH Packages instead of DockerHub

lubosmj:
[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_pull_content.py
+++ b/pulp_container/tests/functional/api/test_pull_content.py
@@ -20,14 +20,14 @@ from pulp_container.tests.functional.utils import (
     core_client,
     gen_container_client,
     gen_container_remote,
-    get_docker_hub_remote_blobsums,
+    get_blobsums_from_remote_registry,
     BearerTokenAuth,
     AuthenticationHeaderQueries,
 )
 from pulp_container.tests.functional.constants import (
     CONTAINER_CONTENT_NAME,
-    REPO_UPSTREAM_NAME,
-    REPO_UPSTREAM_TAG,
+    REGISTRY_V2_REPO_HELLO_WORLD,
+    PULP_HELLO_WORLD_LINUX_TAG,
 )
 from pulp_container.constants import EMPTY_BLOB, MEDIA_TYPE
 
@@ -139,7 +139,7 @@ class PullContentTestCase(unittest.TestCase):
         # Assert that at least one layer is synced from remote:latest
         # and the checksum matched with remote
         self.assertTrue(
-            any([result["blobSum"] in checksums for result in get_docker_hub_remote_blobsums()]),
+            any([checksum in checksums for checksum in get_blobsums_from_remote_registry()]),
             "Cannot find a matching layer on remote registry.",
         )
 
@@ -236,11 +236,11 @@ class PullContentTestCase(unittest.TestCase):
         self.teardown_cleanups.append((registry.rmi, local_url))
         local_image = registry.inspect(local_url)
 
-        registry.pull(REPO_UPSTREAM_NAME)
-        remote_image = registry.inspect(REPO_UPSTREAM_NAME)
+        registry.pull(REGISTRY_V2_REPO_HELLO_WORLD)
+        remote_image = registry.inspect(REGISTRY_V2_REPO_HELLO_WORLD)
 
         self.assertEqual(local_image[0]["Id"], remote_image[0]["Id"])
-        registry.rmi(REPO_UPSTREAM_NAME)
+        registry.rmi(REGISTRY_V2_REPO_HELLO_WORLD)
 
     def test_pull_image_from_repository_version(self):
         """Verify that a client can pull the image from Pulp.
@@ -260,11 +260,11 @@ class PullContentTestCase(unittest.TestCase):
         self.teardown_cleanups.append((registry.rmi, local_url))
         local_image = registry.inspect(local_url)
 
-        registry.pull(REPO_UPSTREAM_NAME)
-        remote_image = registry.inspect(REPO_UPSTREAM_NAME)
+        registry.pull(REGISTRY_V2_REPO_HELLO_WORLD)
+        remote_image = registry.inspect(REGISTRY_V2_REPO_HELLO_WORLD)
 
         self.assertEqual(local_image[0]["Id"], remote_image[0]["Id"])
-        registry.rmi(REPO_UPSTREAM_NAME)
+        registry.rmi(REGISTRY_V2_REPO_HELLO_WORLD)
 
     def test_pull_image_with_tag(self):
         """Verify that a client can pull the image from Pulp with a tag.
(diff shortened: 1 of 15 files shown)


lubosmj:
The only problem with GH Packages is that it does not support schema conversion and two of our tests will need to be modified/removed.

fao89:
> The only problem with GH Packages is that it does not support schema conversion and two of our tests will need to be modified/removed.

you can make these tests only run nightly (when docker secret var is set):
we did something similar on pulp_ansible:
https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/api/collection/v3/test_sync.py#L255-L259

pulp_container/tests/functional/utils.py
```diff
@@ -157,18 +157,16 @@ def gen_container_remote(url=REGISTRY_V2_FEED_URL, **kwargs):
...
+    token_url = "https://ghcr.io/token?service=ghcr.io&scope=repository:{0}:pull".format(
+        upstream_name
+    )
     token_response = requests.get(token_url)
     token_response.raise_for_status()
     token = token_response.json()["token"]
 
-    blob_url = ("{0}/v2/library/{1}/manifests/latest").format(REGISTRY_V2_FEED_URL, upstream_name)
+    blob_url = ("{0}/v2/{1}/manifests/latest").format(REGISTRY_V2_FEED_URL, upstream_name)
     response = requests.get(blob_url, headers={"Authorization": "Bearer " + token})
```
ipanova, line 170: add aceept header for manifest.2 mediatype and the tests should pass

pulp_container/tests/functional/utils.py
```diff
@@ -157,18 +157,16 @@ def gen_container_remote(url=REGISTRY_V2_FEED_URL, **kwargs):
...
     response.raise_for_status()
     return response.json()["fsLayers"]
```
ipanova, line 172: this should be changed then respectively to 'layers'

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -139,7 +141,7 @@ def test_api_returns_same_checksum(self):
         # Assert that at least one layer is synced from remote:latest
         # and the checksum matched with remote
         self.assertTrue(
-            any([result["blobSum"] in checksums for result in get_docker_hub_remote_blobsums()]),
+            any([result["blobSum"] in checksums for result in get_blobsums_from_remote_registry()]),
```
ipanova, line 144: and here parsing would be changed as well a bit, it will be list of dictionaries
```
 [
        {
            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "size": 32654,
            "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"
        },
        {
            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "size": 16724,
            "digest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b"
        },
        {
            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
            "size": 73109,
            "digest": "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736"
        }
    ]
```

pulp_container/tests/functional/utils.py
lubosmj: The GH Packages Registry does not consider `Accept` headers and returns the content like it was uploaded.

pulp_container/tests/functional/utils.py
lubosmj: The GH Packages Registry does not support schema conversion. Thus, we cannot rely on `fsLayers` that was previously returned by DockerHub when we sent a request with no `accept` headers (which in result end up with converting manifest list from manifest.v2.list to manifest.v1).

ipanova: correct, that's why i am suggesting to change that to 'layers' because this is format for manifest.v2 version.
or you can leave fsLayers if you plan to push manifest.v1 schema.

ipanova: which when i look at your code you did exactly that :) https://github.com/pulp/pulp_container/pull/563/files#diff-092a0ab013f705709186cc205228a7603fe06485b818333a4225577bc843a0d6R181

pulp_container/tests/functional/constants.py
```diff
@@ -26,47 +26,18 @@
...
-# hello-world is the smalest container image available on docker hub 1.84kB
-REPO_UPSTREAM_NAME = "hello-world"
-"""The name of a Container repository.
+# GH Packages registry
+REGISTRY_V2 = "ghcr.io"
+REGISTRY_V2_FEED_URL = "https://ghcr.io"
 
-This repository has several desireable properties:
+# a repository having the size of 1.84kB
+REPO_UPSTREAM_NAME = "pulp/hello-world"
```
ipanova, line 34: can you rename this to a more explicit name? pulp_hello_world for example?

pulp_container/tests/functional/constants.py
```diff
@@ -26,47 +26,18 @@
...
-.. code-block:: sh
-
-    curl --location --silent \
-    https://registry.hub.docker.com/v2/repositories/$this_constant/tags \
-    | python -m json.tool
-"""
+# a repository containing 4 manifest lists and 5 manifests
+PULP_FIXTURE_1 = "pulp/test-fixture-1"
 
+# an alternative tag for the REPO_UPSTREAM_NAME image
```
ipanova, line 39: pulp_hello_world_tag, respectively

pulp_container/tests/functional/constants.py
lubosmj: Did you mean this:
```suggestion
PULP_HELLO_WORLD_REPO = "pulp/hello-world"
```
?

pulp_container/tests/functional/utils.py
```diff
@@ -157,21 +157,31 @@ def gen_container_remote(url=REGISTRY_V2_FEED_URL, **kwargs):
...
+    token_server_response = requests.get(
+        f"{REGISTRY_V2_FEED_URL}/token?service=ghcr.io&scope=repository:{upstream_name}:pull"
+    )
+    token_server_response.raise_for_status()
+    token = token_server_response.json()["token"]
+
+    s = requests.Session()
+    s.headers.update({"Authorization": "Bearer " + token})
+
+    manifest_url = f"{REGISTRY_V2_FEED_URL}/v2/{upstream_name}/manifests/latest"
```
ipanova, line 171: can you ad a comment that latest tag references a manifest list?

pulp_container/tests/functional/constants.py
ipanova, line 39: can you add a comment that linux tag references a manifest list and not an image?

ipanova:
Looks good, thank you for working on this. For the sake of better readability please update some variable names :medal_sports:

ipanova (approved):

PR #596, opened 2022-02-16
Successfully re-upload artifact in case it was previously removed.

### LINKED ISSUE #595: add ability to re-upload artifact to the content if it has been previously removed
**Version**
Please provide the versions of the pulpcore and pulp_container packages in use, and how they are installed.  If you are using Pulp via Katello, please provide the Katello version.

**Describe the bug**
1. sync a repo
2. run reclaim space on the synced repo. artifacts are removed
3. remote source stopped serving the content ( no longer available, broken link, etc)
4. try to re-upload artifact back via podman push
5. fail

**To Reproduce**

**Expected behavior**

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

ipanova:
closes #595

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -656,8 +656,10 @@ class BlobUploads(ContainerRegistryApiMixin, ViewSet):
                 )
                 blob_artifact.save()
             except IntegrityError:
-                pass
-
+                ca = ContentArtifact.objects.get(content=blob, relative_path=digest)
+                if not ca.artifact:
+                    ca.artifact = artifact
+                    ca.save(update_fields=["artifact"])
             upload.delete()
 
             dispatched_task = dispatch(
@@ -833,7 +835,10 @@ class Manifests(RedirectsMixin, ContainerRegistryApiMixin, ViewSet):
         try:
             ca.save()
         except IntegrityError:
-            pass
+            ca = ContentArtifact.objects.get(content=manifest, relative_path=manifest.digest)
+            if not ca.artifact:
+                ca.artifact = artifact
+                ca.save(update_fields=["artifact"])
         # both docker/oci format should contain layers, digest, media_type, size
         layers = content_data.get("layers")
         blobs = {}
--- a/CHANGES/595.bugfix
+++ b/CHANGES/595.bugfix
@@ -1,0 +1,1 @@
+Successfully re-upload artifact in case it was previously removed.


lubosmj (commented):
The proposed changes resolve the problem; however, when an artifact exists and was not reclaimed, Pulp will be still trying to update the artifact field (and the update procedure will be always called according to https://docs.djangoproject.com/en/4.0/ref/models/instances/#specifying-which-fields-to-save: `Specifying update_fields will force an update.`). Is this what we want?

pulp_container/app/registry_api.py
```diff
@@ -656,8 +656,9 @@ def put(self, request, path, pk=None):
                 )
                 blob_artifact.save()
             except IntegrityError:
-                pass
-
+                ca = ContentArtifact.objects.get(content=blob, relative_path=digest)
+                ca.artifact = artifact
+                ca.save(update_fields=["artifact"])
```
mdellweg, line 661: Trying to understand this...
The integrity error will be called on the condition that `(content, relative_path)` must be unique. So we trigger this exception handler when such a content artifact exists already. Now we look for it and it may (immediate) or may not (on demand, reclaimed) already point to that artifact. Does the call to `save` check whether a SQL call to the database is necessary?

What if after receiving the `IntegrityError`, the `ContentArtifact` is deleted? Can we do a `select_for_update` first? [maybe a separate refactoring effort]

ipanova:
> The proposed changes resolve the problem; however, when an artifact exists and was not reclaimed, Pulp will be still trying to update the artifact field (and the update procedure will be always called according to https://docs.djangoproject.com/en/4.0/ref/models/instances/#specifying-which-fields-to-save: `Specifying update_fields will force an update.`). Is this what we want?

it would not do any harm but this would be an unnecessary operation, i agree. i can add a check whether the artifact is associated to the CA and only in case it's missing to perform the update.

pulp_container/app/registry_api.py
ipanova: what is the real case scenario when content artifact is deleted?

ipanova: > Trying to understand this... The integrity error will be called on the condition that `(content, relative_path)` must be unique. So we trigger this exception handler when such a content artifact exists already. Now we look for it and it may (immediate) or may not (on demand, reclaimed) already point to that artifact. Does the call to `save` check whether a SQL call to the database is necessary?
i don't think there is any check, so i think we'd have one unnecessary update. see https://github.com/pulp/pulp_container/pull/596#issuecomment-1047762501
> 
> What if after receiving the `IntegrityError`, the `ContentArtifact` is deleted? Can we do a `select_for_update` first? [maybe a separate refactoring effort]

mdellweg: > what is the real case scenario when content artifact is deleted?

I was thinking about orphan cleanup. But that should not be a problem anymore once we arrive at this line.

ipanova: orphan cleanup would not be an issue, artifact and blob has been just touched.

mdellweg (approved):

mdellweg (approved):

PR #600, opened 2022-02-22
Don't store blob's media_type on the model.

### LINKED ISSUE #493: push operation does not save config blob with the correct media_type

Redmine Issue: 9229, https://pulp.plan.io/issues/9229

---

After push operation Config Blob is saved with media_type from a Regular Blob. This issue does not seem to affect pull operation hence priority is not 'everything is on fire ' :)

After podman push :
```

In [5]: Manifest.objects.first().config_blob.media_type
Out[5]: 'application/vnd.docker.image.rootfs.diff.tar.gzip'

In [6]: Blob.objects.distinct("media_type").values_list('media_type', flat=True)
Out[6]: <QuerySet ['application/vnd.docker.image.rootfs.diff.tar.gzip']>
```
https://github.com/pulp/pulp_container/blob/master/pulp_container/app/registry_api.py#L685

ipanova:
closes #493


ipanova:
@newswangerd FYI

mdellweg (approved):
Great change. Assuming all the tests still pass.

newswangerd:
@ipanova We use this data in the automation hub UI to determine if a manifest is a manifest list so we can show the manifest list UI. https://github.com/ansible/ansible-hub-ui/blob/7d7c0379a4b65c623287202d55e91052fc8e3325/src/containers/execution-environment-detail/execution_environment_detail_images.tsx#L562

Is there another way for us to make this determination?

ipanova:
> @ipanova We use this data in the automation hub UI to determine if a manifest is a manifest list so we can show the manifest list UI. https://github.com/ansible/ansible-hub-ui/blob/7d7c0379a4b65c623287202d55e91052fc8e3325/src/containers/execution-environment-detail/execution_environment_detail_images.tsx#L562
> 
> Is there another way for us to make this determination?

@newswangerd this is fine, since the media_type on the manifest model stays. it's removed only from the blob model https://github.com/pulp/pulp_container/issues/493#issuecomment-1047887882

PR #604, opened 2022-02-28
Consider remote webserver authentication in the Registry API

### LINKED ISSUE #558: Push api seems to not support PulpRemoteUserAuthentication
As part of katello we use the pulpcore.app.authentication.PulpRemoteUserAuthentication class for authentication.

But if we try to push blobs when using: `TOKEN_AUTH_DISABLED=True`, (and specifying a certificate with CN=admin), we get a  forbidden

If i edit this line: 

https://github.com/pulp/pulp_container/blob/508b693afdc0e0252f9350b0551dc18f7b83be5c/pulp_container/app/registry_api.py#L277

and change it to:
```
            return [RegistryAuthentication, PulpRemoteUserAuthentication]
```
(after properly importing PulpRemoteUserAuthentication) it works!  I have no idea if this is actually correct, and even if it is it likely needs to be done on a conditional basis (if PulpRemoteUserAuthentication is part of REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES in the settings)

lubosmj:
Required PR: https://github.com/pulp/pulpcore/pull/2263
closes #558

### DIFF
--- a/CHANGES/558.bugfix
+++ b/CHANGES/558.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug that disallowed users who were authenticated by a remote webserver to access the
+Registry API endpoints when token authentication was disabled.
--- a/pulp_container/app/token_verification.py
+++ b/pulp_container/app/token_verification.py
@@ -7,7 +7,11 @@ from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth import get_user_model
 
-from rest_framework.authentication import BaseAuthentication, BasicAuthentication
+from rest_framework.authentication import (
+    BaseAuthentication,
+    BasicAuthentication,
+    RemoteUserAuthentication,
+)
 from rest_framework.exceptions import AuthenticationFailed, NotAuthenticated
 from rest_framework.permissions import BasePermission, SAFE_METHODS
 
@@ -52,16 +56,36 @@ class RegistryAuthentication(BasicAuthentication):
     A basic authentication class that accepts empty username and password as anonymous.
     """
 
+    PULP_AUTHENTICATION_CLASS = "pulpcore.app.authentication.PulpRemoteUserAuthentication"
+    AUTH_CLASSES = settings.REST_FRAMEWORK["DEFAULT_AUTHENTICATION_CLASSES"]
+
     def authenticate(self, request):
         """
         Perform basic authentication with the exception to accept empty credentials.
+
         For anonymous user, Podman sends 'Authorization': 'Basic Og=='.
         This represents ":" in base64.
+
+        If basic authentication could not success, remote webserver authentication is considered.
         """
         if request.headers.get("Authorization") == "Basic Og==":
             return (AnonymousUser, None)
 
-        return super().authenticate(request)
+        try:
+            return super().authenticate(request)
+        except AuthenticationFailed:
+            if self.PULP_AUTHENTICATION_CLASS in self.AUTH_CLASSES:
+                return RemoteUserRegistryAuthentication().authenticate(request)
+            else:
+                raise
+
+
+class RemoteUserRegistryAuthentication(RemoteUserAuthentication):
+    """
+    A class that authenticates users who were authenticated by a remote web server.
+    """
+
+    header = settings.REMOTE_USER_ENVIRON_NAME
 
 
 class TokenAuthentication(BaseAuthentication):
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -1,16 +1,79 @@
 .. _authentication:
 
-Registry Token Authentication
-=============================
+Authentication
+==============
 
-Pulp registry supports the `token authentication <https://docs.docker.com/registry/spec/auth/token/>`_.
-This enables users to pull/push content with an authorized access. A token server grants access based on the
-user's privileges and current scope.
+The Pulp Registry supports `token authentication <https://docs.docker.com/registry/spec/auth/token/>`_.
+The token authentication is enabled by default and does not come pre-configured out of the box. See
+the section :ref:`token-authentication-label` for more details.
 
-The feature is enabled by default. However, it is possible to disable it from the settings by declaring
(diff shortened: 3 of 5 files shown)


ipanova:
This means we need to bump the pulpcore version requirement to 3.19.dev.
This also means that katello who's on older version won't be able to use this fix unless we make it  'backportable' together with the changes in core.

pulp_container/app/registry_api.py
```diff
@@ -204,7 +205,7 @@ def authentication_classes(self):
         List of authentication classes to check for this view.
         """
         if settings.get("TOKEN_AUTH_DISABLED", False):
-            return [RegistryAuthentication]
+            return [RemoteUserRegistryAuthentication, RegistryAuthentication]
```
ipanova, line 208: can you please remind me whether the order matters and if yes how this is evaluated?

lubosmj:
After some discussion, we decided to benefit from the PR that exposes the authentication class in pulpcore. When backporting, the code from pulpcore (`header = settings.REMOTE_USER_ENVIRON_NAME`) will be copied.

pulp_container/app/registry_api.py
lubosmj: The order matters. The authentication procedure ends when one of the authentication classes succeeds and returns an authenticated user from its method `authenticate`. If none of the authentication classes succeeds, a user shall not be authenticated.

There might be a problem with the Nginx configuration where data from both the basic authentication and remote authentication methods are passed to Pulp at the same time. Pulp will not be able to determine which method should be used and traverses the list of authentication classes one by one. Do you think we should check the Pulp settings and set the list of authentication classes accordingly to prevent unnecessary failures that could not be seen after misconfiguring Nginx?

```python
if  "pulpcore.app.authentication.PulpRemoteUserAuthentication" in settings.REST_FRAMEWORK["DEFAULT_AUTHENTICATION_CLASSES"]:
    return [RemoteUserRegistryAuthentication]  # remote auth
else:
    return [RegistryAuthentication]  # basic auth
```

ipanova: looking at our docs we mention how to configure remote user auth and that seems to remove basic auth but keep session auth. In that case the webserver would not have such dilemma. I am not sure whether a configuration where both remote user and basic auth  enabled is valid? @bmbouter can you comment?

CHANGES/558.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed a bug that disallowed users who were authenticated by a remote webserver to access the
```
ipanova, line 1: can you add a line about this in our docs? I think in the authentication.rst would be good?

pulp_container/app/registry_api.py
bmbouter: The thinking in those docs (I wrote years ago) is that when a user opts to use webserver auth they want everyone to auth with that and there shouldn't be another "side-door" of users so to speak. Removing the basic auth removes that "side-door" authentication mechanism basically.

Perhaps one use case (a guess really) would be if an admin setup Pulp and used the webserver auth to connect all of their enterprise users to it, but then they wanted a few built-in users as admins to debug it if the enterprise auth was misconfigured or unavailable.

Is this helpful at all?

CHANGES/558.bugfix
lubosmj: Sure, I extracted the whole section to the main index.

pulp_container/app/token_verification.py
```diff
@@ -61,7 +68,26 @@ def authenticate(self, request):
         if request.headers.get("Authorization") == "Basic Og==":
             return (AnonymousUser, None)
 
-        return super().authenticate(request)
+        try:
+            return super().authenticate(request)
+        except AuthenticationFailed:
+            if any(
+                self.PULP_AUTHENTICATION_CLASS in auth_class for auth_class in self.AUTH_CLASSES
```
ipanova, line 75: is not self.AUTH_CLASSES a tuple? you could just do if `self.PULP_AUTHENTICATION_CLASS n self.AUTH_CLASSES` or i am missing something?

pulp_container/app/token_verification.py
```diff
@@ -61,7 +68,26 @@ def authenticate(self, request):
...
+    """
+    A class that authenticates users who were authenticated by a remote web server.
+    """
+
+    header = settings.REMOTE_USER_ENVIRON_NAME
+
+    def authenticate(self, request):
+        """Initialize an authenticated user based on the REMOTE_USER header."""
+        user = super().authenticate(request)
+        return (AnonymousUser, None) if user is None else user
```
ipanova, line 90: Is this needed? it seems like AnonymousUser is handled earlier in the code on line 68

ipanova (changes requested):
please address those questions

pulp_container/app/token_verification.py
```diff
@@ -61,7 +68,26 @@ def authenticate(self, request):
         if request.headers.get("Authorization") == "Basic Og==":
             return (AnonymousUser, None)
 
-        return super().authenticate(request)
+        try:
+            return super().authenticate(request)
```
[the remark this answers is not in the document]
lubosmj: If the latter is not configured, the method returns `None` - Not authenticated.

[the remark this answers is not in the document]
lubosmj: Yes, I am missing `raise` in the except block to preserve the current behaviour.

pulp_container/app/token_verification.py
lubosmj: Yes, I should be able to do what you have sketched.

pulp_container/app/token_verification.py
lubosmj: `AnonymousUser` is handled only for Basic authentication scenarios. We need to handle it here as well.

The solution comes from the following assumption. When a user could not be authenticated by a remote webserver or is trying to perform actions as an anonymous user, no `REMOTE_USER` header will be set by nginx.

ipanova (approved):

PR #607, opened 2022-03-03
Rewrite an rbac test with gen_user fixture

mdellweg:
Demonstration for: https://github.com/pulp/pulp-smash/pull/1279

### DIFF
--- a/pulp_container/tests/functional/api/test_rbac_sync_repositories.py
+++ b/pulp_container/tests/functional/api/test_rbac_sync_repositories.py
@@ -1,98 +1,81 @@
 # coding=utf-8
 """Tests that container sync repositories have RBAC."""
-import unittest
+import pytest
 
 from pulp_smash import utils
 from pulp_smash.pulp3.bindings import monitor_task
 
-from pulp_container.tests.functional.utils import (
-    del_user,
-    gen_user,
-    skip_if,
-)
-
 from pulpcore.client.pulp_container.exceptions import ApiException
 
 
-class RBACSyncRepositoriesTestCase(unittest.TestCase):
+@pytest.mark.parallel
+def test_rbac_sync_repositories(gen_user, container_repository_api):
     """RBAC sync repositories."""
 
-    @classmethod
-    def setUpClass(cls):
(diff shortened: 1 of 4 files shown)


pulp_container/tests/functional/api/test_rbac_sync_repositories.py
```diff
@@ -1,98 +1,81 @@
...
+    with user2, pytest.raises(ApiException):
+        container_repository_api.delete(repository.pulp_href)
+    with user3, pytest.raises(ApiException):
+        container_repository_api.delete(repository.pulp_href)
+    with user1:
+        response = container_repository_api.delete(repository.pulp_href)
         monitor_task(response.task)
-        with self.assertRaises(ApiException):
-            self.user1["repository_api"].read(self.repository.pulp_href)
+        with pytest.raises(ApiException):
```
bmbouter, line 80: nice I didn't realize this was part of pytest.

mdellweg: Well they sure needed to provide something like this.

lubosmj:
I wonder if we could extract particular tests into separate functions. Right now, we have all scenarios included within a single function. We can create fixtures for users and pass them to those separate functions. Then, we would have something like this (`test_rbac_remotes.py`):
```
def test_create_remote(users, container_api):
    ...

def test_read_remote(users, container_api):
    ...

def test_update_remote(users, container_api):
    ...
```

Is it worth the time?

mdellweg:
> I wonder if we could extract particular tests into separate functions. Right now, we have all scenarios included within a single function. We can create fixtures for users and pass them to those separate functions. Then, we would have something like this (`test_rbac_remotes.py`):
> 
> ```
> def test_create_remote(users, container_api):
>     ...
> 
> def test_read_remote(users, container_api):
>     ...
> 
> def test_update_remote(users, container_api):
>     ...
> ```
> 
> Is it worth the time?

This is not even about the users. This is about the lifecycle of the remote under test. You need it created to update and later delete it. The fact that we don't need to recreate the users for every single test is just a nice plus.

PR #620, opened 2022-03-08
Fixed check for the signature source location.

### LINKED ISSUE #617: Downloader exception is checked for http code 200
**Version**
Current main branch.

**Describe the bug**
A suspicious line of code is found:
https://github.com/pulp/pulp_container/pull/436/files#diff-b702cbd3ebdff19bf8e4dfc14688aa3fb36e1a47ef6cf84d24e0f75b61247d36R89

The downloader will probably not trigger the exception on a 200 response, and therefore the api extensions will never be checked if the registry goes without authentication.

ipanova:
closes #617

### DIFF
--- a/CHANGES/617.bugfix
+++ b/CHANGES/617.bugfix
@@ -1,0 +1,1 @@
+Fixed check for the signature source location.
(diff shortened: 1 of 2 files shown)


ipanova:
> Thanks for fixing my silly bug (:

technically this has not been release yet so there is no official bug :lollipop: :eyes:

mdellweg (approved):

PR #621, opened 2022-03-09
Consider roles instead of raw permission classes

lubosmj:
[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_rbac_repo_versions.py
+++ b/pulp_container/tests/functional/api/test_rbac_repo_versions.py
@@ -184,7 +184,7 @@ class PushRepoVersionTestCase(unittest.TestCase, rbac_base.BaseRegistryTest):
             ]
         )
         cls.user_reader = gen_user(model_roles=["container.containerdistribution_consumer"])
-        cls.user_helpless = gen_user([])
+        cls.user_helpless = gen_user()
 
         # create a push repo
         image_path = f"{REGISTRY_V2_REPO_PULP}:manifest_d"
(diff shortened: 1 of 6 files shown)


lubosmj:
The changes address a bug in `PushCrossRepoBlobMountTestCase` where model and object permissions are no longer used after merging the roles PR.

pulp_container/tests/functional/api/test_rbac_repo_versions.py
```diff
@@ -271,17 +271,18 @@ def setUpClass(cls):
         cls.distribution = cls.distributions_api.list(name="test-1").results[0]
 
         cls.user_pull = gen_user(
-            object_permissions=[
-                ("container.pull_containerdistribution", cls.distribution.pulp_href)
+            object_roles=[
+                ("container.containerdistribution_consumer", cls.distribution.pulp_href),
+                ("container.containernamespace_consumer", cls.distribution.namespace),
```
mdellweg, line 276: Would not giving them the namespace role be sufficient? (I think it i should be.)

pulp_container/tests/functional/api/test_rbac_repo_versions.py
```diff
@@ -271,17 +271,18 @@ def setUpClass(cls):
...
             ]
         )
         cls.user_push = gen_user(
-            [
-                "container.push_containerdistribution",
-                "container.namespace_push_containerdistribution",
+            object_roles=[
+                ("container.containerdistribution_collaborator", cls.distribution.pulp_href),
+                ("container.containernamespace_collaborator", cls.distribution.namespace),
```
mdellweg, line 282: Again, i think one of them should be sufficient.

pulp_container/tests/functional/utils.py
```diff
@@ -57,14 +57,8 @@
 TOKEN_AUTH_DISABLED = utils.get_pulp_setting(cli_client, "TOKEN_AUTH_DISABLED")
 
 
-def gen_user(model_permissions=None, object_permissions=None, model_roles=None, object_roles=None):
```
mdellweg, line 60: :+1:

pulp_container/tests/functional/api/test_rbac_repo_versions.py
lubosmj: Yes, you are right. I tested the push permission with the namespace permission and it was working.

mdellweg: Now i'm asking myself whether we should have a test for both individually.

lubosmj: The test case tries to verify whether a user has all the required permissions to perform blob mounting (`pull` for one repository & `push` for another repository). We are not trying to test the whole access_policy workflow for pushing content. I am fine with namespace permissions only.

lubosmj: Maybe, I should move the test case to another file. The name evokes that we are trying to test the whole RBAC suite, which is wrong.

mdellweg: Well, we are creating users and so we testing for rbac features. I think, the name fits.

mdellweg (approved):

PR #633, opened 2022-03-14
Add a test case for pushing image signatures

### LINKED ISSUE #606: Write tests for pushing image signatures
The test case should verify the following workflow:
1. A user signs an image and pushes it to Pulp via registry extensions API
2. Pulp stores the image with the corresponding signatures.
3. The signatures can be retrieved from the Registry extensions API.
4. The signatures can be browsed in the Pulp API.

lubosmj:
closes #606

### DIFF
--- a/pulp_container/tests/functional/api/test_push_signatures.py
+++ b/pulp_container/tests/functional/api/test_push_signatures.py
@@ -45,6 +45,9 @@ def distribution(
     add_to_cleanup,
 ):
     """Return a distribution created after pushing a signed content to the Pulp Registry."""
+    if registry_client.name != "podman":
+        pytest.skip("This test requires podman to sign pulled content", allow_module_level=True)
+
     registry_client.pull(f"{REGISTRY_V2_REPO_PULP}:manifest_a")
 
     image_path = f"{REGISTRY_V2_REPO_PULP}:manifest_a"
--- a/pulp_container/tests/functional/api/test_push_signatures.py
+++ b/pulp_container/tests/functional/api/test_push_signatures.py
@@ -1,0 +1,119 @@
+"""Tests that verify that an image signature can be pushed to Pulp."""
+import gnupg
+import base64
+import json
+import requests
+import pytest
+import tempfile
+
+from pulp_container.tests.functional.constants import REGISTRY_V2_REPO_PULP
+from pulp_container.constants import SIGNATURE_TYPE
+
+KEY = "https://raw.githubusercontent.com/pulp/pulp-fixtures/master/common/GPG-PRIVATE-KEY-pulp-qe"
+
+
+@pytest.fixture(scope="session")
+def tempdir():
+    """A temporary directory."""
+    with tempfile.TemporaryDirectory() as directory:
+        yield directory
+
+
+@pytest.fixture(scope="session")
+def gpg_metadata(tempdir):
+    """A fixture for a GPG instance and related metadata."""
+    private_key = requests.get(KEY).text
+
+    gpg = gnupg.GPG(gnupghome=tempdir)
+
+    gpg.import_keys(private_key)
+
+    fingerprint = gpg.list_keys()[0]["fingerprint"]
+    keyid = gpg.list_keys()[0]["keyid"]
+
+    gpg.trust_keys(fingerprint, "TRUST_ULTIMATE")
+
+    return gpg, fingerprint, keyid
+
+
+@pytest.fixture
+def distribution(
+    registry_client,
+    local_registry,
+    container_distribution_api,
+    gpg_metadata,
+    add_to_cleanup,
+):
+    """Return a distribution created after pushing a signed content to the Pulp Registry."""
+    registry_client.pull(f"{REGISTRY_V2_REPO_PULP}:manifest_a")
+
+    image_path = f"{REGISTRY_V2_REPO_PULP}:manifest_a"
+    gpg, fingerprint, keyid = gpg_metadata
+
+    with registry_client._client.machine.env(GNUPGHOME=gpg.gnupghome):
+        local_registry.tag_and_push(image_path, "test-1:manifest_a", "--sign-by", keyid)
+
+        # push the same image for the second time with a different signature (timestamp)
+        local_registry.tag_and_push(image_path, "test-1:manifest_a", "--sign-by", keyid)
+
+    distribution = container_distribution_api.list(name="test-1").results[0]
+    add_to_cleanup(container_distribution_api, distribution.pulp_href)
+
+    return distribution
+
+
+def test_assert_signed_image(
+    local_registry,
+    container_push_repository_api,
+    container_manifest_api,
+    container_signature_api,
+    gpg_metadata,
+    distribution,
+):
+    """Test whether an admin user can fetch a signature from the Pulp Registry."""
+    gpg, fingerprint, keyid = gpg_metadata
+
+    repository = container_push_repository_api.read(distribution.repository)
+    manifest = container_manifest_api.list(
+        repository_version=repository.latest_version_href
+    ).results[0]
+
+    signature = container_signature_api.list(
+        repository_version=repository.latest_version_href
+    ).results[0]
+
+    assert manifest.digest in signature.name
+    assert signature.signed_manifest == manifest.pulp_href
+    assert signature.key_id == keyid
+
+    path = f"/extensions/v2/test-1/signatures/{manifest.digest}"
+    response, _ = local_registry.get_response("GET", path)
+
+    signatures = response.json()["signatures"]
+
+    assert len(signatures) == 2
+
+    timestamps = []
+    for s in signatures:
+        raw_s = base64.b64decode(s["content"])
+        decrypted = gpg.decrypt(raw_s)
+
+        assert decrypted.key_id == keyid
+        assert decrypted.fingerprint == fingerprint
+        assert decrypted.status == "signature valid"
+        assert decrypted.username == "Pulp QE"
+
+        json_s = json.loads(decrypted.data)
+
+        image_path = json_s["critical"]["identity"]["docker-reference"]
+        assert image_path == f"{local_registry.name}/test-1:manifest_a"
+
+        s_type = json_s["critical"]["type"]
+        assert s_type == SIGNATURE_TYPE.ATOMIC_FULL
+
+        referenced_manifest = json_s["critical"]["image"]["docker-manifest-digest"]
+        assert referenced_manifest == manifest.digest
+
+        timestamps.append(json_s["optional"]["timestamp"])
+
+    assert timestamps[0] != timestamps[1]
--- a/pulp_container/tests/functional/api/test_rbac_repo_versions.py
+++ b/pulp_container/tests/functional/api/test_rbac_repo_versions.py
@@ -12,12 +12,7 @@ from pulp_smash.pulp3.utils import gen_repo
 from pulpcore.client.pulp_container.exceptions import ApiException
 
 from pulp_container.tests.functional.constants import PULP_FIXTURE_1, REGISTRY_V2_REPO_PULP
-from pulp_container.tests.functional.utils import (
-    TOKEN_AUTH_DISABLED,
-    gen_container_remote,
-    BearerTokenAuth,
-    AuthenticationHeaderQueries,
-)
+from pulp_container.tests.functional.utils import TOKEN_AUTH_DISABLED, gen_container_remote
 
 from pulpcore.client.pulp_container import (
     ContainerContainerRepository,
@@ -274,41 +269,19 @@ def test_cross_repository_blob_mount(
 
 
 @pytest.fixture
-def mount_blob(
-    pulp_cfg, bindings_cfg, add_to_cleanup, container_namespace_api, container_distribution_api
-):
-    """Fixture to mount blobs to new repositories, that are added to autocleanup."""
+def mount_blob(local_registry, container_distribution_api, container_namespace_api, add_to_cleanup):
+    """A fixture function to mount blobs to new repositories that will be cleaned up."""
 
     def _mount_blob(blob, source, dest):
-        """Try to mount the blob with the provided credentials."""
-        mount_url = f"/v2/{dest}/blobs/uploads/?from={source}&mount={blob.digest}"
-        url = urljoin(pulp_cfg.get_base_url(), mount_url)
+        mount_path = f"/v2/{dest}/blobs/uploads/?from={source}&mount={blob.digest}"
+        response, auth = local_registry.get_response(
+            "POST", mount_path, scope=f"repository:{source}:pull"
+        )
 
-        basic_auth = (bindings_cfg.username, bindings_cfg.password)
-        if TOKEN_AUTH_DISABLED:
-            auth = basic_auth
-        else:
-            response = requests.post(url, auth=basic_auth)
-            assert response.status_code == 401
-
-            authenticate_header = response.headers["Www-Authenticate"]
-            queries = AuthenticationHeaderQueries(authenticate_header)
-            response = requests.get(
-                queries.realm,
-                params={
-                    "service": queries.service,
-                    "scope": [queries.scope, "repository:{source}:pull"],
-                },
-                auth=basic_auth,
-            )
-            response.raise_for_status()
-            token = response.json()["token"]
-            auth = BearerTokenAuth(token)
-
-        response = requests.post(url, auth=auth)
         if response.status_code == 201:
             distribution = container_distribution_api.list(name=dest).results[0]
             add_to_cleanup(container_namespace_api, distribution.namespace)
+
         return response, auth
 
     return _mount_blob
--- a/CHANGES/606.misc
+++ b/CHANGES/606.misc
@@ -1,0 +1,1 @@
+Added a test case for pushing image signatures.
--- a/functest_requirements.txt
+++ b/functest_requirements.txt
@@ -1,2 +1,3 @@
 pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pytest
+python-gnupg
--- a/pulp_container/tests/conftest.py
+++ b/pulp_container/tests/conftest.py
@@ -1,6 +1,8 @@
 import pytest
 
-from urllib.parse import urlparse
+from urllib.parse import urljoin, urlparse
+
+import requests
 from pulp_smash.cli import RegistryClient
 from pulpcore.client.pulp_container import (
     ApiClient,
@@ -14,6 +16,13 @@ from pulpcore.client.pulp_container import (
     ContentTagsApi,
     ContentManifestsApi,
     ContentBlobsApi,
+    ContentSignaturesApi,
+)
+
+from pulp_container.tests.functional.utils import (
+    TOKEN_AUTH_DISABLED,
+    AuthenticationHeaderQueries,
+    BearerTokenAuth,
 )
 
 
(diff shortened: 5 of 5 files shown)


pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,172 @@
...
+
+from pulpcore.client.pulp_container import (
+    DistributionsContainerApi,
+    RepositoriesContainerPushApi,
+    ContentSignaturesApi,
+    ContentManifestsApi,
+)
+
+
+class PushSignatureRepoTestCase(PulpTestCase, rbac_base.BaseRegistryTest):
```
mdellweg, line 34: I would rather prefer new tests to be written with the new pytest style.

lubosmj: Okay, I am waiting for your PR (https://github.com/pulp/pulp_container/pull/632) to be merged. I see that you added fixture `local_storage` which I would like to use.

mdellweg: I undrafted it so we can get that work unblocked.

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,125 @@
...
+    delete_orphans,
+    monitor_task,
+)
+from pulp_container.tests.functional.constants import REGISTRY_V2_REPO_PULP
+from pulp_container.tests.functional.utils import get_auth_for_url
+
+gpg = gnupg.GPG()
+
+@pytest.fixture
+def key_data():
```
mdellweg, line 21: I've seen us using this key for tests: https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-PRIVATE-KEY-pulp-qe

It's probably better than regenerating a new one all the time.

mdellweg: Also this fixture is probably worth to be session scoped.

lubosmj: Where can I find the corresponding public key?

lubosmj: Okay, the private key can be imported and then I can export the public key...

lubosmj: > Also this fixture is probably worth to be session scoped.

Is there any reason for this?

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,120 @@
...
+
+KEY = "https://raw.githubusercontent.com/pulp/pulp-fixtures/master/common/GPG-PRIVATE-KEY-pulp-qe"
+
+gpg = gnupg.GPG()
+
+
+@pytest.fixture
+def key_data():
+    """Return the ID and fingerprint of the Pulp QE GPG key."""
+    private_key = requests.get(KEY).text
```
mdellweg, line 26: I think we should use the tmpdir fixture here. (Not sure if it allows session scope.)
https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,120 @@
...
+        repository_version=repository.latest_version_href
+    ).results[0]
+
+    assert manifest.digest in signature.name
+    assert manifest.pulp_href == signature.signed_manifest
+    assert keyid == signature.key_id
+
+    url = f"{pulp_cfg.get_base_url()}/extensions/v2/test-1/signatures/{manifest.digest}"
+    auth = get_auth_for_url(url)
+    response = requests.get(url, auth=auth)
```
mdellweg, line 102: Wondering if that should be a "feature" of the local_registry.

lubosmj: If you mean making `get_auth_for_url` a fixture, I am against it. The least I can do is to move `response = requests.get()` to `get_auth_for_url` and rename it to `get_response_with_auth`.

pulp_container/tests/functional/api/test_push_signatures.py
lubosmj: I am receiving: `ScopeMismatch: You tried to access the function scoped fixture tmpdir with a module scoped request object`. So, I will not use `tmpdir` and leave it as it is.

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,119 @@
...
+        repository_version=repository.latest_version_href
+    ).results[0]
+
+    signature = container_signature_api.list(
+        repository_version=repository.latest_version_href
+    ).results[0]
+
+    assert manifest.digest in signature.name
+    assert manifest.pulp_href == signature.signed_manifest
+    assert keyid == signature.key_id
```
mdellweg, line 98: :+1:
Usually i write the "==" tests the other way around. On the left the thing that should be and on the right what it should be. Read like "assert that <something> is <value>".

pulp_container/tests/functional/conftest.py
```diff
@@ -0,0 +1,79 @@
...
+from pulp_container.tests.functional.utils import (
+    TOKEN_AUTH_DISABLED,
+    AuthenticationHeaderQueries,
+    BearerTokenAuth,
+)
+
+
+@pytest.fixture
+def get_response(bindings_cfg):
+    """Return a response considering token authentication."""
```
mdellweg, line 15: I kind of expect mount blob to use this fixture. But it's using requests.post instead of get. Can we implement these in a way that we can call `local_registry.get(relative_url)` and `local_registry.post(relative_path, payload)`?

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,119 @@
...
+
+from pulp_smash.pulp3.bindings import (
+    delete_orphans,
+    monitor_task,
+)
+from pulp_container.tests.functional.constants import REGISTRY_V2_REPO_PULP
+
+KEY = "https://raw.githubusercontent.com/pulp/pulp-fixtures/master/common/GPG-PRIVATE-KEY-pulp-qe"
+
+gpg = gnupg.GPG()
```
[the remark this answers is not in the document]
lubosmj: No. The `gpg` utility is only used to verify if a pushed signature is correctly parsed in Pulp and is not "tampered". In one of the fixtures, I am importing the keyring locally to sign a manifest with it and then I test whether the signed manifest returned from the extensions API is valid.

pulp_container/tests/functional/conftest.py
lubosmj: Yes, I can work on that.

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,119 @@
...
+        decrypted = gpg.decrypt(raw_s)
+
+        assert decrypted.key_id == keyid
+        assert decrypted.fingerprint == fingerprint
+        assert decrypted.status == "signature valid"
+        assert decrypted.username == "Pulp QE"
+
+        json_s = json.loads(decrypted.data)
+
+        image_path = json_s["critical"]["identity"]["docker-reference"]
```
ipanova, line 115: since we're here maybe worth asserting on signature type

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,119 @@
...
+
+    assert manifest.digest in signature.name
+    assert manifest.pulp_href == signature.signed_manifest
+    assert keyid == signature.key_id
+
+    url = f"{pulp_cfg.get_base_url()}/extensions/v2/test-1/signatures/{manifest.digest}"
+    response = get_response(url)
+
+    signatures = response.json()["signatures"]
+    for s in signatures:
```
ipanova, line 104: can you assert that this list is one element long?

ipanova: this is not really mentioned in the test plan but if you want to extend it, you can execute this command once again `local_registry.push(image_path, local_image_path, "--sign-by", identity)` and assert that pulp stores 2 signatures ( diff is in timestamp)

pulp_container/tests/functional/api/test_push_signatures.py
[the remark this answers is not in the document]
lubosmj: I decided to not make any changes with regards to the initialization of `gpg`. The problem is that we are dealing with two versions of `gpg` -- for a user and an administrator (root).

The root's `gpg` is used for signing images when this is called: `local_registry.push(image_path, local_image_path, "--sign-by", identity)` -- representing a scenario where administrator signs content on his machine. The user's `gpg` is used for verifying signatures served by the extensions API -- representing a scenario where the user fetches the signatures from the extensions API and validates them.

As a result, we have two separate `gpg` utilities residing at different scopes of an OS. I cannot influence the way how the root's `gpg` is used when we are doing `sudo podman` by default. Apart from that, I am cleaning up everything that was created in the setup.

pulp_container/tests/functional/api/test_push_signatures.py
lubosmj: In the latest changes, I am pushing the same image two times to verify the aforementioned scenario.

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,129 @@
...
+        assert decrypted.status == "signature valid"
+        assert decrypted.username == "Pulp QE"
+
+        json_s = json.loads(decrypted.data)
+
+        image_path = json_s["critical"]["identity"]["docker-reference"]
+        assert image_path == f"{local_registry.name}/test-1:manifest_a"
+
+        s_type = json_s["critical"]["type"]
+        assert s_type == "atomic container signature"
```
ipanova, line 122: we have this in constants

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,129 @@
...
+
+        referenced_manifest = json_s["critical"]["image"]["docker-manifest-digest"]
+        assert referenced_manifest == manifest.digest
+
+        timestamps.append(json_s["optional"]["timestamp"])
+
+    assert timestamps[0] != timestamps[1]
```
ipanova, line 129: nice

ipanova (approved):

pulp_container/tests/functional/api/test_push_signatures.py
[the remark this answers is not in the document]
mdellweg: That's not entirely true. We are manipulating the users keyring. And if the user decides to have that key, we are deleting it anyway. Also while the test is running the user will have that key in the keyring even if undesired. It is much cleaner to just start with an empty gpg environment (and compared to key generation that is a cheap operation).

Also i sure hope we do not call `sudo` in our tests.

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,155 @@
...
+
+KEY = "https://raw.githubusercontent.com/pulp/pulp-fixtures/master/common/GPG-PRIVATE-KEY-pulp-qe"
+
+
+@pytest.fixture(scope="session")
+def tempdir():
+    """A temporary directory."""
+    directory = tempfile.mkdtemp()
+    yield directory
+    shutil.rmtree(directory)
```
mdellweg, line 22:
```suggestion
    with tempfile.TemporaryDirectory() as directory:
        yield directory
```
Context managers are _made_ for this...

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@pytest.fixture(scope="session")
+def gpg_metadata(tempdir):
+    """A fixture for a GPG instance and related metadata."""
+    private_key = requests.get(KEY).text
+
+    gpg = gnupg.GPG(gnupghome=tempdir)
```
mdellweg, line 30: :100:

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,155 @@
...
+
+    gpg.import_keys(private_key)
+
+    fingerprint = gpg.list_keys()[0]["fingerprint"]
+    keyid = gpg.list_keys()[0]["keyid"]
+
+    gpg.trust_keys(fingerprint, "TRUST_ULTIMATE")
+
+    return gpg, (fingerprint, keyid)
```
mdellweg, line 39: I think this is a result of some evolution, right? Maybe just return
```suggestion
    return gpg, fingerprint, keyid
```

Or we create a named tuple for this?
But we can also take it as is and think about reusability later.

mdellweg: I really like the key handling as a fixture, and eventually i want to see this move to pulp_smash in a way that can be used by pulp_ansible tests too.

pulp_container/tests/functional/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    AuthenticationHeaderQueries,
+    BearerTokenAuth,
+)
+
+
+@pytest.fixture
+def fetch_response(bindings_cfg):
+    """Return a response considering token authentication."""
+
+    def _fetch_response(method, url, **kwargs):
```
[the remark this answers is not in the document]
lubosmj: It is in a separate file because `fetch_response` is used in two different places (conftest.py, test_push_signatures) and we are importing some classes from `pulp_container.tests.functional.utils`.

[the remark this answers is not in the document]
lubosmj: I would expect `local_registry` to be used for "fundamental" podman operations, not for working with raw requests.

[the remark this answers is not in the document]
mdellweg: Well all of the things in the other conftest are also only used for the functional tests, right? Maybe that is the right location for the conftest after all.
Isn't that special token handling only interesting to interactions with our local_registry? But you may be right in that we do not need podman installed to do this, but we need it to use the local_registry fixture.
My concern with this fixture is, that the name does not speak to me. It should reflect that we talk to the registry and that we handle the authentication. And ideally it would get the relative path only and add the local registry automatically.

pulp_container/tests/functional/api/test_rbac_repo_versions.py
```diff
@@ -275,41 +270,19 @@ def test_cross_repository_blob_mount(
...
-                },
-                auth=basic_auth,
-            )
-            response.raise_for_status()
-            token = response.json()["token"]
-            auth = BearerTokenAuth(token)
-
-        response = requests.post(url, auth=auth)
+        mount_path = f"/v2/{dest}/blobs/uploads/?from={source}&mount={blob.digest}"
+        response, auth = local_registry.get_response(
```
mdellweg, line 278: mount_blob using local_registry sounds exactly right to me!
:+1:

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    # push another signature with a different timestamp
+    registry_client.tag(image_path, local_image_path)
+    registry_client.login(
+        "-u", bindings_cfg.username, "-p", bindings_cfg.password, local_registry.name
+    )
+    subprocess.check_output(
+        [
+            "sudo",
```
mdellweg, line 80: Have you tried not to sudo at all?

lubosmj: Yes. But, every other podman call is invoked with "sudo". So, I decided to stick to the standard. Besides that, I could easily set the environment variable with a GnuPG home directory.

lubosmj: Oh, you mean using something like `monkeypatch.setenv(name, value, prepend=False)` from https://docs.pytest.org/en/6.2.x/monkeypatch.html to set the environment variable? :open_mouth: That looks interesting.

lubosmj: This does not seem to be working:
```
    monkeypatch.setenv("GNUPGHOME", gpg.gnupghome)
    local_registry.push(image_path, "test-1:manifest_a", "--sign-by", keyid)
```
I can see the environment variable through `os.getenv("GNUPGHOME")`. No idea why podman cannot see the variable...

```
(pulp) [vagrant@pulp3-source-fedora35 ~]$ gpg --home /tmp/tmpbvcflazb --list-keys --keyid-format LONG
/tmp/tmpbvcflazb/pubring.kbx
----------------------------
pub   rsa2048/**05A5E6DA269D9D98** 2016-08-05 [SC]
      6EDF301256480B9B801EBA3D05A5E6DA269D9D98
uid                 [ultimate] Pulp QE
sub   rsa2048/E8DC6B59B70CEECB 2016-08-05 [E]
```
```
os.getenv("GNUPGHOME")
Out[11]: '/tmp/tmpbvcflazb'
local_registry.push(image_path, "test-1:manifest_a", "--sign-by", keyid)
Traceback (most recent call last):
  File "/usr/local/lib/pulp/lib64/python3.10/site-packages/IPython/core/interactiveshell.py", line 3369, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-12-c8e783671e35>", line 1, in <cell line: 1>
    local_registry.push(image_path, "test-1:manifest_a", "--sign-by", keyid)
  File "/home/vagrant/devel/pulp_container/pulp_container/tests/conftest.py", line 128, in push
    registry_client.push(local_image_path, *args)
  File "/usr/local/lib/pulp/lib64/python3.10/site-packages/pulp_smash/cli.py", line 1028, in _dispatch_command
    result = self._client.run(cmd)
  File "/usr/local/lib/pulp/lib64/python3.10/site-packages/pulp_smash/cli.py", line 320, in run
  File "/usr/local/lib/pulp/lib64/python3.10/site-packages/pulp_smash/cli.py", line 54, in code_handler
  File "/usr/local/lib/pulp/lib64/python3.10/site-packages/pulp_smash/cli.py", line 128, in check_returncode
    raise exceptions.CalledProcessError(
pulp_smash.exceptions.CalledProcessError: Command ('podman', 'push', 'pulp3-source-fedora35.localhost.example.com:443/test-1:manifest_a', '--sign-by', '05A5E6DA269D9D98') returned non-zero exit status 125.
```

lubosmj: Okay, I need to do the following in order to set the environment variable:
```
registry_client._client.machine.env["GNUPGHOME"] = gpg.gnupghome
```
I am pushing the changes and merging the PR since it reached 45 comments.

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -0,0 +1,119 @@
...
+    registry_client.pull(f"{REGISTRY_V2_REPO_PULP}:manifest_a")
+
+    image_path = f"{REGISTRY_V2_REPO_PULP}:manifest_a"
+    gpg, fingerprint, keyid = gpg_metadata
+
+    with registry_client._client.machine.env(GNUPGHOME=gpg.gnupghome):
+        local_registry.push(image_path, "test-1:manifest_a", "--sign-by", keyid)
+
+        # push the same image for the second time with a different signature (timestamp)
+        local_registry.push(image_path, "test-1:manifest_a", "--sign-by", keyid)
```
mdellweg, line 57: This looks nice.
BTW, heads up: On the other PR it was suggested to call this `tag_and_push`. Don't change it yet.

mdellweg: I'm now wondering why this works without the pulp-smash change.

PR #635, opened 2022-03-16
Transform RBAC repository content test to pytest

mdellweg:
[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_push_content.py
+++ b/pulp_container/tests/functional/api/test_push_content.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 """Tests that verify that images can be pushed to Pulp."""
+import pytest
 import unittest
 
 from urllib.parse import urlparse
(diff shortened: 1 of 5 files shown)


pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -16,313 +17,345 @@
...
+    with user_dist_consumer:
+        assert container_push_repository_api.list(name=repo_name).count == 1
+    with user_namespace_collaborator:
+        assert container_push_repository_api.list(name=repo_name).count == 1
+    with user_reader:
+        assert container_push_repository_api.list(name=repo_name).count == 1
+    # TODO Can someone verify that this repository should be seen by any user?
+    # Is public default?
+    # with user_helpless:
+    #     assert container_push_repository_api.list(name=repo_name).count == 1
```
lubosmj, line 138: Yes, users with no permissions should see the repository.

I tested this scenario locally. I pushed an image as an administrator. Then, I tried to access the push-repository as a user with no permissions via the Pulp API.

pulp_container/tests/functional/api/test_rbac_repo_content.py
```diff
@@ -1,215 +1,157 @@
...
+    container_remote_api,
+    container_repository_api,
+    container_push_repository_api,
+    container_distribution_api,
+    container_namespace_api,
+    container_tag_api,
+):
+    """Assert that certain users can list and read content."""
+
+    user_creator = gen_user(
```
ipanova, line 37: i don't seem to see where we cleanup created users?

pulp_container/tests/functional/api/test_rbac_repo_content.py
```diff
@@ -1,215 +1,157 @@
...
-        cls.repository = cls.user_creator["repository_api"].create(
-            ContainerContainerRepository(**gen_repo())
-        )
-        cls.remote = cls.user_creator["remote_api"].create(
-            gen_container_remote(upstream_name=PULP_FIXTURE_1)
-        )
-        sync_data = ContainerRepositorySyncURL(remote=cls.remote.pulp_href)
-        sync_response = cls.user_creator["repository_api"].sync(cls.repository.pulp_href, sync_data)
+def test_rbac_repository_content(
+    delete_orphans_pre,  # TODO evaluate
```
ipanova, line 22: we should probably call clean up so then later when we list tags,  orphaned tags do not interfere?

pulp_container/tests/functional/api/test_push_content.py
ipanova: I think we have a mess here. It seems like we allow users to see container push repo if the distribution is public, however when it comes to the content queryset, I don't think we show such content

pulp_container/tests/functional/api/test_rbac_repo_content.py
lubosmj: A clean-up procedure is already included in the fixture: https://github.com/pulp/pulp-smash/blob/7f8d1de5fb5fba221f3de98e001bc4990009eb42/pulp_smash/pulp3/pytest_plugin/__init__.py#L369.

pulp_container/tests/functional/api/test_rbac_repo_content.py
```diff
@@ -1,215 +1,157 @@
...
+    user_reader3 = gen_user(model_roles=["container.containerdistribution_consumer"])
+    user_helpless = gen_user()
+
+    # create a push repo with user_creator
+    image_path1 = f"{REGISTRY_V2_REPO_PULP}:manifest_a"
+    registry_client.pull(image_path1)
+    repo_name1 = "testcontent1/perms"
+    local_url1 = f"{repo_name1}:1.0"
+    with user_creator:
+        local_registry.push(image_path1, local_url1)
```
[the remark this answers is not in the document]
ipanova: This is not explicit but turns out that `local_registry.push` also tags the image, hence image_path1 is needed. Does it make sense to maybe rename this to `local_registry.tag_and_push`?

pulp_container/tests/functional/api/test_push_content.py
ipanova: we also don't allow "helpless users" to retrieve a specific public-push-repo, just list
```
$ http  https://pulp3-source-fedora34.fluffy.example.com/pulp/api/v3/repositories/container/container-push/ --ignore-netrc --auth pupsik:12345678!
HTTP/1.1 200 OK
Allow: GET, HEAD, OPTIONS
Connection: keep-alive
Content-Length: 561
Correlation-ID: df8fac8b819b49c1a68e659abb91e00f
Date: Tue, 12 Apr 2022 10:43:02 GMT
Referrer-Policy: same-origin
Server: nginx/1.20.1
Strict-Transport-Security: max-age=15768000
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "description": null,
            "latest_version_href": "/pulp/api/v3/repositories/container/container-push/d7adb15a-a7e5-4681-a220-724192275814/versions/5/",
            "manifest_signing_service": null,
            "name": "test/test-repo",
            "pulp_created": "2022-04-12T10:15:37.064498Z",
            "pulp_href": "/pulp/api/v3/repositories/container/container-push/d7adb15a-a7e5-4681-a220-724192275814/",
            "pulp_labels": {},
            "retain_repo_versions": null,
            "versions_href": "/pulp/api/v3/repositories/container/container-push/d7adb15a-a7e5-4681-a220-724192275814/versions/"
        }
    ]
}


(pulp) [vagrant@pulp3-source-fedora34 pulpcore]$ http  https://pulp3-source-fedora34.fluffy.example.com/pulp/api/v3/repositories/container/container-push/d7adb15a-a7e5-4681-a220-724192275814/ --ignore-netrc --auth pupsik:12345678!
HTTP/1.1 403 Forbidden
Allow: GET, PUT, PATCH, HEAD, OPTIONS
Connection: keep-alive
Content-Length: 63
Correlation-ID: 3e8e2d524ec84f13b14d8f727b0ab495
Date: Tue, 12 Apr 2022 10:43:08 GMT
Referrer-Policy: same-origin
Server: nginx/1.20.1
Vary: Accept, Cookie
X-Frame-Options: DENY

{
    "detail": "You do not have permission to perform this action."
}


```

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -16,313 +17,345 @@
...
-        distribution = self.user_helpless["distribution_api"].read(created_resources[0])
+        distribution = container_distribution_api.read(created_resources[0])
 
-        # cleanup, namespace removal also removes related distributions
-        namespace = self.namespace_api.list(name=namespace_name).results[0]
-        self.addCleanup(self.namespace_api.delete, namespace.pulp_href)
+        add_to_cleanup(container_namespace_api, distribution.namespace)
 
 
 class PushManifestListTestCase(PulpTestCase, rbac_base.BaseRegistryTest):
```
ipanova, line 361: not sure this test belongs here? we only test with admin perms so maybe it should be moved out?

mdellweg: To be fair, the file is not called `test_rbac_push_content`.

pulp_container/tests/functional/api/test_push_content.py
mdellweg: As far as i remember, the test still allows helpless to see the repository and i do not understand how.

ipanova: for example here https://github.com/pulp/pulp_container/blob/main/pulp_container/app/viewsets.py#L1147 maybe we should require only `has_namespace_or_obj_perms:container.view_containerdistribution` because this is purely request to Pulp API

mdellweg: Good question. I thought we opted for consistency between the two API's. But i agree we should _decide_ eventually.

At least this explains why the test will show that `user_helpless` can see the repo.

ipanova: I agree that we decided to go  with consistency. But I start questioning whether this is worth the additional complexity we are carrying - for example  we have statement - 'I can pull on public repo with podman pull' .In pulp language it is  translated to 'private' flag on the distribution, and right now we are showing only push-repo types that have public distribution but we forgot about 'regular' repos. 
We also forgot to include 'public' distribution when we compose content queryset . By the logic we have, one should be able to pull content from public repo, as well as see it from pulp API. Right now it is not the case https://github.com/pulp/pulp_container/blob/main/pulp_container/app/viewsets.py#L177 
That's one of few oversights and inconsistencies I've noticed, but I am pretty sure we have more :/

ipanova: Btw, I am not suggesting to address these concerns in this PR. But the test will need to be updated once more, when/if we do that.

lubosmj (approved):

PR #649, opened 2022-03-22
Workaround pipeline maxsize issues

mdellweg:
A recent change in pulpcore to reduce the maxsize in the sync pipeline
to 1 leads to uniqueness constraint violations in pulp_container. This
is a workaround to revert that change specifically in the plugin.

[noissue]

### DIFF
--- a/pulp_container/app/tasks/synchronize.py
+++ b/pulp_container/app/tasks/synchronize.py
@@ -1,5 +1,7 @@
 from gettext import gettext as _
+import asyncio
 import logging
+import tempfile
 
 from pulpcore.plugin.stages import (
     ArtifactDownloader,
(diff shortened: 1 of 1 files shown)


pulp_container/app/tasks/synchronize.py
```diff
@@ -77,3 +82,23 @@ def pipeline_stages(self, new_version):
...
+
+        # This method is copied from pulpcore with the addition of maxsize=100 to circumvent sync
+        # issues with remote artifact creation.
+        with tempfile.TemporaryDirectory(dir="."):
+            with self.repository.new_version() as new_version:
+                loop = asyncio.get_event_loop()
+                stages = self.pipeline_stages(new_version)
+                stages.append(ContentAssociation(new_version, self.mirror))
+                stages.append(EndStage())
+                pipeline = create_pipeline(stages, maxsize=100)
```
ipanova, line 101: if you're talking specifically about reverting the change then it should be 1000

mdellweg:
Let's see if this is still needed now with
https://github.com/pulp/pulpcore/pull/2382

mdellweg:
Fixed in pulpcore.

PR #652, opened 2022-03-24
Add tests for syncing signatures and implement validation for signature payload

lubosmj:


pulp_container/tests/functional/api/test_sync_signatures.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if request.param:
+        data["sigstore"] = request.param
+
+    remote = gen_object_with_cleanup(container_remote_api, data)
+
+    data = ContainerContainerRepository(**gen_repo())
+    repository = gen_object_with_cleanup(container_repository_api, data)
+
+    signed_only = True if request.param else False
```
ipanova, line 43: It does not mean that signed_only is flipped automatically to True if sigstore is provided. It can happen that not all content is signed so we sync all content and available signatures, unless signed_only is set to True then we sync only content that has signatures + its signatures.

pulp_container/app/json_schemas.py
```diff
@@ -0,0 +1,54 @@
...
+                    "type": "string"
+                },
+                "timestamp": {
+                    "type": "number",
+                    "minimum": 0
+                }
+            }
+        }
+    },
+    "required": ["critical"],
```
ipanova, line 52: i think optional is a required field too, it just can be an empty json

pulp_container/app/json_schemas.py
```diff
@@ -0,0 +1,54 @@
+SIGNATURE_SCHEMA = """{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://example.com/product.schema.json",
+    "title": "Signature",
+    "description": "JSON signature payload",
```
ipanova, line 5: json shema validation for the signature payload

pulp_container/app/utils.py
```diff
@@ -89,16 +92,12 @@ def extract_data_from_signature(signature_raw, man_digest):
...
-                    signature_type, man_digest
-                )
-            )
-        )
+    errors = []
+    for error in validator.iter_errors(sig_json):
+        errors.append(error.message)
+
+    if errors:
+        log.info(_("The signature for {} is not valid due to: {}".format(man_digest, errors)))
```
ipanova, line 100: please keep in the message the info about signature is not going to be synced.

pulp_container/app/utils.py
```diff
@@ -5,13 +5,16 @@
...
+from jsonschema import Draft7Validator
 from rest_framework.exceptions import Throttled
 
 from pulpcore.plugin.models import Task
 
-from pulp_container.constants import SIGNATURE_TYPE
+from pulp_container.app.json_schemas import SIGNATURE_SCHEMA
 
 
+validator = Draft7Validator(json.loads(SIGNATURE_SCHEMA))
```
ipanova, line 16: niice :seal:

pulp_container/tests/functional/api/test_sync_signatures.py
ipanova: you can add a test though with no sigstoreurl and signed_only=true

ipanova: there is also one more test, where there is signatore url and signed_only=true but some content turns out not to be signed but i think in the deprecated repo all the content is signed..

ipanova:
please file issue for 2nd commit as we discussed

pulp_container/app/json_schemas.py
```diff
@@ -0,0 +1,54 @@
...
+        },
+        "optional": {
+            "type": "object",
+            "properties": {
+                "creator": {
+                    "type": "string"
+                },
+                "timestamp": {
+                    "type": "number",
+                    "minimum": 0
```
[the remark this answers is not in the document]
lubosmj: The timestamp should be always greater than 0.

pulp_container/tests/functional/api/test_sync_signatures.py
lubosmj: > It does not mean that signed_only is flipped automatically to True if sigstore is provided. It can happen that not all content is signed so we sync all content and available signatures, unless signed_only is set to True then we sync only content that has signatures + its signatures.
Yes, I know that it does not mean that. I wanted only to initialize a fixture based on the passed sigstore URL. I can update the fixture to consider more scenarios.

ipanova: and this is exactly what i am saying - the fixture right now is flipping the signed_only to true if url to sigstore is present :)

pulp_container/tests/unit/test_json_schemas.py
```diff
@@ -0,0 +1,106 @@
...
+                "identity": {
+                    "docker-reference": "docker.io/library/busybox:latest"
+                }
+            },
+            "optional": {}
+        }"""
+        signature = json.loads(signature)
+        self.assertEqual(list(validator.iter_errors(signature)), [])
+
+    def test_missing_optional(self):
```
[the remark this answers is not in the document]
lubosmj: No. We have two tests: test_missing_optional_fields and test_missing_optional. The first is for testing the presence of optional fields. The latter one is for testing the presence of the `optional` section.

lubosmj:
Please, do not merge this PR until https://github.com/pulp/pulp_container/pull/633 will be merged.

pulp_container/tests/unit/test_json_schemas.py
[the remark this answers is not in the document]
lubosmj: Yikes, I forgot to write a test for this. We were supposed to test a missing `optional` attribute.

PR #694, opened 2022-04-08
Fix not using the worker working directory

dralley:
[noissue]

### DIFF
--- a/pulp_container/app/tasks/builder.py
+++ b/pulp_container/app/tasks/builder.py
@@ -113,7 +113,7 @@ def build_image_from_containerfile(
     containerfile = Artifact.objects.get(pk=containerfile_pk)
     repository = ContainerRepository.objects.get(pk=repository_pk)
     name = str(uuid4())
-    with tempfile.TemporaryDirectory(".") as working_directory:
+    with tempfile.TemporaryDirectory(dir=".") as working_directory:
         path = "{}/".format(working_directory.path)
         for key, val in artifacts.items():
             artifact = Artifact.objects.get(pk=key)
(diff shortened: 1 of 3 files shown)


dralley:
Q: does it need to be fixed here instead or is this code intending to use `/tmp/`

pulp_container/app/tasks/builder.py
```diff
@@ -113,7 +113,7 @@ def build_image_from_containerfile(
     containerfile = Artifact.objects.get(pk=containerfile_pk)
     repository = ContainerRepository.objects.get(pk=repository_pk)
     name = str(uuid4())
-    with tempfile.TemporaryDirectory(".") as working_directory:
```
dralley, line 116: First kwarg is not `dir`, it's `suffix`, even though the former would make way more sense : /

lubosmj (approved):

mdellweg (approved):

mdellweg:
Actually this may need a changelog entry. Do we want to backport it?

ipanova:
@dralley both of instances are tasks so they meant to use workers' dir. Please file an issue. We'd want to backport this to 2.11

dralley:
@ipanova @mdellweg Done

dralley:
My Q: was intended to apply to this line https://github.com/pulp/pulp_container/blob/b1725ad6bb8803dc293cd9938cac5bde443f19e5/pulp_container/app/models.py#L494=

ipanova:
> My Q: was intended to apply to this line https://github.com/pulp/pulp_container/blob/b1725ad6bb8803dc293cd9938cac5bde443f19e5/pulp_container/app/models.py#L494=

no, this one is correct, this code path can be called outside of task.

PR #695, opened 2022-04-08
WIP: Use global access conditions

mdellweg:
Required-PR: https://github.com/pulp/pulpcore/pull/2496


pulp_container/app/access_policy.py
```diff
@@ -2,167 +2,14 @@
...
-            )
-        )
-
-
-class NamespaceAccessPolicy(AccessPolicyFromDB):
-    """
-    Access policy for ContainerNamespaceViewSet.
-    """
-
-    def namespace_is_username(self, request, view, action):
```
ipanova, line 158: we at least will need to re-write namespace_is_username and has_distribution_perms because we have multiple implementations of those depending which viewset it is meant for.

ipanova: you already did that :+1:

mdellweg: Yes. Flake is rather picky with multiply defined functions.

ipanova:
I like it, it looks so much cleaner

mdellweg:
It's actually an important change to be able to configure a different Permission Class for pulp.

ipanova:
I've taken another pass at this PR. The core part is already merged. Is there anything else you wanted to add or this is ready for review?

mdellweg:
You are right. It is ready now. I should add a requirements bump on pulpcore though. Also there is still the `RegistryAccessPolicy`, not sure if that one is supposed to go anywhere soon.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # remember to also update unittest_requirements.txt when updating this file
-pulpcore>=3.18.0.dev
+pulpcore>=3.19.0.dev
```
ipanova, line 2: :+1:

ipanova:
> You are right. It is ready now. I should add a requirements bump on pulpcore though. Also there is still the `RegistryAccessPolicy`, not sure if that one is supposed to go anywhere soon.

it's being used in `AuthorizationService` https://github.com/pulp/pulp_container/blob/main/pulp_container/app/authorization.py#L47
I am not sure sure about this access_policy_class class sttribute though https://github.com/pulp/pulp_container/blob/main/pulp_container/app/registry_api.py#L447

ipanova (approved):

PR #708, opened 2022-04-14
Remove getttext calls within log messages

### LINKED ISSUE #690: Audit log statements for i18n
https://github.com/pulp/pulpcore/issues/2477

lubosmj:
closes #690

### DIFF
--- a/CHANGES/690.feature
+++ b/CHANGES/690.feature
@@ -1,0 +1,1 @@
+Log messages are now not being translated.
--- a/pulp_container/app/downloaders.py
+++ b/pulp_container/app/downloaders.py
@@ -1,5 +1,3 @@
-from gettext import gettext as _
-
 import aiohttp
 import asyncio
 import json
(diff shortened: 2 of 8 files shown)


lubosmj:
I am not sure whether we should also refactor i18l messages within RuntimeError exceptions. These types of messages are not directly shown to users but rather to administrators in standard error logs. Do you think I should also remove the remaining `gettext` calls from any other uncaught exceptions (which result in HTTP 500 responses)?

mdellweg:
> I am not sure whether we should also refactor i18l messages within RuntimeError exceptions. These types of messages are not directly shown to users but rather to administrators in standard error logs. Do you think I should also remove the remaining `gettext` calls from any other uncaught exceptions (which result in HTTP 500 responses)?

That is not a bad idea. But i want to see more peoples opinions.

ipanova:
> > I am not sure whether we should also refactor i18l messages within RuntimeError exceptions. These types of messages are not directly shown to users but rather to administrators in standard error logs. Do you think I should also remove the remaining `gettext` calls from any other uncaught exceptions (which result in HTTP 500 responses)?
> 
> That is not a bad idea. But i want to see more peoples opinions.

+1. @bmbouter since you're working on the pulpcore issue

lubosmj:
We concluded that if it is either a log entry or traceback, we should not use i8ln.

mdellweg:
Nitpick `s/getttext/gettext/` in the commit message.

PR #717, opened 2022-04-19
Validate unknown fields

### LINKED ISSUE #475: Validate unknown fields
Author: @lubosmj (lmjachky)

Redmine Issue: 7546, https://pulp.plan.io/issues/7546

---

At the moment, we do not validate unknown fields in serializers. For example, this call is executed without any errors:

~~~ text
http :24817/pulp/api/v3/repositories/container/container/bb221b58-bcfa-45ae-96ea-363ca381da47/tag/ tag="tag" digest=$MANIFEST_DIGEST blabla="halabala"
~~~

I propose to add the validation by leveraging `pulpcore.plugin.serializers.ValidateFieldsMixin`.

lubosmj:
closes #475

### DIFF
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -27,7 +27,7 @@ from pulpcore.plugin.serializers import (
     RepositorySyncURLSerializer,
     RepositoryVersionRelatedField,
     SingleArtifactContentSerializer,
-    validate_unknown_fields,
+    ValidateFieldsMixin,
 )
 
 from pulp_container.app import models
@@ -366,7 +366,7 @@ class ContainerDistributionSerializer(DistributionSerializer):
         )
 
 
-class TagOperationSerializer(serializers.Serializer):
+class TagOperationSerializer(ValidateFieldsMixin, serializers.Serializer):
     """
     A base serializer for tagging and untagging manifests.
     """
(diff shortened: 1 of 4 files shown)


pulp_container/app/serializers.py
```diff
@@ -641,6 +610,8 @@ def validate(self, data):
         """
         Validate and extract the latest repository version, signatures from the passed data.
         """
+        super().validate(data)
+
         new_data = {}
         new_data.update(self.initial_data)
```
mdellweg, line 616:
```suggestion
        new_data = super().validate(data)
```

pulp_container/app/viewsets.py
```diff
@@ -459,10 +462,11 @@ def untag(self, request, pk):
         Create a task which is responsible for untagging an image.
         """
         repository = self.get_object()
-        request.data["repository"] = repository
+        repository_version = repository.latest_version()
```
mdellweg, line 465: Why don't we add the repository to the serializer context instead?
This kind of implies that we could tag something from a baseversion, which we cannot. (With retained_version=1, the task may see a completely different single version of this repo as the serializer.)

lubosmj: We cannot pass objects to a serializer directly in this case. There are two options:
1. Pass the HREF or PK of a repository as a string to the serializer (we will need to add a new serializer field that will be visible in the REST API documentation),
2. Pass a repository object to a serializer with the context argument (https://www.django-rest-framework.org/api-guide/serializers/#including-extra-context).

The second option looks better to me. Also, we are doing this everywhere where we are dealing with repository operations.

Yes, I should rename `repository_version` back to `latest_version`.

mdellweg: I agree for option 2. My point is just we pass the repository (not the version) to the task, therefore we should use it to validate too.
So we would pass `context={"request": request, "repository": repository}`

mdellweg (approved):
LGTM

PR #719, opened 2022-04-20
Fix the reference to a serializer

### LINKED ISSUE #718: An incorrect serializer is used for the `build_image` method
https://github.com/pulp/pulp_container/blob/dbc25c83311800742b1715d3c052bc2e2c3bc8cb/pulp_container/app/viewsets.py#L833-L851

lubosmj:
closes #718

### DIFF
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -420,7 +420,6 @@ class TagOperationsMixin:
         description="Trigger an asynchronous task to tag an image in the repository",
         summary="Create a Tag",
         responses={202: AsyncOperationResponseSerializer},
-        request=serializers.TagImageSerializer,
     )
     @action(detail=True, methods=["post"], serializer_class=serializers.TagImageSerializer)
     def tag(self, request, pk):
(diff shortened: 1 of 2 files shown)


pulp_container/app/viewsets.py
```diff
@@ -839,7 +839,7 @@ def copy_manifests(self, request, pk):
         responses={202: AsyncOperationResponseSerializer},
         request=serializers.OCIBuildImageSerializer,
```
mdellweg, line 840: Can you check whether this can be omitted now? Spectacular should derive it from the action.

lubosmj: According to the documentation (https://drf-spectacular.readthedocs.io/en/latest/customization.html?highlight=discover#step-2-extend-schema), we can leave the parameter `request` empty: ` you only need to override what was not properly discovered in the introspection`.

lubosmj: The serializer is correctly detected in the REST documentation too.

mdellweg (approved):

PR #735, opened 2022-04-28
Add tests for signing manifests with a service

### LINKED ISSUE #687: Add a test case for manifest_signing_service scenario
Test that one can sign with manifest_signing_service 1. unsigned content 2. previously signed content whether it was synced or pushed.

lubosmj:
closes #687

### DIFF
--- a/pulp_container/tests/functional/api/test_push_signatures.py
+++ b/pulp_container/tests/functional/api/test_push_signatures.py
@@ -1,39 +1,10 @@
 """Tests that verify that an image signature can be pushed to Pulp."""
-import gnupg
 import base64
 import json
-import requests
 import pytest
-import tempfile
 
 from pulp_container.tests.functional.constants import REGISTRY_V2_REPO_PULP
 from pulp_container.constants import SIGNATURE_TYPE
-
-KEY = "https://raw.githubusercontent.com/pulp/pulp-fixtures/master/common/GPG-PRIVATE-KEY-pulp-qe"
-
-
-@pytest.fixture(scope="session")
-def tempdir():
-    """A temporary directory."""
-    with tempfile.TemporaryDirectory() as directory:
-        yield directory
-
-
-@pytest.fixture(scope="session")
-def gpg_metadata(tempdir):
-    """A fixture for a GPG instance and related metadata."""
-    private_key = requests.get(KEY).text
-
-    gpg = gnupg.GPG(gnupghome=tempdir)
-
-    gpg.import_keys(private_key)
-
-    fingerprint = gpg.list_keys()[0]["fingerprint"]
-    keyid = gpg.list_keys()[0]["keyid"]
-
-    gpg.trust_keys(fingerprint, "TRUST_ULTIMATE")
-
-    return gpg, fingerprint, keyid
 
 
 @pytest.fixture
@@ -41,17 +12,17 @@ def distribution(
     registry_client,
     local_registry,
     container_distribution_api,
-    gpg_metadata,
+    signing_gpg_metadata,
     add_to_cleanup,
 ):
     """Return a distribution created after pushing a signed content to the Pulp Registry."""
     if registry_client.name != "podman":
         pytest.skip("This test requires podman to sign pulled content", allow_module_level=True)
 
-    registry_client.pull(f"{REGISTRY_V2_REPO_PULP}:manifest_a")
-
     image_path = f"{REGISTRY_V2_REPO_PULP}:manifest_a"
-    gpg, fingerprint, keyid = gpg_metadata
+    registry_client.pull(image_path)
+
+    gpg, fingerprint, keyid = signing_gpg_metadata
 
     with registry_client._client.machine.env(GNUPGHOME=gpg.gnupghome):
         local_registry.tag_and_push(image_path, "test-1:manifest_a", "--sign-by", keyid)
@@ -70,11 +41,11 @@ def test_assert_signed_image(
     container_push_repository_api,
     container_manifest_api,
     container_signature_api,
-    gpg_metadata,
+    signing_gpg_metadata,
     distribution,
 ):
     """Test whether an admin user can fetch a signature from the Pulp Registry."""
-    gpg, fingerprint, keyid = gpg_metadata
+    gpg, fingerprint, keyid = signing_gpg_metadata
 
     repository = container_push_repository_api.read(distribution.repository)
     manifest = container_manifest_api.list(
--- a/pulp_container/tests/functional/api/test_sign_manifests.py
+++ b/pulp_container/tests/functional/api/test_sign_manifests.py
@@ -1,0 +1,57 @@
+import pytest
+
+from pulp_smash.pulp3.bindings import monitor_task
+
+from pulpcore.client.pulp_container import RepositorySign
+
+from pulp_container.constants import SIGNATURE_TYPE
+from pulp_container.tests.functional.constants import REGISTRY_V2_REPO_PULP
+
+MANIFEST_TAG = "manifest_a"
+
+
+@pytest.fixture
+def distribution(registry_client, local_registry, container_distribution_api, add_to_cleanup):
+    """The fixture for a distribution that references a repository of the push type."""
+    image_path = f"{REGISTRY_V2_REPO_PULP}:{MANIFEST_TAG}"
+    registry_client.pull(image_path)
+    local_registry.tag_and_push(image_path, f"test-1:{MANIFEST_TAG}")
+
+    distribution = container_distribution_api.list(name="test-1").results[0]
+    add_to_cleanup(container_distribution_api, distribution.pulp_href)
+
+    return distribution
+
+
+def test_sign_manifest(
+    signing_gpg_metadata,
+    distribution,
+    container_signing_service,
+    container_push_repository_api,
+    container_signature_api,
+    container_tag_api,
+    container_manifest_api,
+):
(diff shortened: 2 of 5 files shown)


lubosmj:
The work on this PR is postponed until we decide how to test signing services across all plugins.

At the moment, it is not possible to customize the GnuPG home directory or keyring from tests. Using side channels for setting environment variables might alleviate the problem, but it will not provide a robust solution. Also, in our CI, the tests are not running in the same environment as Pulp. The current test case is not passing in the CI because `gpg` reads keys from a default home directory.

lubosmj:
Maybe, the issue https://github.com/pulp/pulpcore/issues/2476 should be resolved first.

pulp_container/tests/functional/constants.py
```diff
@@ -40,3 +40,5 @@
 
 REGISTRY_V2_REPO_PULP = f"{REGISTRY_V2}/{PULP_FIXTURE_1}"
 REGISTRY_V2_REPO_HELLO_WORLD = f"{REGISTRY_V2}/{PULP_HELLO_WORLD_REPO}"
+
+KEY = "https://raw.githubusercontent.com/pulp/pulp-fixtures/master/common/GPG-PRIVATE-KEY-pulp-qe"
```
mdellweg, line 44: Let's be a bit more descriptive with that constants name. How about GPG_TEST_KEY, TEST_PRIVATE_KEY, ...?

pulp_container/tests/functional/api/test_sign_manifests.py
```diff
@@ -0,0 +1,88 @@
...
+    assert signing_service.pubkey_fingerprint == fingerprint
+    assert signing_service.public_key == gpg.export_keys(keyid)
+
+    yield signing_service
+
+    cmd = (
+        "from pulpcore.app.models import SigningService;"
+        f"SigningService.objects.filter(name='{service_name}').delete()"
+    )
+    execute_pulpcore_python(cli_client, cmd)
```
mdellweg, line 70: This shows a gap. Seems like everyone wanted to have a signing service, no one ever wanted to delete one.

.github/workflows/scripts/install.sh
```diff
@@ -66,6 +66,7 @@ services:
     volumes:
       - ./settings:/etc/pulp
       - ./ssh:/keys/
+      - /tmp/:/tmp/
```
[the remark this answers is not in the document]
lubosmj: It is just a hacky way to bypass copying files between a host and container.

I believed that mounting the `/tmp` directory should be fine even for other plugins (moving it to the plugin template might be reasonable). This hack will allow us to fiddle with data created inside the host environment.

.ci/ansible/Containerfile.j2
```diff
@@ -33,4 +33,6 @@ RUN export plugin_path="$(pip3 show {{ item.name }} | sed -n -e 's/Location: //p
     ln $plugin_path/app/webserver_snippets/nginx.conf /etc/nginx/pulp/{{ item.name }}.conf || true
 {% endfor %}
 
+RUN yum install -y skopeo
+
```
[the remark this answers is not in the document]
lubosmj: Skopeo is required only by pulp_container. Moving it to the plugin template might not be needed.

pulp_container/tests/functional/api/test_sign_manifests.py
lubosmj: Do we want to move the creation and removal of signing services to pulpcore as separate fixtures?

mdellweg: No, i'm just saying that we provide a command to create one, but not for deletion, which is an asymmetry.

.github/workflows/scripts/install.sh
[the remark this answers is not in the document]
lubosmj: I am linking a pulp-smash issue where we can discuss the further details: https://github.com/pulp/pulp-smash/issues/1295.

.ci/ansible/Containerfile.j2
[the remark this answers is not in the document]
mdellweg: I know. But we also install pulp_rpm (yum-)dependencies in the base container. I'm not even talking about the plugin template here.

.github/workflows/scripts/install.sh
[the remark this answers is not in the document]
mdellweg: But that is the point, it's the host and the container and therefore specific to the ci setup. We use pulp-smash to abstract away from the specific setup, and that is why we need to use it here.

bmbouter:
I am hoping that the tests can be written to just use the API assuming a SigningService is already created and in place. Then have the CI configure a signing service via the plugin_template. This is kind of like how the import/export tests work in the sense that they require a server_side config change that is a non-default. So any system under test, needs that config change, they would also need a SigningService configured.

Also by using `pytest.mark` you could make a label that could disinclude tests that don't have a SigningService, or maybe write a pytest hook that skips those tests if the server under test doesn't have a SigningService configured.

pulp_container/tests/functional/api/test_sign_manifests.py
```diff
@@ -0,0 +1,88 @@
...
+from pulp_smash.pulp3.bindings import monitor_task
+from pulp_smash.utils import execute_pulpcore_python, uuid4
+
+from pulpcore.client.pulp_container import RepositorySign
+
+from pulp_container.tests.functional.constants import REGISTRY_V2_REPO_PULP
+
+
+@pytest.fixture
+def distribution(registry_client, local_registry, container_distribution_api, add_to_cleanup):
```
ipanova, line 14: should not fixtures live in conftes.py?

pulp_container/tests/functional/api/test_sign_manifests.py
```diff
@@ -0,0 +1,88 @@
...
+    registry_client.pull(image_path)
+    local_registry.tag_and_push(image_path, "test-1:manifest_a")
+
+    distribution = container_distribution_api.list(name="test-1").results[0]
+    add_to_cleanup(container_distribution_api, distribution.pulp_href)
+
+    return distribution
+
+
+@pytest.fixture
```
ipanova, line 26: is this something that can be shared?

lubosmj: In case some other tests will require it, then yes. But right now, we do not use it from anywhere else.

lubosmj: Okay, will move it!

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -48,9 +19,9 @@ def distribution(
     if registry_client.name != "podman":
         pytest.skip("This test requires podman to sign pulled content", allow_module_level=True)
 
-    registry_client.pull(f"{REGISTRY_V2_REPO_PULP}:manifest_a")
-
     image_path = f"{REGISTRY_V2_REPO_PULP}:manifest_a"
+    registry_client.pull(image_path)
+
```
mdellweg, line 24: :+1:

pulp_container/tests/functional/conftest.py
```diff
@@ -145,6 +154,118 @@ def inspect(local_url):
...
 
+@pytest.fixture(scope="session")
+def tempdir():
+    """A temporary directory."""
+    with tempfile.TemporaryDirectory() as directory:
+        yield directory
+
+
+@pytest.fixture(scope="session")
+def gpg_metadata(tempdir):
```
mdellweg, line 165: There is a pulpcore fixture now (originally partly based on this design) can you look into using it?

https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/gpg_ascii_armor_signing_service.py

You'd probably need to wait for https://github.com/pulp/pulpcore/pull/2913 though.

mdellweg:
In its current form, this depends on a pulpcore change.

bmbouter:
Which change is that? Is it https://github.com/pulp/pulpcore/pull/2913 which merged?

mdellweg:
It's https://github.com/pulp/pulpcore/pull/2915 (also specified as a required pr).

pulp_container/tests/functional/conftest.py
```diff
@@ -145,6 +150,85 @@ def inspect(local_url):
     return _LocalRegistry()
 
 
+@pytest.fixture(scope="session")
+def tempdir():
+    """A temporary directory."""
+    with tempfile.TemporaryDirectory() as directory:
+        yield directory
+
+
```
mdellweg, line 159: This is not used. (Also there is a pytest fixture for this. Even on the session scope.)

mdellweg (commented):
I think this is almost good to go.

pulp_container/tests/functional/conftest.py
```diff
@@ -145,6 +149,78 @@ def inspect(local_url):
...
+    assert signing_service.pubkey_fingerprint == fingerprint
+    assert signing_service.public_key == gpg.export_keys(keyid)
+
+    yield signing_service
+
+    cmd = (
+        "from pulpcore.app.models import SigningService;"
+        f"SigningService.objects.filter(name='{service_name}').delete()"
+    )
+    execute_pulpcore_python(cli_client, cmd)
```
mdellweg, line 221: Did we capture this in a ticket, that there is no _nice_ way to remove a signing service?

lubosmj: No, I opened a ticket in pulpcore: https://github.com/pulp/pulpcore/issues/2967

PR #753, opened 2022-05-05
Allow comments in the requirements.txt file

ipanova:
[noissue]

### DIFF
--- a/.ci/scripts/upper_bound.py
+++ b/.ci/scripts/upper_bound.py
@@ -3,10 +3,11 @@ from pkg_resources import Requirement
 packages = []
 with open("requirements.txt", "r") as fd:
     for line in fd.readlines():
-        req = Requirement.parse(line)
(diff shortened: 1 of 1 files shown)


.ci/scripts/upper_bound.py
```diff
@@ -3,10 +3,11 @@
 packages = []
 with open("requirements.txt", "r") as fd:
     for line in fd.readlines():
-        req = Requirement.parse(line)
-        spec = str(req.specs)
-        if len(req.specs) < 2 and "~=" not in spec and "==" not in spec and "<" not in spec:
-            packages.append(req.name)
+        if not line.startswith("#"):
```
fao89, line 6: could you add it on plugin-template?

ipanova: i was planning too, i just used this PR as guinea pig ;)

PR #755, opened 2022-05-06
Fixed nightly job.

ipanova:
removed unused/outdated job.

[noissue]


fao89 (approved):

mdellweg:
Do we know, why this job failed? I do not think that it is outdated...

ipanova:
This job was introduced 2 year ago, other plugins did not seem to have used it, i disabled it here too.

PR #757, opened 2022-05-06
Fix the exponential evaluation of the expression

### LINKED ISSUE #756: Invalid regular expression is used for validating base paths
https://github.com/pulp/pulp_container/blob/4f8fa8fbbff6a1b783129d2d1be631c18ecc8dfc/pulp_container/app/serializers.py#L38

The security scanner marks the attached regex as prone to errors: `This part of the regular expression may cause exponential backtracking on strings containing many repetitions of '0'.` One passive group seems to be redundant.

lubosmj:
closes #756

### DIFF
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -35,7 +35,7 @@ from pulp_container.constants import SIGNATURE_TYPE
 
 VALID_SIGNATURE_NAME_REGEX = r"^sha256:[0-9a-f]{64}@[0-9a-f]{32}$"
 VALID_TAG_REGEX = r"^[A-Za-z0-9][A-Za-z0-9._-]*$"
-VALID_BASE_PATH_REGEX_COMPILED = re.compile(r"^[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?$")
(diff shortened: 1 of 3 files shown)


pulp_container/app/serializers.py
```diff
@@ -35,7 +35,7 @@
 
 VALID_SIGNATURE_NAME_REGEX = r"^sha256:[0-9a-f]{64}@[0-9a-f]{32}$"
 VALID_TAG_REGEX = r"^[A-Za-z0-9][A-Za-z0-9._-]*$"
-VALID_BASE_PATH_REGEX_COMPILED = re.compile(r"^[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?$")
+VALID_BASE_PATH_REGEX_COMPILED = re.compile(r"^[a-z0-9]+((?:[._]|__|[-]*)[a-z0-9])*$")
```
mdellweg, line 38: We could make the outer capture group anonymous too:
```suggestion
VALID_BASE_PATH_REGEX_COMPILED = re.compile(r"^[a-z0-9]+(?:(?:[._]|__|[-]*)[a-z0-9])*$")
```

Also possible, we could add the slashes here and not split later:
`r"^[a-z0-9]+(?:(?:[._]|__|[-]*)[a-z0-9])(?:/[a-z0-9]+(?:(?:[._]|__|[-]*)[a-z0-9]))*$"`

mdellweg (approved):
LGTM

PR #766, opened 2022-05-09
Fix S3 URL

fao89:
[noissue]
Ref: https://github.com/pulp/pulpcore/pull/2684


ipanova:
Thank you for opening a PR here too. I've opened an issue, would you mind adding a changelog entry https://github.com/pulp/pulp_container/issues/767?

ipanova (approved):

PR #777, opened 2022-05-12
Update CI

ipanova:
[noissue]


fao89:
if you don't change template-config vars, you can trigger this trhough UI:
image

ipanova:
@fao89 nice, i did not know that!

PR #788, opened 2022-05-16
ManifestSignature cannot be reclaimed.

ipanova:
ManifestSignature has no artifact.

[noissue]


mdellweg (approved):
Is this worth a changelog?

ipanova:
> Is this worth a changelog?

not sure, current state does not break anything, just adds unnecessary logic evaluation in core's reclaim task.

PR #798, opened 2022-05-25
Reassess the pyjwt dependency

lubosmj:
[noissue]

### DIFF
--- a/pulp_container/app/authorization.py
+++ b/pulp_container/app/authorization.py
@@ -82,13 +82,12 @@ class AuthorizationService:
         )
 
         with open(settings.PRIVATE_KEY_PATH, "rb") as private_key:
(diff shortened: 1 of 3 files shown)


lubosmj:
This PR should resolve the alerts reported by dependabot.

pulp_container/app/authorization.py
```diff
@@ -82,13 +82,12 @@ def generate_token(self):
         )
 
         with open(settings.PRIVATE_KEY_PATH, "rb") as private_key:
-            binary_token = jwt.encode(
```
ipanova, line 85: since you're touching the application code, does it make sense to create a misc entry with the explanation?

lubosmj: 🤷🏻‍♂️

PR #800, opened 2022-05-25
Update CI files from main branch

pulpbot:
[noissue]


template_config.yml
```diff
@@ -72,6 +72,7 @@ single_commit_check: true
 stalebot: false
 stalebot_days_until_close: 30
 stalebot_days_until_stale: 90
+stalebot_limit_to_pulls: true
```
fao89, line 75: we are limiting stalebot to PRs now, should we enable it now?
@ipanova @lubosmj

lubosmj: We can try it because low priority issues will always be marked as stale.

lubosmj (approved):

PR #807, opened 2022-05-25
Enable queryset_scoping from core.

ipanova:
[noissue]

### DIFF
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -8,8 +8,6 @@ import logging
 
 from gettext import gettext as _
 
-from collections import namedtuple
-
 from django.db import IntegrityError
 from django.db.models import Q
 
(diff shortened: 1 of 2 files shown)


pulp_container/app/viewsets.py
```diff
@@ -1082,8 +1108,19 @@ class ContainerPushRepositoryVersionViewSet(
...
+        "queryset_scoping": {
+            "function": "get_push_repos_qs",
+            "parameters": {
+                "ns_perm": "container.view_containernamespace",
+                "dist_perm": "container.view_containerdistribution",
+            },
+        },
     }
 
+    def get_push_repos_qs(self, qs, ns_perm, dist_perm):
```
ipanova, line 1120: @gerrod3 for the viewsets that have a parent viewset, does it make sense to just add a check here in core and just take the queryset_scoping automatically from the parent viewset https://github.com/pulp/pulpcore/blob/main/pulpcore/app/access_policy.py#L68 or you see a usecase where 'child viewset' would have it's own scoping?

ipanova: nevermind, i realized that queryset_scoping it is not even needed here.

mdellweg (approved):
This looks good to me. Also i think we have tests to back up this refactor.

PR #808, opened 2022-05-26
Clarify signing docs.

ipanova:
[noissue]

### DIFF
--- a/docs/workflows/sign-images.rst
+++ b/docs/workflows/sign-images.rst
@@ -19,7 +19,7 @@ The example below demonstrates how a manifest signing service can be created usi
 
        {"signature_path": "signature"}
 
-   Below is an example of a signing script:
+   Below is a snippet of the signing script to be used for signing content:
 
    .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/workflows/sign-images.rst
```diff
@@ -19,23 +19,23 @@ The example below demonstrates how a manifest signing service can be created usi
 
        {"signature_path": "signature"}
 
-   Below is an example of a signing script:
+   Below is a snippet of the signing script to be used for signing content:
```
lubosmj, line 22: If we state this, should not we also ship a signing script with the plugin?

docs/workflows/sign-images.rst
```diff
@@ -19,23 +19,23 @@ The example below demonstrates how a manifest signing service can be created usi
...
    .. code-block:: bash
 
        #!/usr/bin/env bash
 
         MANIFEST_PATH1=$1
         FINGEPRINT="$PULP_SIGNING_KEY_FINGERPRINT"
-        REFERENCE="$IMAGE_REFERENCE"
-        SIG_PATH="$SIGNATURE_PATH"
+        IMAGE_REFERENCE="$REFERENCE"
+        SIGNATURE_PATH="$SIG_PATH"
```
lubosmj, line 31: It might not be that straightforward for administrators to see where environment variables `$ REFERECE` and` $ SIG_PATH` come from. Are they set by them before running a task? (No). Maybe, it would be sufficient just to say that these variables are set inside a Pulp task.

lubosmj (approved):

PR #810, opened 2022-05-30
Improved image upload process

ipanova:
[noissue]
Required PR: https://github.com/pulp/pulpcore/pull/2779


pulp_container/app/registry_api.py
```diff
@@ -620,35 +622,56 @@ def partial_update(self, request, path, pk=None):
...
 
-            upload.append(chunk, upload.size)
-            upload.size += chunk.size
+            # if more chunks
+            if range_header:
+                upload.append(chunk, upload.size)
+            else:
+                # 1 chunk
+                # do not add to the upload, create artifact right away
+                import pydevd_pycharm
```
ipanova, line 648: leftovers

CHANGES/797.bugfix
```diff
@@ -0,0 +1,3 @@
+Improved image upload process from podman/docker clients.
```
ipanova, line 1: todo - check also skopeo

pulp_container/app/registry_api.py
```diff
@@ -547,11 +547,13 @@ class BlobUploads(ContainerRegistryApiMixin, ViewSet):
...
     queryset = models.Upload.objects.all()
 
-    content_range_pattern = re.compile(r"^(?P<start>\d+)-(?P<end>\d+)$")
+    content_range_pattern = re.compile(r"^([0-9]+)-([0-9]+)$")
 
     def create(self, request, path):
         """
         Create a new upload.
+
+        Note: We do not support monolithic upload.
```
mdellweg, line 556: Is "monolithic upload" and "all the file in one chunk" a different thing?

pulp_container/app/registry_api.py
```diff
@@ -620,35 +622,54 @@ def partial_update(self, request, path, pk=None):
...
+                with NamedTemporaryFile("ab") as temp_file:
+                    temp_file.write(chunk.read())
+                    temp_file.flush()
+
+                    uploaded_file = PulpTemporaryUploadedFile.from_file(
+                        File(open(temp_file.name, "rb"))
+                    )
+                try:
+                    artifact = Artifact.init_and_validate(uploaded_file)
+                    artifact.save()
```
mdellweg, line 657: This artifact we created here does not seem to be related to the upload we patched. So it is created as an orphan right? Don't we need to associate it with an new repo-version right away?

pulp_container/app/registry_api.py
```diff
@@ -684,54 +705,57 @@ def put(self, request, path, pk=None):
...
+            artifact.touch()
+        try:
+            blob = models.Blob(digest=digest)
+            blob.save()
+        except IntegrityError:
+            blob = models.Blob.objects.get(digest=digest)
+            blob.touch()
+        try:
+            blob_artifact = ContentArtifact(artifact=artifact, content=blob, relative_path=digest)
+            blob_artifact.save()
```
mdellweg, line 738: The Blob and it's content-artifact should be created in a transaction. So the Blob once released into existence will already have it's artifact attached.

pulp_container/app/registry_api.py
```diff
@@ -620,35 +622,54 @@ def partial_update(self, request, path, pk=None):
...
         """
         _, repository = self.get_dr_push(request, path)
+        upload = get_object_or_404(models.Upload, repository=repository, pk=pk)
         chunk = request.META["wsgi.input"]
-        if "Content-Range" in request.headers or "digest" not in request.query_params:
-            whole = False
-        else:
-            whole = True
+        if range_header := request.headers.get("Content-Range"):
+            found = self.content_range_pattern.match(range_header)
```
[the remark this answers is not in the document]
ipanova: the pattern was taken from quay, i can raise an error though

pulp_container/app/registry_api.py
ipanova: yes, monolitic upload uploads all the data during PUT request

pulp_container/app/registry_api.py
```diff
@@ -684,54 +705,57 @@ def put(self, request, path, pk=None):
...
-                    "add_content_units": [str(blob.pk)],
-                    "remove_content_units": [],
-                },
-            )
-
-            if has_task_completed(dispatched_task):
-                return BlobResponse(blob, path, 201, request)
         else:
-            raise Exception("The digest did not match")
+            artifact = Artifact.objects.get(sha256=digest[len("sha256:") :])
```
ipanova, line 728: catch does not exist

ipanova:
> I'm not sure i understand the change properly, but i have the feeling, that one would be able to "steal" an artifact here by claiming to upload a blob with a certain checksum, not showing any chunks to the server and then committing said upload object.

I would not be so worried about artifact stealing because we have many other windows when it can happen. For example in the  sync pipeline. Artifact can be stolen by the time it reaches Content stage.
This PR change becomes clear and simple once you grasp the docker api specs. Docker client uses it a weird way - there is monolithic upload and chunked upload. Monolithic upload starts with POST by creating upload and then proceeds with PUT where one big chunk would be uploaded.
Chunk upload starts with POST by creating upload and then proceeds  with PATCH where each PATCH contains a chunk with the range header. Then it finishes with PUT where it is allowed to whether submit last chunk or just empty body. This the signal of 'i have finished my chunked upload'.
Docker uses a hybrid of these 2 approaches. It uses chunked upload, but sends just one chunk and does not seem to use range header at all :)
Because of this hybrid approach we just perform extra actions like - create a chunk, save it to storage, retrieve all upload chunks( always one ) read it back from storage, assemble artifact from 1 chunk ( i.e. again write to storage)

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
mdellweg: The pattern is fine, but the client could be evil and pass anything in the header. Yes we should error in a controlled way.

mdellweg:
I do not think there is a way to steal artifacts by using uploads in the current implementation. (I agree, until we have all the necessary queryset-scoping on the artifacts the holes are wide open in other areas, but we intend to close them.)
Creating the blob when the big chunk arrives is probably not possible, because that is really the duty of that last PUT call. So i would try to keep the artifact attached to the upload with on_delete=PROTECT.
Another approach may be to see if we can rename the single chunk in the storage backend without moving data around to turn it into an artifact.

ipanova:
I don't understand your last comment. This PR does not address or solve the artifact stealing issues at all - it focuses on performance. The data is not read nor written twice. This is  important because even calculating the digest of the same data adds time. More importantly how object storage is involved here because users pay money for every write/read request to the bucket.

mdellweg:
> I don't understand your last comment. This PR does not address or solve the artifact stealing issues at all - it focuses on performance. The data is not read nor written twice. This is important because even calculating the digest of the same data adds time. More importantly how object storage is involved here because users pay money for every write/read request to the bucket.

What i wanted to say is that it introduces a new way to steal an artifact. One more we that need to think about later if we don't solve it right away.

pulp_container/app/models.py
```diff
@@ -633,3 +634,4 @@ class Upload(CoreUpload):
     """
 
     repository = models.ForeignKey(Repository, related_name="uploads", on_delete=models.CASCADE)
+    artifact = models.ForeignKey(Artifact, related_name="uploads", null=True, on_delete=models.PROTECT)
```
ipanova, line 637: @mdellweg I am hesitant to add `protect` here.
In case orphan clean up will run, task will fail with `ProtectedError: ("Cannot delete some instances of model 'Artifact' because they are referenced through protected foreign keys: 'Upload.artifact'.", {<Upload: pk=9a0acf29-0371-4f72-af3f-8f5c9667173e>})
`
In core were are checking the timestamp and the content membership so it would be identified as `orphaned` given that the timestamp will be small.

I think I am ok not to put protect here just because the time between artifact creation in patch call and put call this line https://github.com/pulp/pulp_container/pull/810/files#diff-1b944e15118f82233aa03a435e9e05c188ad0d371b71be43260506b4a5eeddffR709 would be few seconds. Obviously this is a window but it should be ok unless orphan_protection time is set to 0 or few seconds. WDYT?

mdellweg: We can still "touch" the artifact, right? Then this should be ok.

pulp_container/app/registry_api.py
```diff
@@ -620,35 +622,56 @@ def partial_update(self, request, path, pk=None):
...
 
-        chunk = ContentFile(chunk.read())
         with transaction.atomic():
             if upload.size != start:
                 raise Exception
 
-            upload.append(chunk, upload.size)
-            upload.size += chunk.size
+            # if more chunks
+            if range_header:
```
mdellweg, line 643: What if the range header is there, but it still covers all the blob. I think we can do the same trick. Also if an artifact is set here, i think we should reject further chunks.

ipanova: You mean there is a range header but only one chunk will be ever sent? I don't think this would be a valid case

ipanova: If you don't mind I would rather refrain from adding any additional logic for hypothetical cases. This is already not the most straightforward as is but at least there is a practical use of that code-path.

pulp_container/app/registry_api.py
```diff
@@ -620,35 +623,56 @@ def partial_update(self, request, path, pk=None):
...
+        upload = get_object_or_404(models.Upload, repository=repository, pk=pk)
         chunk = request.META["wsgi.input"]
-        if "Content-Range" in request.headers or "digest" not in request.query_params:
-            whole = False
-        else:
-            whole = True
+        if range_header := request.headers.get("Content-Range"):
+            found = self.content_range_pattern.match(range_header)
+            if not found:
+                raise InvalidRequest(message="Invalid range header")
```
mdellweg, line 631: :+1:

pulp_container/app/models.py
```diff
@@ -633,3 +634,6 @@ class Upload(CoreUpload):
     """
 
     repository = models.ForeignKey(Repository, related_name="uploads", on_delete=models.CASCADE)
+    artifact = models.ForeignKey(
+        Artifact, related_name="uploads", null=True, on_delete=models.SET_NULL
+    )
```
mdellweg, line 639: :+1:

pulp_container/app/registry_api.py
```diff
@@ -682,23 +706,30 @@ def put(self, request, path, pk=None):
...
-            for chunk in chunks:
-                temp_file.write(chunk.file.read())
-                chunk.file.close()
-            temp_file.flush()
-
-            uploaded_file = PulpTemporaryUploadedFile.from_file(File(open(temp_file.name, "rb")))
-
-        if uploaded_file.hashers["sha256"].hexdigest() == digest[len("sha256:") :]:
+        if artifact := upload.artifact:
+            artifact.touch()
```
mdellweg, line 710: I think in this case we also need to check the sha256 for this artifact against digest.

mdellweg (commented):
One last concern i found.

pulp_container/app/registry_api.py
ipanova: yes this a good check, let me add it.
should i raise same exception like in the codepath below? `raise Exception("The digest did not match")`?

mdellweg (approved):

PR #811, opened 2022-05-31
Refactor the RBAC documentation for the plugin

### LINKED ISSUE #641: Add docs that facilitate manual steps from permissions to roles
Pulp-container deployments that used to have custom permissions added outside of default groups that are created during podman push need manual documented steps how perform the switch from custom permisions to roles.

Pulp-container deployments that used default groups don't need to take any manul steps, there is a migration that transforms those grouped permissions into roles.

lubosmj:
closes #641

### DIFF
--- a/docs/rbac/index.rst
+++ b/docs/rbac/index.rst
@@ -1,0 +1,26 @@
+.. _rbac-index:
+
+Role-based Access Control
+=========================
+
+Role-based access control (RBAC) **restricts** access to entities based on a user's role within an
+organization. A role consists of one or more permissions. Users having a proper set of roles can
+view, modify, or delete resources hosted on different endpoints.
+
+By default, container repositories' content is accessible via ``podman`` or ``docker`` pull
+commands, unless the opposite is *explicitly* specified. A private repository can be created via the
+REST API for container distributions. An existing distribution can be updated with the parameter
+``private=True``.
+
+.. note::
+
+    Users logged in as administrators (staff) always bypass any authorization checks.
+
+Visit the following sections to understand how the plugin implements RBAC:
+
+.. toctree::
+   :maxdepth: 2
+
+   roles
+   permissions
+   migrating-perms-to-roles
--- a/docs/rbac/migrating-perms-to-roles.rst
+++ b/docs/rbac/migrating-perms-to-roles.rst
@@ -1,0 +1,46 @@
+.. _migrating-perms-to-roles:
+
+Migrating Permissions
+=====================
+
+As of release 2.11.0, the plugin started to support roles instead of separate groups and
+permissions. Default permission classes provided by Pulp are **automatically** migrated when
+upgrading from older releases. But, custom permissions created before release 2.11.0 require
+additional **post-upgrade steps** to preserve the initial behaviour.
+
+Usually, administrators define permissions for two types of operations:
+
+1. **pull** - Pulling content from all or a number of specific repositories
+2. **push** - Pushing content to all or concrete repositories
+
+During the upgrade, the custom permissions need to be manually revised and assigned. To do so, one
+can proceed as follows:
+
+1. Make all repositories private:
+
+.. code-block:: bash
+
+    for name in $(pulp container distribution list | jq -re '.[].name')
+    do
+        pulp container distribution update --name $name --private
+    done
+
+2. Start assigning Pulp-provided/adjusted roles to a particular user. For instance, use the role
+   ``container.containerdistribution_consumer`` to enable user ``alice`` to consume content from
+   distributions ``dist1``, ``dist2``, ``dist3``:
+
+.. code-block:: bash
+
+    for distribution in "dist1" "dist2" "dist3"
+    do
+        DISTRIBUTION_HREF=$(pulp container distribution show --name $distribution | jq -r ".pulp_href")
+        pulp user role-assignment add --username "alice" --role "container.containerdistribution_consumer" --object $DISTRIBUTION_HREF
+    done
+
+Similarly, execute an adjusted script for other repository objects that were asserted under
+the permissions' scope.
+
+.. note::
+
+    As of release 2.13.0, administrators should use the ``pulpcore-manager dump-permissions``
+    command to list deprecated permissions not yet translated into roles.
--- a/docs/rbac/permissions.rst
+++ b/docs/rbac/permissions.rst
@@ -1,25 +1,16 @@
-Role Based Access Control
-=========================
+.. _permissions:
 
-Role based access control in Pulp Container is configured using Access Policies for the following
-``viewset_names``:
+Permissions
+===========
 
-* ``pulp_container/namespaces``
-* ``distributions/container/container``
-* ``repositories/container/container-push``
-* ``remotes/container/container``
-* ``repositories/container/container``
-* ``repositories/container/container-push/versions``
-* ``repositories/container/container/versions``
-* ``content/container/blobs``
-* ``content/container/manifests``
-* ``content/container/tags``
+A role is defined by one or more permissions. In this section, details of permissions used within
+the container plugin are discussed.
 
-This document describes the default access policies shipped with Pulp Container. Each of the above
-policies can be modified to achieve a different RBAC behavior.
+.. warning::
 
-Repositories that are created using ``podman push`` or ``docker push`` are considered public and anyone
-can ``pull`` the images from them. See below about creating private repositories.
+    The concept of managing granular permissions is obsolete. As of release 2.11.0, the plugin uses
+    :ref:`roles` instead of separate permission classes. To migrate the customized permission
+    classes to roles, follow the instructions shown at :ref:`migrating-perms-to-roles`.
 
 Namespaces
 ----------
@@ -171,10 +162,6 @@ public, then anyone can ``pull`` from the repository associated with the Distrib
 
 Private Repositories
 --------------------
-
-A private repository can be created using Pulp's API for Container Distributions. A distribution
-can be created before pushing to the repository or an existing distribution can be updated with
-``private=True``.
 
 Users wishing to ``pull`` from a Container Distribution with ``private=True``
 will require the following object level permission on the Distribution::
--- a/docs/rbac/roles.rst
+++ b/docs/rbac/roles.rst
@@ -1,0 +1,285 @@
+.. _roles:
+
+Roles
+=====
+
+Role based access control (RBAC) is configured using access policies for the following endpoints:
+
+* ``pulp_container/namespaces``
+* ``distributions/container/container``
+* ``repositories/container/container-push``
+* ``remotes/container/container``
+* ``repositories/container/container``
+* ``repositories/container/container-push/versions``
+* ``repositories/container/container/versions``
+* ``content/container/blobs``
+* ``content/container/manifests``
+* ``content/container/tags``
+
+
+Default Roles
+-------------
+
+For each endpoint, a different set of roles is defined. The roles can be assigned at the model
+or object level for every user or group. In the following sections, the *Creator*, *Owner*,
+*Consumer*, and *Collaborator* roles are introduced. The *Consumer* and *Collaborator* roles are
+defined only for namespaces and distributions (i.e., container repositories served by the Pulp Registry).
+
+Creator Role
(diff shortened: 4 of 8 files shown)


docs/rbac/index.rst
```diff
@@ -0,0 +1,25 @@
+.. _rbac-index:
+
+Role-based Access Control
+=========================
+
+Role-based access control (RBAC) **restricts** access to content based on a user's role within an
```
mdellweg, line 6: Are you talking specifically about "content" here (i think not). Maybe use entities instead. In the context of Pulp, "content" already has quite a distinct meaning.

docs/rbac/index.rst
```diff
@@ -0,0 +1,25 @@
...
+organization. The role consists of one or more permissions. Users having a proper set of roles can
```
mdellweg, line 7:
```suggestion
organization. A role consists of one or more permissions. Users having a proper set of roles can
```

docs/rbac/index.rst
```diff
@@ -0,0 +1,25 @@
...
+view, modify, or delete content hosted on different endpoints.
+
+By default, all container repositories are accessible by anyone, unless the opposite is *explicitly*
```
mdellweg, line 10: I think you are referring to registry pull access here. Maybe something like `... repositories content is accessible via podman pull by anyone...`

docs/rbac/permissions.rst
```diff
@@ -1,25 +1,16 @@
...
-* ``repositories/container/container-push``
-* ``remotes/container/container``
-* ``repositories/container/container``
-* ``repositories/container/container-push/versions``
-* ``repositories/container/container/versions``
-* ``content/container/blobs``
-* ``content/container/manifests``
-* ``content/container/tags``
+A role is defined by one or more permissions. In this section, there are discussed details of
+permissions used within the container plugin.
```
mdellweg, line 7:
```suggestion
A role is defined by one or more permissions. In this section, details of
permissions used within the container plugin are discussed.
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+* ``repositories/container/container/versions``
+* ``content/container/blobs``
+* ``content/container/manifests``
+* ``content/container/tags``
+
+
+Default Roles
+-------------
+
+For each endpoint, there is defined a different set of roles. The roles can be assigned at the model
```
mdellweg, line 23:
```suggestion
For each endpoint, a different set of roles is defined. The roles can be assigned at the model
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+or object level for every user or group. In the following sections, there are introduced *Creator*,
+*Owner*, *Consumer*, and *Collaborator* roles. The *Consumer* and *Collaborator* roles are defined
```
mdellweg, line 25:
```suggestion
or object level for every user or group. In the following sections, the *Creator*,
*Owner*, *Consumer*, and *Collaborator* roles are introduced. The *Consumer* and *Collaborator* roles are defined
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+only for namespaces and distributions (i.e., container repositories served by the Pulp Registry).
+
+Creator Role
+~~~~~~~~~~~~
+
+The *Creator* role contains the ``add`` permission for objects present on a particular endpoint.
+For the endpoint which hosts distributions, only users with the ``container.add_containerdistribution``
```
mdellweg, line 32:
```suggestion
For the distribution endpoint, only users with the ``container.add_containerdistribution``
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+        "create"
+      ],
+      "effect": "allow",
+      "condition": "namespace_is_username",
+      "principal": "authenticated"
+    }
+
+.. note::
+
+    A user, who has the *Creator* role for namespaces assigned, does not need to have any additional
```
mdellweg, line 88:
```suggestion
    A user with the *Creator* role for namespaces does not need to have any additional
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+    roles to create distributions within the given namespaces. Similarly, the user is allowed to create
+    distributions within the owning username namespace (e.g., user ``alice`` can create container
+    repositories like ``alice/repo1``).
+
+
+Owner Role
+~~~~~~~~~~
+
+The *Owner* role contains all of the permissions available for ViewSet's objects but the
```
mdellweg, line 97:
```suggestion
The *Owner* role contains all of the permissions available for a ViewSet's objects but the
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+``add`` permission. For namespaces, the next set of permissions is considered:
```
mdellweg, line 98:
```suggestion
``add`` permission. For namespaces, the set of permissions reads:
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+
+    Repositories of the push type created with container clients (e.g., by running ``podman push``)
+    are considered public and anyone can ``pull`` the images from them.
+
+
+Consumer Role
+~~~~~~~~~~~~~
+
+The *Consumer* role contains only the ``view`` and ``pull`` permissions. Below is showcased a list
+of associated permissions for distributions:
```
mdellweg, line 147:
```suggestion
The *Consumer* role contains only the ``view`` and ``pull`` permissions. Below, a list
of associated permissions for distributions is showcased:
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+      "description": null,
+      "permissions": [
+        "container.pull_containerdistribution",
+        "container.view_containerdistribution"
+      ],
+      "locked": true
+    }
+
+Having the ``view`` and ``pull`` permissions allows a user to see and pull content from the Pulp
+Registry. Assigning this role only at the object level allows administrators to select what the
```
mdellweg, line 168:
```suggestion
Registry. Assigning this role only at the object level allows administrators and owners to select what the
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+        "content_object": "/pulp/api/v3/distributions/container/container/5b8ec13c-d578-4b3a-9b99-80986e5e00b6/"
+      }
+    ]
+
+
+Collaborator Role
+~~~~~~~~~~~~~~~~~
+
+The *Collaborator* role represents a set of permissions that a co-worker working within a same user-space
+should have. In comparison to the *Consumer* role, users with the *Collaborator* role are allowed
```
mdellweg, line 194:
```suggestion
should have. In addition to the *Consumer* role, users with the *Collaborator* role are allowed
```

docs/rbac/roles.rst
```diff
@@ -0,0 +1,275 @@
...
+.. code-block:: bash
+
+    pulp role create --name "container.containerrepository_syncer" \
+        --permission "container.view_containerrepository" \
+        --permission "container.view_containerremote" \
+        --permission "container.change_containerrepository" \
+        --permission "container.modify_content_containerrepository" \
+        --permission "container.sync_containerrepository"
+
+    pulp user role-assignment add --username "alice" --role "container.containerrepository_syncer" object ""
```
mdellweg, line 260: Should this be an example about changing the AP?

docs/tech-preview.rst
```diff
@@ -1,7 +1,7 @@
 Tech previews
 =============
 
-The following features are currently being released as part of a tech preview.
+The following features are currently being released as part of a tech preview:
```
mdellweg, line 4: :D :+1:

docs/rbac/roles.rst
lubosmj: I shall not use an indefinite article for plural forms.

docs/rbac/roles.rst
lubosmj: Of course. I provide an example of how to create, assign, and hook a role.

For the sake of completeness, I will add an example of assigning default permissions to a user in one of the descriptions of the default roles.

docs/rbac/roles.rst
lubosmj: I am referring to `ViewSet's objects`. And, `objects` are in a plural form.

ggainey: Hm. I believe that "a" does refer to "ViewSet" . If I use this alternate construction "...available to **the objects** of **a ViewSet**", does that make the difference clearer?

And as always, my apologies for English...

docs/rbac/roles.rst
mdellweg: I would still consider these two commands as an example to create and use a "custom role". "changing the access policy" to me is the next level of customization. My suggestion: Make this example its own thing, and move the headline for AP below it.

mdellweg: sounds good!

docs/rbac/roles.rst
```diff
@@ -0,0 +1,285 @@
...
+        "container.namespace_view_containerpushrepository",
+        "container.view_containernamespace"
+      ],
+      "locked": true
+    }
+
+Customizing Roles
+-----------------
+
+In Pulp, users are allowed to create or update their roles. To create a role with permissions
```
mdellweg, line 254: Maybe this is admins...

mdellweg (approved):
One nit-pick if you agree.

docs/rbac/index.rst
```diff
@@ -0,0 +1,26 @@
+.. _rbac-index:
+
+Role-based Access Control
+=========================
+
+Role-based access control (RBAC) **restricts** access to entities based on a user's role within an
+organization. A role consists of one or more permissions. Users having a proper set of roles can
+view, modify, or delete content hosted on different endpoints.
```
ipanova, line 8: s/content/resources. There is rbac not just for content.

docs/rbac/index.rst
```diff
@@ -0,0 +1,26 @@
...
+
```
ipanova, line 9: maybe worth adding a section called 'Content access'

docs/rbac/migrating-perms-to-roles.rst
```diff
@@ -0,0 +1,41 @@
+.. _migrating-perms-to-roles:
+
+Migrating Permissions
+=====================
+
+As of release 2.11.0, the plugin started to support roles instead of separate groups and
+permissions. Permission classes provided by Pulp are **automatically** migrated when upgrading
```
ipanova, line 7: Default permissions classes

docs/rbac/migrating-perms-to-roles.rst
```diff
@@ -0,0 +1,41 @@
...
+.. code-block:: bash
+
+    for distribution in "dist1" "dist2" "dist3"
+    do
+        DISTRIBUTION_HREF=$(pulp container distribution show --name $distribution | jq -r ".pulp_href")
+        pulp user role-assignment add --username "alice" --role "container.containerdistribution_consumer" --object $DISTRIBUTION_HREF
+    done
+
+Similarly, execute an adjusted script for other repository objects that were asserted under
+the permissions' scope.
```
ipanova, line 41: Can you mention the dump-permission management command Matthias has added?
Also please add a note that with pulp-container 2.13 this would be the only way to process unmigrated permissions because of removed django-guardian form the stack.

docs/rbac/index.rst
ipanova: can you also add information about the queryset scoping?

lubosmj: Not sure how the information about the queryset scoping will be useful to users. It tells plugin-writers that Pulp restricts access to objects based on users' permissions (which is generally known): https://github.com/pulp/pulpcore/blob/8ab3fdfba539b81fa25b67876dde7327992a6ce4/docs/plugins/plugin-writer/concepts/rbac/queryset_scoping.rst

docs/rbac/migrating-perms-to-roles.rst
lubosmj: I added a note about the version and the `dump-permission` command.

ipanova (approved):

PR #826, opened 2022-06-07
Create repositories automatically on import

### LINKED ISSUE #825: As a user, I want my imported repositories automatically created
Right now, there is a need to create an empty repository before proceeding with the import procedure. This will not be necessary once Pulp will automatically create missing repositories on-the-fly.

lubosmj:
Required PR: https://github.com/pulp/pulpcore/pull/2713
closes #825

### DIFF
--- a/CHANGES/825.feature
+++ b/CHANGES/825.feature
@@ -1,0 +1,3 @@
+Added support for automatically creating missing repositories during the import procedure. The
+creation is disabled by default. Use ``create_repositories=True`` to tell Pulp to create missing
+repositories when executing the import procedure.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -27,12 +27,12 @@ Features
 * :ref:`Build an OCI format image from a Containerfile <build-Containerfile-workflow>` and make it available from the Pulp Registry
 * Host content either `locally or on S3 <https://docs.pulpproject.org/installation/storage.html>`_
 * De-duplication of all saved content
-* Support disconnected and air-gapped environments with pulp import/export facility for synced container repositories
+* Support disconnected and air-gapped environments with the Pulp Import/Export facility for container repositories
 
 Tech Preview
 ------------
 
-Some additional features are being supplied as a tech preview.  There is a possibility that
+Some additional features are being supplied as a tech preview. There is a possibility that
 backwards incompatible changes will be introduced for these particular features.  For a list of
 features currently being supplied as tech previews only, see the :doc:`tech preview page
 <tech-preview>`.
--- a/docs/tech-preview.rst
+++ b/docs/tech-preview.rst
@@ -4,4 +4,4 @@ Tech previews
 The following features are currently being released as part of a tech preview:
 
 * Build an OCI image from a Containerfile
-* Import/Export with synced container repositories
+* Import/Export container repositories
--- a/docs/workflows/import-export.rst
+++ b/docs/workflows/import-export.rst
@@ -1,0 +1,42 @@
+Export and Import a Repository
+==============================
+
+When maintaining an **air-gapped** environment, one can benefit from using the import/export
+machinery. A common workflow usually resembles the following steps:
+
+#. **An administrator exports Pulp's content on a system with the internet connectivity.** The
+   system runs a Pulp instance that syncs content from remote repositories.
+#. **The exported content (tarball) is moved to another (air-gapped) system.** The transfer can
+   be made through the intranet or via an external hard drive.
+#. **The administrator imports the exported content by initiating an import task.** The
+   procedure takes care of importing the content to another Pulp instance running in the air-gapped
+   environment.
+
+Exporting a Repository
+----------------------
+
+To export a repository, run the following set of commands:
+
+.. literalinclude:: ../_scripts/export_repository.sh
+   :language: bash
+
+If the exported content is no longer needed to be managed on the system, delete it:
+
+.. literalinclude:: ../_scripts/cleanup_export.sh
+   :language: bash
+
+Importing the Repository
+------------------------
+
+Import the exported content by running the next commands and monitor the task:
+
+.. literalinclude:: ../_scripts/import_repository.sh
+   :language: bash
+
+.. note::
+    Pass ``create_repositories=True`` to the ``http POST ${BASE_ADDR}${IMPORTER_HREF}imports/``
+    request to tell Pulp to create missing repositories during the import procedure on the fly.
+    Otherwise, the repositories need to be created ahead of the import.
+
+.. warning::
+    Repositories of the push type are automatically converted to sync repositories at import time.
--- a/docs/workflows/index.rst
+++ b/docs/workflows/index.rst
@@ -45,8 +45,8 @@ set is the hostname and port::
    $ export BASE_ADDR=http://<hostname>:24817
 
 
-Container Workflows
--------------------
+Basic Workflows
+---------------
 
 .. toctree::
    :maxdepth: 2
@@ -58,5 +58,13 @@ Container Workflows
    build-containerfile
    managing-credentials
    push
+   import-export
+
+Managing Signatures
+-------------------
+
+.. toctree::
+   :maxdepth: 2
+
    sign-images
    verify-images
--- a/docs/workflows/manage-content.rst
+++ b/docs/workflows/manage-content.rst
@@ -1,7 +1,7 @@
 .. _content-management:
 
-Manage Container Content in a Repository
-========================================
+Manage Content in a Repository
+==============================
 
 There are multiple ways that users can manage Container content in repositories:
 
@@ -14,8 +14,8 @@ a new repository version will have been created.
 
 .. _tagging-workflow:
 
-Tagging
--------
+Tag an Image
+------------
 
 Images are described by manifests. The procedure of an image tagging is
 related to manifests because of that. In pulp, it is required to specify
@@ -44,8 +44,8 @@ Reference: `Container Tagging Usage <../restapi.html#tag/Content:-Tags>`_
 
 .. _untagging-workflow:
 
-Untagging
----------
+Untag an Image
+--------------
 
 An untagging is an inverse operation to the tagging. To remove a tag
 applied to an image, it is required to issue the following calls.
@@ -62,8 +62,8 @@ Reference: `Container Untagging Usage <../restapi.html#operation/repositories_co
 
 .. _recursive-add:
 
-Recursively Add Content to a Repository
----------------------------------------
+Add Content to a Repository
+---------------------------
 
 Any Container content can be added to a repository version with the
 recursive-add endpoint. Here, "recursive" means that the content will be
@@ -147,8 +147,8 @@ Reference: `Container Recursive Add Usage <../restapi.html#operation/repositorie
 
 .. _recursive-remove:
 
-Recursively Remove Content from a Repository
---------------------------------------------
+Remove Content from a Repository
+--------------------------------
 
 Any Container content can be removed from a repository version with the
 recursive-remove endpoint. Recursive remove is symmetrical with
@@ -214,8 +214,8 @@ Reference: `Container Recursive Remove Usage <../restapi.html#operation/reposito
 
 .. _tag-copy:
 
-Recursively Copy Tags from a Source Repository
-----------------------------------------------
+Copy Tags from a Repository
+---------------------------
 
 Tags in one repository can be copied to another repository using the tag
 copy endpoint.
@@ -284,8 +284,8 @@ Reference: `Container Copy Tags Usage <../restapi.html#operation/repositories_co
 
 .. _manifest-copy:
 
-Recursively Copy Manifests from a Source Repository
----------------------------------------------------
+Copy Manifests from a Repository
+--------------------------------
 
 Manifests in one repository can be copied to another repository using
 the manifest copy endpoint.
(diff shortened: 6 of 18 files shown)


lubosmj:
The proposed change can be easily tested by running the following snippet:
```
pulp container remote create --name lubosmj --url https://registry-1.docker.io --upstream-name lubosmj/mirror
SYNC_REPOSITORY=$(pulp container repository create --name lubosmj --remote lubosmj | jq -r ".pulp_href")
pulp container repository sync --name lubosmj

podman pull ghcr.io/pulp/test-fixture-1:manifest_d
podman tag ghcr.io/pulp/test-fixture-1:manifest_d localhost:24817/pulp/fixture:manifest_d
podman push localhost:24817/pulp/fixture:manifest_d --tls-verify=false
PUSH_REPOSITORY=$(pulp container distribution show --name "pulp/fixture" | jq -r ".repository")

SYNC_EXPORTER=$(http :24817/pulp/api/v3/exporters/core/pulp/ name=sync repositories:="[\"$SYNC_REPOSITORY\"]" path=/tmp/exports | jq -r ".pulp_href")
PUSH_EXPORTER=$(http :24817/pulp/api/v3/exporters/core/pulp/ name=push repositories:="[\"$PUSH_REPOSITORY\"]" path=/tmp/exports | jq -r ".pulp_href")
BOTH_EXPORTER=$(http :24817/pulp/api/v3/exporters/core/pulp/ name=both repositories:="[\"$PUSH_REPOSITORY\", \"$SYNC_REPOSITORY\"]" path=/tmp/exports | jq -r ".pulp_href")

http POST :24817${SYNC_EXPORTER}exports/
http POST :24817${PUSH_EXPORTER}exports/
http POST :24817${BOTH_EXPORTER}exports/
```

Then, you can try to import the exported content one by one to see how repositories are created on-the-fly.

CHANGES/825.feature
```diff
@@ -0,0 +1,3 @@
+Added support for automatically creating missing repositories during the import procedure. The
+creation is disabled by default. Use ``create_repositories=True`` to force Pulp to create
```
ipanova, line 2: s/to force/to tell

pulp_container/app/modelresource.py
```diff
@@ -1,7 +1,46 @@
 from import_export import fields, widgets
 from pulpcore.plugin.importexport import QueryModelResource, BaseContentResource
+from pulpcore.plugin.modelresources import RepositoryResource
 
-from pulp_container.app.models import Blob, Manifest, ManifestListManifest, ManifestSignature, Tag
+from pulp_container.app.models import (
+    Blob,
```
ipanova, line 6: alphabetical order, please

pulp_container/app/modelresource.py
```diff
@@ -1,7 +1,46 @@
...
+    ManifestSignature,
+    Tag,
+    ContainerRepository,
+    ContainerPushRepository,
+)
+
+
+class ContainerRepositoryResource(RepositoryResource):
+    """
+    A resource for importing/exporting repositories of the pull/sync type
```
ipanova, line 18: just sync type. both repo types are pullable

pulp_container/app/modelresource.py
```diff
@@ -1,7 +1,46 @@
...
+    """
+
+    def set_up_queryset(self):
+        """
+        :return: A queryset containing one sync repository that will be exported.
+        """
+        return ContainerRepository.objects.filter(pk=self.repo_version.repository)
```
ipanova, line 25: should not you also order_by the results?

pulp_container/app/modelresource.py
```diff
@@ -1,7 +1,46 @@
...
+
+    class Meta:
+        model = ContainerRepository
```
ipanova, line 28: what about import_id_fields?

pulp_container/app/modelresource.py
```diff
@@ -128,4 +167,6 @@ class Meta:
     ManifestListManifestResource,
     ManifestSignatureResource,
     TagResource,
+    ContainerRepositoryResource,
```
ipanova, line 170: asked a question on the core PR about the order

pulp_container/app/modelresource.py
ipanova: exclude fields too?

ipanova:
please update docs

ipanova:
@mdellweg @lubosmj  not sure what to do with push repo type. We claim that push repo cannot exist on its own and that it is tightly coupled with the distributions. We even defer all the push repo rbac to the distribution/namespace https://github.com/pulp/pulp_container/blob/main/pulp_container/app/viewsets.py#L965
Now the question on the day - should also distribution and namespace be created atomically with the push repo during import operation?
As of today the import/export is expected to be performed only by admins, so then later on they can normally create needed missing objects to expose the repo for consumption but in case in the future we'd allow this to non-admin users(?) they won't be able to even retrieve the imported repo.

mdellweg:
> @mdellweg @lubosmj not sure what to do with push repo type. We claim that push repo cannot exist on its own and that it is tightly coupled with the distributions. We even defer all the push repo rbac to the distribution/namespace https://github.com/pulp/pulp_container/blob/main/pulp_container/app/viewsets.py#L965 Now the question on the day - should also distribution and namespace be created atomically with the push repo during import operation? As of today the import/export is expected to be performed only by admins, so then later on they can normally create needed missing objects to expose the repo for consumption but in case in the future we'd allow this to non-admin users(?) they won't be able to even retrieve the imported repo.

Can we export push-repos as sync-repos, so one can export the content and make it available over the airgap without creating Push-repos there?

ipanova:
> > @mdellweg @lubosmj not sure what to do with push repo type. We claim that push repo cannot exist on its own and that it is tightly coupled with the distributions. We even defer all the push repo rbac to the distribution/namespace https://github.com/pulp/pulp_container/blob/main/pulp_container/app/viewsets.py#L965 Now the question on the day - should also distribution and namespace be created atomically with the push repo during import operation? As of today the import/export is expected to be performed only by admins, so then later on they can normally create needed missing objects to expose the repo for consumption but in case in the future we'd allow this to non-admin users(?) they won't be able to even retrieve the imported repo.
> 
> Can we export push-repos as sync-repos, so one can export the content and make it available over the airgap without creating Push-repos there?

Yes, we could and this was one the ideas on how to circumvent this sort of issues, however we still did not get any clear requirements and expectations from our stakeholders. See https://docs.google.com/document/d/1q8kTEOgZyXUmItRl9uXWWQ8Sonz2Iz0dtyjBe4ZBzqU/edit?disco=AAAAMxpC4TU

Given that the env is air gaped, to me personally, it makes sense to not allow income of the content in any other way than import/export, so having push repos converted into sync repo on the import plays well into this thinking. However maybe others might have a different opinion on that.
@rochacbruno what's your thinking on this?

rochacbruno:
@ipanova I agree with having it imported as a sync repo (read only) I don't see the requirement for having it push-able nowhere in the docs.

pulp_container/app/modelresource.py
lubosmj: `exclude_fields` is not required to declare. Same for `import_id_fields` ("The default field for object identification is id, you can optionally set which fields are used as the id when importing").

lubosmj: Actually, I should exclude "manifest_signing_service" from the fields. Does it make sense to include signing services in the export? I believe, it does not.

ipanova (commented):
@lubosmj thank you for the PR, it looks good.
Docs needs an update. https://docs.pulpproject.org/pulp_container/tech-preview.html and https://docs.pulpproject.org/pulp_container/index.html#features
I would be good to mention somewhere in the user facing docs, that push repo type will be converted to sync repo type in the import/export process.
I did not find any existing import/export workflows https://docs.pulpproject.org/pulp_container/workflows/index.html#container-workflows if you can document that, it would be great. Our long term goal is to have self contained docs which do not refer to pulpcore docs.

pulp_container/app/modelresource.py
```diff
@@ -1,7 +1,49 @@
...
+
+    def set_up_queryset(self):
+        """
+        :return: A queryset containing one push repository that will be exported.
+        """
+        return ContainerPushRepository.objects.filter(pk=self.repo_version.repository)
+
+    class Meta:
+        # import the repository as a repository of the sync type
+        model = ContainerRepository
```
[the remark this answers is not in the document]
lubosmj: This needs to be reassessed because there might be a conflict between the types that import-export framework is trying to create. In particular, the framework tries to create a sync-repository even though an initial push-repository could still exist during the import procedure. As a result, the import fails on a duplicated key error (e.g., repository "foo" already exists)...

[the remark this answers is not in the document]
mdellweg: And i think it's safe to fail in such circumstances. Just assuming the user may have wanted something they didn't ask for is way to dangerous.

ipanova (approved):
Thank you for adding the docs.

I generally try to avoid any unnecessary changes to the issue and create a separate PR to not create a lot of noise. In this case I am specifically referring to a lot of unrelated docs style changes.

docs/index.rst
```diff
@@ -27,7 +27,7 @@ Features
 * :ref:`Build an OCI format image from a Containerfile <build-Containerfile-workflow>` and make it available from the Pulp Registry
 * Host content either `locally or on S3 <https://docs.pulpproject.org/installation/storage.html>`_
 * De-duplication of all saved content
-* Support disconnected and air-gapped environments with pulp import/export facility for synced container repositories
+* Support disconnected and air-gapped environments with the Pulp import/export facility for container repositories
```
mdellweg, line 30:
```suggestion
* Support disconnected and air-gapped environments with the Pulp Import/Export facility for container repositories
```
If we make PIE a "thing" this starts to be a proper noun.

docs/tech-preview.rst
```diff
@@ -1,7 +1,7 @@
 Tech previews
 =============
 
-The following features are currently being released as part of a tech preview:
+The following features are currently being released as a part of a tech preview:
```
mdellweg, line 4: This was correct before...

docs/tech-preview.rst
```diff
@@ -1,7 +1,7 @@
...
 
-* Build an OCI image from a Containerfile
-* Import/Export with synced container repositories
+* Support for building an OCI image from a Containerfile
+* Support for importing/exporting container repositories
```
mdellweg, line 7: I don't think the "feature" is "supporting something".

docs/workflows/import-export.rst
```diff
@@ -0,0 +1,42 @@
+Export and Import a Repository
+==============================
+
+When maintaining an **air-gapped** environment, one can benefit from using the import/export
+machinery. The following workflow is usually considered:
```
mdellweg, line 5:
```suggestion
machinery. The common workflow usually resembles the following steps:
```

docs/workflows/import-export.rst
```diff
@@ -0,0 +1,42 @@
...
+.. literalinclude:: ../_scripts/import_repository.sh
+   :language: bash
+
+.. note::
+    Pass ``create_repositories=True`` to the ``http POST ${BASE_ADDR}${IMPORTER_HREF}imports/``
+    request to tell Pulp to create missing repositories during the import procedure on the fly.
+    Otherwise, the repositories need to be created ahead of the import.
+
+.. warning::
+    Repositories of the push type are automatically converted to sync repositories after an import.
```
mdellweg, line 42:
```suggestion
    Repositories of the push type are automatically converted to sync repositories at import time.
```

docs/workflows/index.rst
```diff
@@ -45,8 +45,8 @@ set is the hostname and port::
    $ export BASE_ADDR=http://<hostname>:24817
 
 
-Container Workflows
--------------------
+Standard Workflows
```
mdellweg, line 48: Maybe basic or common workflows.

docs/workflows/manage-content.rst
```diff
@@ -1,7 +1,7 @@
 .. _content-management:
 
-Manage Container Content in a Repository
-========================================
+Manage Content in a Repository
+==============================
```
mdellweg, line 4: :shrug:

docs/workflows/sign-images.rst
```diff
@@ -1,7 +1,7 @@
 .. _sign-images:
 
-Image signature configuration
-==============================
+Set Image Signature Configration
```
mdellweg, line 3: Either "setup" or just the title case change please.

docs/workflows/sync.rst
```diff
@@ -81,8 +81,8 @@ Reference: `Container Remote Usage <../restapi.html#tag/Remotes:-Container>`_
 
 .. _sync-repository:
 
-Sync repository using a Remote
-------------------------------
+Sync the Repository
+-------------------
```
mdellweg, line 85: "Sync Content into a Repository" maybe.

mdellweg (commented):
Only looking at the docs changes: Left some comments.

docs/workflows/manage-content.rst
lubosmj: The whole point of these changes was to shrink the names of chapters to make them more pleasant for users' eyes. Why would a user need to see again that we are dealing with Container content if it is known from the context?

docs/tech-preview.rst
lubosmj: No. https://www.oxfordlearnersdictionaries.com/definition/english/part_1?q=part

mdellweg (approved):

PR #828, opened 2022-06-10
Add plugin package name to the app config

ipanova:
[noissue]
RequiredPR: https://github.com/pulp/pulpcore/pull/2823

### DIFF
--- a/pulp_container/app/__init__.py
+++ b/pulp_container/app/__init__.py
@@ -7,3 +7,4 @@ class PulpContainerPluginAppConfig(PulpPluginAppConfig):
     name = "pulp_container.app"
     label = "container"
     version = "2.13.0.dev"
+    python_package_name = "pulp-container"


bmbouter (approved):
Thanks!

pulp_container/app/__init__.py
```diff
@@ -7,3 +7,4 @@ class PulpContainerPluginAppConfig(PulpPluginAppConfig):
     name = "pulp_container.app"
     label = "container"
     version = "2.13.0.dev"
+    python_package_name = "pulp_container"
```
gerrod3, line 10: Is this correct? Are we suppose to use the underscore or the dash version of the name. In `setup.py` the name has a dash.

bmbouter: Dang this is the one I didn't look up. It is with the `-` https://pypi.org/project/pulp-container/

bmbouter (changes requested):
@ipanova it needs the dash

ipanova:
> @ipanova it needs the dash

oops

PR #841, opened 2022-06-15
Improve manist_list push test.

ipanova:
[noissue]


lubosmj:
After applying your commit, we will no longer compare the pulled manifest list (the original one retrieved from an external registry) with the pushed manifest list. We are comparing the digests referenced by the very same manifest list in api-app and content-app. Is this what we really want?

lubosmj:
We need to ensure in some way that what we have pushed is correctly represented in the Pulp Registry. At least a check for the right number of manifests in the manifest list should suffice.

lubosmj (approved):

PR #872, opened 2022-06-24
Remove pulled images from the local storage

lubosmj:
[noissue]


lubosmj:
As I am thinking about this more and more, do we want to do this? It could remove images pulled intentionally by developers and they do not want to remove them. @ipanova

ipanova:
@lubosmj it could remove you're right. let's close this issue.

PR #885, opened 2022-07-01
Update the template to prevent failures

lubosmj:
[noissue]


mdellweg (approved):

ipanova:
Instead of manually opening the PR you can just trigger this job https://github.com/pulp/pulp_container/actions/workflows/update_ci.yml

PR #887, opened 2022-07-03
Update CI files from 2.7 branch

pulpbot:
[noissue]

PR #896, opened 2022-07-04
Update the plugin template

lubosmj:
[noissue]


lubosmj:
This PR was opened to resolve the errors shown in https://github.com/pulp/pulp_container/pull/887. The main difference is that I replaced the HTTP schema with HTTPS. There is still a failure in the docs check.

ipanova (approved):

PR #976, opened 2022-08-17
Fixed 404 during sync from registry.redhat.io

### LINKED ISSUE #974: Repo sync from RH registry fails with 404
**Version**
main but also older branches

**To Reproduce**

Steps to Reproduce:
1. create a container repo
2. create a container remote that would be syncing content from RH registry, for example rhosp-rhel8/openstack-aodh-api
3. Hit the sync button
4. And the task fails with 404 - Not found

Sync fails with:

```
404, message='Not Found', url=URL('https://registry.redhat.io/containers/content/dist/containers/rhel8/multiarch/containers/redhat-rhosp-rhel8-openstack-aodh-api/manifests/1/sha256:2881fb630130572a95d370ba801fec3340d6738c001abcab8d70bfc063ad9cd2?_auth_=exp=1659560550~hmac=caae5ac4009b16a60c419da47fca4713362f5a3c24f6f6d6f288476a31f96161')
```

**Additional context**
https://bugzilla.redhat.com/show_bug.cgi?id=2115970

ipanova:
Whenever a token expires a new token is requested.
    The 404 sync problem was triggered because the accept headers that
    were passed originally with the expired token were not passed to the
    re-newed token. As a result the registry was redirecting to the wrong
    schema version (schema1, which is unsupported since july 2022
    https://access.redhat.com/articles/6138332) because of absense of the
    accept headers.

closes #974

### DIFF
--- a/CHANGES/974.bugfix
+++ b/CHANGES/974.bugfix
@@ -1,0 +1,1 @@
+Fixed an HTTP 404 response during sync from registry.redhat.io.
--- a/pulp_container/app/downloaders.py
+++ b/pulp_container/app/downloaders.py
@@ -78,12 +78,12 @@ class RegistryAuthHttpDownloader(HttpDownloader):
 
                             self.registry_auth["bearer"] = None
                             await self.update_token(response_auth_header, this_token, repo_name)
-                        return await self._run(handle_401=False)
+                        return await self._run(handle_401=False, extra_data=extra_data)
                     elif "Basic" in response_auth_header:
                         if self.remote.username:
                             basic = aiohttp.BasicAuth(self.remote.username, self.remote.password)
                             self.registry_auth["basic"] = basic.encode()
-                        return await self._run(handle_401=False)
+                        return await self._run(handle_401=False, extra_data=extra_data)
                 else:
                     raise
             to_return = await self._handle_response(response)


ipanova:
FYI I still cannot create a local reproducer because it is not that easy to hit the point when the token expires during the image manifest request by digest https://access.redhat.com/articles/6138332

CHANGES/974.bugfix
```diff
@@ -0,0 +1 @@
+Fixed 404 during sync from registry.redhat.io
```
lubosmj, line 1:
```suggestion
Fixed an HTTP 404 response during sync from registry.redhat.io.
```

ipanova:
@ianballou what's the earliest pulp-container version you want this backported to ?

lubosmj (approved):

dkliban (approved):

ianballou:
@ipanova the very earliest would be 2.8

ipanova:
> FYI I still cannot create a local reproducer because it is not that easy to hit the point when the token expires during the image manifest request by digest https://access.redhat.com/articles/6138332

to help with the reproducer one can set `TOKEN_EXPIRATION_TIME` to a lower value than the default 300s

PR #1003, opened 2022-08-22
Add a repair command for fixing media types of v1 manifests

### LINKED ISSUE #977: Issue 883 persists for the existing content
As a follow up of this issue we need to repair data already present in the db https://github.com/pulp/pulp_container/issues/883
Current fix does not affect content which is already stored in the db, only newly synced one. We'd whether need to write a migration which will be fixing the DB records or adding a pulpcore-manager repair command specifically for this issue. This will be expensive because most likely we would need to read and parse manifest.json file from the disk.

Migration cannot be backported but repair command if it is considered as bugfix then can be.

https://issues.redhat.com/browse/AAH-1871

lubosmj:
closes #977

### DIFF
--- a/CHANGES/977.bugfix
+++ b/CHANGES/977.bugfix
@@ -1,0 +1,2 @@
+Introduced the ``pulpcore-manager container-repair-media-type`` command to fix incorrect media
+types of manifests that could have been stored in the database as a result of a sync task.
--- a/pulp_container/app/management/commands/container-repair-media-type.py
+++ b/pulp_container/app/management/commands/container-repair-media-type.py
@@ -1,0 +1,60 @@
+import json
+
+from gettext import gettext as _
+
+from django.conf import settings
+from django.core.management import BaseCommand
+
+from pulpcore.plugin.cache import SyncContentCache
+
+from pulp_container.app.models import Manifest, ContainerDistribution
+from pulp_container.app.utils import determine_media_type_from_json
+
+from pulp_container.constants import MEDIA_TYPE
+
+
+class Command(BaseCommand):
+    """
+    A django management command to repair the media types of manifests.
+
+    Older versions of pulp_container could sometimes assign an invalid media type to a manifest.
+    If the media type could not be extracted from the Content-Type header, the sync pipeline
+    assumed that the media type is "application/vnd.docker.distribution.manifest.v1+json". The
+    repair command iterates over synced manifests and updates their media types based on the
+    internals of the associated manifest.json files if needed.
+
+    This command also deletes the cache for all distributions across the plugin.
+    """
+
+    help = _(__doc__)
+
+    def handle(self, *args, **options):
+        """Run the management command."""
+        manifests_schema_v1 = Manifest.objects.filter(
+            media_type=MEDIA_TYPE.MANIFEST_V1
+        ).prefetch_related("_artifacts")
+
+        manifests_to_update = []
+        for manifest in manifests_schema_v1:
+            artifact_file = manifest._artifacts.first().file
+            json_data = json.load(artifact_file)
+            artifact_file.close()
+
+            media_type = determine_media_type_from_json(json_data)
+            if media_type != MEDIA_TYPE.MANIFEST_V1:
+                manifest.media_type = media_type
+                manifests_to_update.append(manifest)
+
+        if manifests_to_update:
+            Manifest.objects.bulk_update(manifests_to_update, ["media_type"], batch_size=100)
+
+        self.stdout.write(
+            self.style.SUCCESS("Successfully repaired %d manifests." % len(manifests_to_update))
+        )
+
+        if settings.CACHE_ENABLED:
+            base_paths = ContainerDistribution.objects.values_list("base_path", flat=True)
+            if base_paths:
+                SyncContentCache().delete(base_key=base_paths)
+
+            self.stdout.write(self.style.SUCCESS("Successfully flushed the cache."))
--- a/pulp_container/app/migrations/0033_raise_warning_for_repair.py
+++ b/pulp_container/app/migrations/0033_raise_warning_for_repair.py
@@ -1,0 +1,28 @@
+import warnings
+
+from django.db import migrations
+
+from pulp_container.constants import MEDIA_TYPE
+
+
+def print_warning_for_repair(apps, schema_editor):
+    Manifest = apps.get_model("container", "Manifest")
+    if Manifest.objects.filter(media_type=MEDIA_TYPE.MANIFEST_V1).exists():
+        warnings.warn(
+            "Manifests with potentially invalid media types were detected. Please, run the "
+            "'pulpcore-manager container-repair-media-type' command to repair the media types "
+            "of the manifests that could be incorrectly parsed during the sync procedure."
+        )
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [('container', '0032_upload_artifact')]
+
+    operations = [
+        migrations.RunPython(
+            code=print_warning_for_repair,
+            reverse_code=migrations.RunPython.noop,
+            elidable=True,
+        ),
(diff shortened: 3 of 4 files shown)


pulp_container/app/utils.py
```diff
@@ -144,9 +144,14 @@ def has_task_completed(dispatched_task):
 
 def determine_media_type(content_data, response):
     """Determine the media type of a manifest either from the JSON data or the response object."""
-    if media_type := content_data.get("mediaType"):
+    if media_type := response.headers.get("content-type"):
```
ipanova, line 147: Can you keep the logic as it was before? I.e the ultimate source of truth is `content_data.get("mediaType")` if available, then `response.headers.get("content-type")` and as last you continue parsing the json file.

ipanova:
@newswangerd to address this issue we're introducing the repair command that can be backported to older releases. There is no precedence for a pulpcore-manager repair command in pulp_container nor pulp_ansible plugins. How do you plan to use it? You would whether need to invoke it somewhere so it runs or document this somewhere so the user/customer takes an extra/manual step.

newswangerd:
I'm not sure. 

@simaishi is this something that we could add as an extra step in the installer? Alternatively, we could just tell customers to run the command when they run into this issue.

CHANGES/977.bugfix
```diff
@@ -0,0 +1,4 @@
+Introduced a repair command for fixing the media types of manifests that could be incorrectly
```
ipanova, line 1: Introduced a repair command to fix incorrectly stored manifest media_type in the db as a result of sync task.

CHANGES/977.bugfix
```diff
@@ -0,0 +1,4 @@
...
+determined during the sync procedure when running the following Y releases of pulp_container:
+2.13.1, 2.10.5, 2.9.6. Run the command ``pulpcore-manager container-repair-media-type`` to
```
ipanova, line 3: i would omit the versions, this issue is there for quite some time.

pulp_container/app/migrations/0033_raise_warning_for_repair.py
```diff
@@ -0,0 +1,21 @@
+import warnings
+
+from django.db import migrations
+
+
+def print_warning_for_repair(apps, schema_editor):
+    warnings.warn(
+        "Please, run the 'pulpcore-manager container-repair-media-type' command to repair "
```
ipanova, line 8: for possible future needs can you make it  follow similar style as in pulp-rpm? https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/app/management/commands/rpm-datarepair.py#L28
you would have `pulpcore-manager container-datarepair --issue 1234`

ipanova (commented):
The repair command looks good to me.
@mdellweg I want to hear from you on whether you're ok having a migration that raises a warning message or you would want to have schedulded task or something else.

pulp_container/app/migrations/0033_raise_warning_for_repair.py
```diff
@@ -0,0 +1,21 @@
+import warnings
+
+from django.db import migrations
+
+
+def print_warning_for_repair(apps, schema_editor):
```
ipanova, line 6: before raising warning you should first query the DB and check for the presence of the schema v1 manifests.

pulp_container/app/migrations/0033_raise_warning_for_repair.py
lubosmj: And what should be the issue number in this case? 997? I would opt not to specify any issue numbers. Is it user-friendly? Also, in this case, there is no issue number: https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/app/management/commands/rpm-trim-changelogs.py.

ipanova: yes, 977.

ipanova: you don't want to have 10 management data repair commands in case you have 10 issues. You want to have one interface.

ipanova: if you really want to keep the name as is i am ok with that too, hopefully we won't have many cases of data repair but in case we do then we'd need to write a new command again

CHANGES/977.bugfix
mdellweg: "... that could have been ..." The moment this code is installed it should no longer be possible.

CHANGES/977.bugfix
mdellweg: Agree, also i would expect a version range on every branch. Better to just skip it.

pulp_container/app/migrations/0033_raise_warning_for_repair.py
```diff
@@ -0,0 +1,21 @@
...
+        "more at https://docs.pulpproject.org/pulp_container/changes.html (#883)."
+    )
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [('container', '0032_upload_artifact')]
+
+    operations = [
+        migrations.RunPython(print_warning_for_repair),
```
mdellweg, line 20:
```suggestion
        migrations.RunPython(code=print_warning_for_repair, reverse_code=migrations.RunPython.noop, elidable=True),
```

lubosmj:
Now, admins are not required to flush the cache manually. Everything is done for them in the repair command automatically.

pulp_container/app/management/commands/container-repair-media-type.py
```diff
@@ -0,0 +1,60 @@
...
+            json_data = json.load(artifact_file)
+            artifact_file.close()
+
+            media_type = determine_media_type_from_json(json_data)
+            if media_type != MEDIA_TYPE.MANIFEST_V1:
+                manifest.media_type = media_type
+                manifests_to_update.append(manifest)
+
+        if manifests_to_update:
+            Manifest.objects.bulk_update(manifests_to_update, ["media_type"], batch_size=100)
```
mdellweg, line 49: What i usually do if i expect the list `manifests_to_update` to blow up in RAM is to issue a bulk_update inside the for loop once the list exceeded a certain size followed by a `manifests_to_update.clear()`. But i don't see this required right now.

mdellweg (approved):

pulp_container/app/management/commands/container-repair-media-type.py
ipanova: I think usually we set batch size to 1000

pulp_container/app/management/commands/container-repair-media-type.py
```diff
@@ -0,0 +1,60 @@
...
+
+        self.stdout.write(
+            self.style.SUCCESS("Successfully repaired %d manifests." % len(manifests_to_update))
+        )
+
+        if settings.CACHE_ENABLED:
+            base_paths = ContainerDistribution.objects.values_list("base_path", flat=True)
+            if base_paths:
+                SyncContentCache().delete(base_key=base_paths)
```
ipanova, line 58: aren't you supposed to also invalidate async-content-cache which us responsible for the served content?

ipanova: calling delete() whether on sync or async cache invalidates all cache.

ipanova (approved):

pulp_container/app/management/commands/container-repair-media-type.py
mdellweg: Yes, i think it's two interfaces to the same cache.

simaishi:
> I'm not sure.
> 
> @simaishi is this something that we could add as an extra step in the installer? Alternatively, we could just tell customers to run the command when they run into this issue.

I think it will be better if this can be added to pulp_installer. I'll follow up and track in the AAH ticket.

PR #1007, opened 2022-08-24
Add a repair command for fixing media types of v1 manifests

### LINKED ISSUE #977: Issue 883 persists for the existing content

lubosmj:
closes #977

(cherry picked from commit 2a9c5e6a51a4e8aed19300f6cefc3601d5656af7)

### DIFF
--- a/CHANGES/977.bugfix
+++ b/CHANGES/977.bugfix
@@ -1,0 +1,2 @@
+Introduced the ``pulpcore-manager container-repair-media-type`` command to fix incorrect media
+types of manifests that could have been stored in the database as a result of a sync task.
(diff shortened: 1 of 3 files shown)


lubosmj:
The only difference in comparison to the backport to 2.13 should be the removed caching utility.

ipanova (approved):

PR #1022, opened 2022-08-25
Copy the GnuPG fixture from pulpcore

lubosmj:
[noissue]


mdellweg (approved):
LGTM Let's hope it fixes the release process.

PR #1094, opened 2022-10-03
Updated deps accoring to the docs guidelines.

### LINKED ISSUE #1093: Add upper and lower bounds to all pulp_container dependencies #3232
Per [the new dependency docs](https://github.com/pulp/pulpcore/blob/main/docs/plugins/plugin-writer/concepts/index.rst#declaring-dependencies) we should add an upper and lower bound for all dependencies.

This does not need to be done for the functional and test dependencies, just the runtime pulp_container dependencies.

ipanova:
closes #1093

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  commit-message:
+    prefix: "[noissue]"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # remember to also update unittest_requirements.txt when updating this file
+ecdsa>=0.14,<=0.18.0
+jsonschema>=4.4,<4.18
 pulpcore>=3.20.0.dev,<3.25
-ecdsa~=0.14
-pyjwkest~=1.4
-pyjwt[crypto]~=2.4
-jsonschema~=4.4
+pyjwkest>=1.4,<=1.4.2
+pyjwt[crypto]>=2.4,<=2.5
--- a/CHANGES/1093.misc
+++ b/CHANGES/1093.misc
@@ -1,0 +1,1 @@
+Adds upper and lower bound to dependencies. Enable dependabot.


ipanova:
@bmbouter is this what you've expected from a plugin adhering to the new guidelines?

requirements.txt
```diff
@@ -1,6 +1,16 @@
 # remember to also update unittest_requirements.txt when updating this file
+aiohttp>=3.8.1,<=3.8.3
+cryptography>=38.0.1<=38.0.1
+Django~=3.2.15  # LTS version, switch only if we have a compelling reason to
+djangorestframework>=3.13.1,<=3.14.0
+django-filter>=22.1,<=22.1
+django-import-export>=2.9,<=2.9.0
+drf-access-policy>=1.1.2,<=1.1.2
+drf-spectacular==0.24.2  # We monkeypatch this so we need a very narrow requirement string
```
bmbouter, line 9: Do you directly import this?

Anything pulp_container doesn't directly import doesn't need to be declared. Perhaps you do directly import this though....

bmbouter:
> @bmbouter is this what you've expected from a plugin adhering to the new guidelines?

I left a question, but style-wise yes this is exactly what I was expecting.

requirements.txt
ipanova: yeah we do
```
[ipanova@puffy pulp_container]$ git grep spectac
.ci/scripts/schema.py:This script modifies drf-spectacular schema validation to accept slashes and curly brackets.
.ci/scripts/schema.py:from drf_spectacular.validation import JSON_SCHEMA_SPEC_PATH
pulp_container/app/viewsets.py:from drf_spectacular.utils import extend_schema
[ipanova@puffy pulp_container]$ git grep extend_schema
pulp_container/app/viewsets.py:from drf_spectacular.utils import extend_schema
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
pulp_container/app/viewsets.py:    @extend_schema(
[ipanova@puffy pulp_container]$ 
```

ipanova:
> > @bmbouter is this what you've expected from a plugin adhering to the new guidelines?
> 
> I left a question, but style-wise yes this is exactly what I was expecting.

ty for proofreading

requirements.txt
mdellweg: I would expect us to have a _very_ broad range here. Pinning to the same version as pulpcore is calling for desaster. Also i think the monkey-patching does not apply here.

bmbouter: I agree, I think this can be very broad and it needs to be or this strategy won't work well.

bmbouter: As an aside, we need pulpcore to stop pinning.

mdellweg: As long as we do the monkey patching there is no stopping.

mdellweg: Nothing without badly breaking the client bindings.

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  commit-message:
+    prefix: "[noissue]"
```
mdellweg, line 9: Ugh this is what will activate Dependabot on all forks.

Any new reason to add this file?

ipanova: yes? who will manage the automatic updates of the released deps?https://github.com/pulp/pulpcore/blob/main/docs/plugins/plugin-writer/concepts/index.rst#declaring-dependencies

I don;t think it will activate on forks, can you point me to the config you then expect to see?

mdellweg: This is rather a _big_ annoyance. Not a reason to stop doing this.

mdellweg: BTW: https://github.com/dependabot/dependabot-core/issues/2804

requirements.txt
ipanova: @bmbouter @mdellweg what range do you expect to be used here?

mdellweg: I would still call drf-spectacular part of the pulpcore plugin interface. But I do not intend to reroll the whole discussion from the beginning.
How about `drf-spectacular>=0.24,<0.25`?

bmbouter: I'd be ok with either drf-spectacular>=0.24,<0.25 or not declaring it at all.

mdellweg (approved):

PR #1110, opened 2022-10-11
Assert on is_superuser within the registry API

### LINKED ISSUE #1109: Staff users should not have permissions like superusers
https://github.com/pulp/pulp_container/blob/61e2307e6346181bcf0cd2ed40c626cad508af15/pulp_container/app/token_verification.py#L176

https://stackoverflow.com/a/20469727/3907906

The solution is to replace the "is_staff" with "is_superuser".

lubosmj:
closes #1109

### DIFF
--- a/CHANGES/1109.bugfix
+++ b/CHANGES/1109.bugfix
@@ -1,0 +1,2 @@
+Fixed an issue that caused all staff users to have superuser permissions when accessing the
+registry without token authentication enabled.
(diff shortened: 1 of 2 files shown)


lubosmj:
It might require some backports.

mdellweg (approved):

ipanova (approved):
Great find, thankfully we have already removed the django admin site and our admin user by default is_staff too.

PR #1111, opened 2022-10-12
Update CI files

mikedep333:
(pulp_container is one of the repos where /.pytest_cache warnings are found, so I am testing out updating it.)

[noissue]


lubosmj (dismissed):

.github/workflows/scripts/install.sh
```diff
@@ -145,6 +145,9 @@ ansible-playbook start_container.yaml
 chmod 777 ~/.config/pulp_smash/
 chmod 666 ~/.config/pulp_smash/settings.json
 sudo chown -R 700:700 ~runner/.config
+# Plugins often write to ~/.config/pulp/cli.toml from the host
+sudo chmod 777 ~runner/.config/pulp
+sudo chmod 666 ~runner/.config/pulp/cli.toml
```
mdellweg, line 150: I'm not seeing this. The cli should not even be installed inside the container at all.

PR #1138, opened 2022-11-22
Podman client could not verity manifest indices signed by pulp signing

### LINKED ISSUE #1135: Podman client cannot verify content signed off by a signing service
**Version**
podman - version 3.4.7
pulpcore, pulp_container - main branch

**Describe the bug**
Content signed by a signing service cannot be pulled from the Registry once the verification policy is set up.

**To Reproduce**
- Generate a new GPG signature and export the public key for further verification purposes:
```
cat > gpg-genkey-unattended <<'EOF'
Key-Type: RSA
Key-Length: 2048
Name-Real: Lubos
Name-Email: lubos@lubos.com
Expire-Date: 0
Passphrase: password
%commit
EOF

gpg --batch --gen-key gpg-genkey-unattended

gpg --output pubkey-lubos.pgp --export --armor lubos@lubos.com
```

- Configure a signing script (with a correct path to the password file):
```
cat > signing-script-skopeo.sh <<'EOF'
#!/usr/bin/env bash

MANIFEST_PATH=$1
FINGEPRINT="$PULP_SIGNING_KEY_FINGERPRINT"
IMAGE_REFERENCE="$REFERENCE"
SIGNATURE_PATH="$SIG_PATH"

# Create container signature
skopeo standalone-sign $MANIFEST_PATH $IMAGE_REFERENCE $FINGEPRINT --output $SIGNATURE_PATH --passphrase-file /home/vagrant/lubos-password
# Check the exit status
STATUS=$?
if [ $STATUS -eq 0 ]; then
  echo {\"signature_path\": \"$SIGNATURE_PATH\"}
else
  exit $STATUS
fi
EOF
```

- Initialize a new signing service:
```
export PUBKEY_FP=$(gpg --list-keys lubos@lubos.com | sed -n 2p | xargs)

pulpcore-manager add-signing-service skopeo-signing-test signing-script-skopeo.sh ${PUBKEY_FP} --class container:ManifestSigningService
```

- Mirror a repository and sign the mirrored content:
```
pulp container remote create --name test --url https://ghcr.io --upstream-name pulp/hello-world --policy immediate

pulp container repository create --name test --remote test

pulp container repository sync --name test

pulp container repository content list --repository test

pulp container distribution create --name test --base-path test --repository test

export SERVICE_HREF=$(http :24817/pulp/api/v3/signing-services/?name=skopeo-signing-test | jq -r ".results[0].pulp_href")

export REPO_HREF=$(pulp container repository show --name test | jq -r ".pulp_href")

http :24817${REPO_HREF}sign/ manifest_signing_service=${SERVICE_HREF} future_base_path=test
```

- Push and sign new content to the Registry:
```
podman pull ghcr.io/pulp/hello-world

podman tag ghcr.io/pulp/hello-world localhost:24817/pulp/hello-world

podman push --sign-by lubos@lubos.com localhost:24817/pulp/hello-world --tls-verify=false
```

- Update the verification policy (`/etc/containers/policy.json`), like so:
```
cat /etc/containers/policy.json 
{
    "default": [
        {
	    "type": "reject"
        }
    ],
    "transports": {
        "docker": {
	    "localhost:24817": [
		{
		    "type": "signedBy",
		    "keyType": "GPGKeys",
		    "keyPath": "/home/vagrant/pubkey-lubos.pgp"
		}
	    ]
	},
        "docker-daemon": {
	    "": [
		{
		    "type": "insecureAcceptAnything"
		}
	    ]
	}
    }
}

```
- Pull the content from `pulp/hello-world` and `test`:
```
podman pull localhost:24817/pulp/hello-world --tls-verify=false

podman pull localhost:24817/test --tls-verify=false

Trying to pull localhost:24817/test:latest...
Error: Source image rejected: Signature for identity localhost:24817/test:sha256:239de6dd745ed1a211123322865b4c342c706e7c1e01644a1bbefe8f8846c5ff is not accepted
```

**Expected behaviour**
The content should be downloadable from the Registry if it is signed or re-signed by a signing service and the corresponding public key is referenced from the `/etc/containers/policy.json` file.

ipanova:
service.
closes #1135

Signed-off-by: Ina Panova <ipanova@redhat.com>

### DIFF
--- a/CHANGES/1135.bugfix
+++ b/CHANGES/1135.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where the Podman client could not verify manifest indices signed with a Pulp signing service.
--- a/pulp_container/app/tasks/sign.py
+++ b/pulp_container/app/tasks/sign.py
@@ -53,19 +53,15 @@ def sign(repository_pk, signing_service_pk, reference, tags_list=None):
     already_signed = []
     for tag in latest_repo_tags:
         tagged_manifest = tag.tagged_manifest
-        if tagged_manifest.media_type in MANIFEST_MEDIA_TYPES.IMAGE:
-            signature_pk = create_signature(
-                tagged_manifest, ":".join((reference, tag.name)), signing_service
-            )
-            added_signatures.append(signature_pk)
-        elif tagged_manifest.media_type in MANIFEST_MEDIA_TYPES.LIST:
-            # parse ML and sign per-arches by digest
+        docker_reference = ":".join((reference, tag.name))
+        signature_pk = create_signature(tagged_manifest, docker_reference, signing_service)
+        added_signatures.append(signature_pk)
+        if tagged_manifest.media_type in MANIFEST_MEDIA_TYPES.LIST:
+            # parse ML and sign per-arches
             for manifest in tagged_manifest.listed_manifests.iterator():
                 # image manifests can be present in multiple ML within the repo
                 if manifest.digest not in already_signed:
-                    signature_pk = create_signature(
-                        manifest, ":".join((reference, manifest.digest)), signing_service
-                    )
+                    signature_pk = create_signature(manifest, docker_reference, signing_service)
                     already_signed.append(manifest.digest)
                     added_signatures.append(signature_pk)
 
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -171,6 +171,12 @@ class ContainerFirstStage(Stage):
                         list_dc.extra_data["listed_manifests"].append(listed_manifest)
 
                     else:
+                        # Manifest indices can be signed too. It is not mandatory.
+                        # If signature is available mirror it.
+                        if signature_source is not None:
+                            list_sig_dcs = await self.create_signatures(list_dc, signature_source)
+                            if list_sig_dcs:
+                                self.signature_dcs.extend(list_sig_dcs)
                         # only pass the manifest list and tag down the pipeline if there were no
                         # issues with signatures (no `break` in the `for` loop)
                         tag_dc.extra_data["tagged_manifest_dc"] = list_dc


pulp_container/app/tasks/sign.py
```diff
@@ -53,19 +53,15 @@ def sign(repository_pk, signing_service_pk, reference, tags_list=None):
...
+        added_signatures.append(signature_pk)
+        if tagged_manifest.media_type in MANIFEST_MEDIA_TYPES.LIST:
+            # parse ML and sign per-arches
             for manifest in tagged_manifest.listed_manifests.iterator():
                 # image manifests can be present in multiple ML within the repo
                 if manifest.digest not in already_signed:
-                    signature_pk = create_signature(
-                        manifest, ":".join((reference, manifest.digest)), signing_service
-                    )
+                    signature_pk = create_signature(manifest, docker_reference, signing_service)
```
lubosmj, line 64: Can we end up in a situation where we will be constantly adding new signatures to the very same manifest listed from different manifest lists? Not sure of any drawback right at the moment. Maybe, it can screw up the number of signatures referenced by manifests?

CHANGES/1135.bugfix
```diff
@@ -0,0 +1 @@
+Podman client could not verify manifest indices signed with pulp signing service.
```
lubosmj, line 1:
```suggestion
Fixed a bug where the Podman client could not verify manifest indices signed with a Pulp signing service.
```

pulp_container/app/tasks/sync_stages.py
```diff
@@ -171,6 +171,12 @@ async def run(self):
                         list_dc.extra_data["listed_manifests"].append(listed_manifest)
 
                     else:
+                        # Manifest indices can be signed too. It is not mandatory.
+                        # If signature is available mirror it.
+                        if signature_source is not None:
+                            list_sig_dcs = await self.create_signatures(list_dc, signature_source)
```
lubosmj, line 177: The `create_signatures` method will need to be adjusted for manifest lists (OCI indexes). It tries to use the digest of a manifest when initializing signatures. There is no digest for the manifest lists in the payload: https://github.com/pulp/pulp_container/blob/343859eedfa64a022a7062c39372e414e89f7b6b/pulp_container/app/tasks/sync_stages.py#L527

lubosmj (changes requested):

pulp_container/app/tasks/sign.py
ipanova: it can be same manifest in other manifest-lists but it will always be referenced by different tag, i.e docker-reference will be different.

pulp_container/app/tasks/sync_stages.py
lubosmj: Invalid.

lubosmj:
```
list_dc.content.digest
Out[8]: 'sha256:8a2bd13f5d1cbc36c2a0a5ee2fdc544b5320c2393768a3e9b96b476c8faeab68'
```

ipanova: we are retrieving the digest not from the payload but from the declarative content of ML we have just created on the line earlier ` list_dc = self.create_tagged_manifest_list`

lubosmj (approved):

pulp_container/app/tasks/sign.py
lubosmj: I was worried about adding new signatures to the same manifest over and over again. But, we will use different tags.

ipanova: if you have manifest A which is used in manifest lists in repos FOO and BAR - that manifest during signing will have different docker-reference: foo:latest ,bar:latest, bar:nightly
But if you trigger sign task again for these repos, yes you will have on top of existing signature created exactly same one, expect for the timestamp. that will be only thing that will be different in the signature payload.

ipanova: since the digest of the signature is calculated based on the payload, this approach brings certain benefits as well as drawbacks, i.e if you in a loop trigger 10 signing tasks on the same repo - it will produce 10 signatures per manifest.

PR #1178, opened 2022-12-12
Added .gitleaks

ipanova:
This will tell scanners to ignore testing credentials otherwie false positives are created.

[noissue]

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,5 @@
+[allowlist]
+  description = "Our test install exports a test only MINIO ACCESS KEY"
+  paths = [
+    ".github/workflows/scripts/install.sh",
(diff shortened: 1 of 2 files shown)


.gitleaks.toml
```diff
@@ -0,0 +1,5 @@
+[allowlist]
+  description = "Our test install exports a test only MINIO ACCESS KEY"
+  paths = [
+    ".github/workflows/scripts/install.sh",
```
lubosmj, line 4: Is it worth changing the path to the `.github/workflows/scripts/` instead?

ipanova: The more precise the better.

lubosmj (approved):

PR #1179, opened 2022-12-12
Delete the deprecated _base_paths implementation

lubosmj:
[noissue]


pulp_container/app/registry.py
```diff
@@ -44,7 +44,7 @@ def _base_paths(path):
             list: Of base paths.
 
         """
-        return [path]
+        return [path.rstrip("/")]
```
ipanova, line 47: I would not say that the implementation is deprecated... Even though we ensure through the serializer validation certain rules for the base_path, later on it is being modified the pulpcore where the slash is being added..
I understand that this method we're overriding is not exposed to the plugin writers, maybe we should. https://github.com/pulp/pulpcore/pull/3434/files#r1047175632

Fortunately this change is harmless in the current state of affairs ( but if in the future if  there will be more breaking changes in core we will end up in the situation again where we'd need to take an action) , I'm only suggesting to update the commit message so it does not contain word "deprecated" but "ensure base_path does not end with slash" or similar

ipanova: Per pulpcore discussions, this fix is not enough, because of forward compat we declare. The fix would have needed to be backported to all pulp-container versions compat with pulpcore >=3.20, <3.25


The conclusion is to make core accountable for the change.

lubosmj: Thanks for the heads up. I am closing this PR then.

ipanova: i was wondering maybe you could still get the tests in? ;)

lubosmj: Okay, I opened a new PR for that: https://github.com/pulp/pulp_container/pull/1180

Issue #490, opened 2021-12-15
Add docs that explain why repos named busybox and busybox/test cannot be pushed


pulpbot:
Author: @ipanova (ipanova@redhat.com)

Redmine Issue: 8885, https://pulp.plan.io/issues/8885

---

User just gets 'invalid repository name' this does not get a lot of information and further guidance.
```
$ podman tag 6dbb9cc54074 localhost:24817/ina/tst
(pulp) [vagrant@pulp3-source-fedora32 ~]$ podman push  localhost:24817/ina/tst
Getting image source signatures
Copying blob b2d5eeeaba3a done  
Copying config 6dbb9cc540 done  
Writing manifest to image destination
Storing signatures
(pulp) [vagrant@pulp3-source-fedora32 ~]$ podman tag 6dbb9cc54074 localhost:24817/ina
(pulp) [vagrant@pulp3-source-fedora32 ~]$ podman push  localhost:24817/ina
Getting image source signatures
Copying blob b2d5eeeaba3a [--------------------------------------] 8.0b / 5.6MiB
Error: error copying image to the remote destination: Error writing blob: Error initiating layer upload to /v2/ina/blobs/uploads/ in localhost:24817: name invalid: Invalid repository name.
```

ipanova:
This limitation is not longer present.

PR #1180, opened 2022-12-13
Set the namespace of distributions to the same value

lubosmj:
This addition verifies whether the plugin correctly handles the lookup for correct distributions within the shared namespace or base_path.

For example, the following scenario is now being evaluated:

distribution(base_path=aaa) != distribution(base_path=aaa/bbb)
distribution(base_path=aaa/bbb) != distribution(base_path=aaa/ccc)

[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_pull_content.py
+++ b/pulp_container/tests/functional/api/test_pull_content.py
@@ -4,6 +4,7 @@ import hashlib
 import json
 import requests
 import unittest
+import uuid
 
 from urllib.parse import urljoin, urlparse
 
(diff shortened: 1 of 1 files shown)


lubosmj:
This test will currently result in internal server errors until we revert the commit https://github.com/pulp/pulpcore/commit/9d85abf914a7c301b6262f6eb2420d14efd54cc7.

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -70,10 +71,23 @@ def setUpClass(cls):
 
         delete_orphans()
 
+        # defining a shared path will help us to identify whether namespaces and
+        # matching distributions are correctly determined from within the plugin
+        random_dist_path = str(uuid.uuid4())
```
ipanova, line 76: I was reviewing this PR and remembered we used to have this limitation https://github.com/pulp/pulp_container/issues/490
Now it is no longer there, so I am going to close it

ipanova (approved):

PR #1204, opened 2023-01-10
Migrate the RegistryClient class from pulp-smash

lubosmj:
Required PR: https://github.com/pulp/pulpcore/pull/3498

[noissue]

### DIFF
--- a/pulp_container/tests/functional/conftest.py
+++ b/pulp_container/tests/functional/conftest.py
@@ -1,13 +1,14 @@
+import json
 import os
 import stat
 import pytest
 import requests
+import subprocess
 
+from contextlib import contextmanager
+from functools import partialmethod
 from urllib.parse import urljoin, urlparse
 from uuid import uuid4
-
-from pulp_smash.utils import execute_pulpcore_python, get_pulp_setting
-from pulp_smash.cli import RegistryClient
 
 from pulpcore.client.pulp_container import (
     ApiClient,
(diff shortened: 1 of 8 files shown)


lubosmj:
I am un-drafting this PR.

I wanted to also include the changes where I replaced `pulp_cfg.get_base_url()` with `bindings_cfg.host` in `_local_registry` and `test_cross_repository_blob_mount`. However, intermittent errors were raised during `podman manifest list push oci` operations in `test_push_content`. Not sure if it is worth investigating the root cause at the moment.

pulp_container/tests/functional/conftest.py
```diff
@@ -30,10 +32,76 @@
...
+        # Scheme should not be part of image path, if so, remove it.
+        cmd = (self.name, command) + tuple(args)
+        result = subprocess.check_output(cmd).decode()
+        try:
+            # most client responses are JSONable
+            return json.loads(result)
+        except json.JSONDecodeError:
+            return result
+
+    pull = partialmethod(_dispatch_command, "pull")
```
decko, line 81: :pinched_fingers: Bravo!

pulp_container/tests/functional/conftest.py
```diff
@@ -30,10 +32,76 @@
...
+            os.environ.update(old_environ)
+
+    def _get_registry_client(self):
+        if subprocess.run(("which", self.NAME)).returncode == 0:
+            return self.NAME
+
+        raise RuntimeError("The client '{}' does not appear to be installed.".format(self.NAME))
+
+    def _dispatch_command(self, command, *args):
+        # Scheme should not be part of image path, if so, remove it.
```
ipanova, line 72: leftovers

ipanova:
> I am un-drafting this PR.
> 
> I wanted to also include the changes where I replaced `pulp_cfg.get_base_url()` with `bindings_cfg.host` in `_local_registry` and `test_cross_repository_blob_mount`. However, intermittent errors were raised during `podman manifest list push oci` operations in `test_push_content`. Not sure if it is worth investigating the root cause at the moment.

Eventually these changes will be included too so i believe you're just postponing the investigation, which is fine too.

lubosmj:
The `required PR` reference is here just to verify whether the change in pulpcore has an impact on the CI in pulp_container.

pulp_container/tests/functional/conftest.py
```diff
@@ -30,10 +32,75 @@
...
+        except RuntimeError:
+            raise exc(message)
+
+    @contextmanager
+    def set_env(self, **environ):
+        old_environ = os.environ.copy()
+        os.environ.update(environ)
+        try:
+            yield
+        finally:
```
ipanova, line 61: mm except missing?

lubosmj: Not really. The goal is the clear up the updated environment variables at all costs.

ipanova: tbh i have never seen try block without except, got any examples?

lubosmj: https://docs.python.org/3/reference/compound_stmts.html#finally-clause

lubosmj: https://github.com/pulp/pulpcore/blob/48d0ef772af1b1484799bd86ba29979a0949be52/pulpcore/app/models/storage.py#L56-L59

lubosmj: @mdellweg, do you see any added value in logging a raised error? Or, do you think it is sufficient to throw the exception to stdout/stderr as it is right now - no `except` statement?

ipanova (approved):

PR #1210, opened 2023-01-28
Fixed signing task that was skipping some image signing.

### LINKED ISSUE #1209: Signing task can skip some image signing
**Version**
main

**Describe the bug**
Image manifests can be present in multiple ML within the repo.
With the current logic if same image manifest is present in 3 ML within the repo, this manifest will be signed just once.

**To Reproduce**
https://github.com/pulp/pulp_container/blob/main/pulp_container/app/tasks/sign.py#L65

ipanova:
closes #1209

### DIFF
--- a/CHANGES/1209.bugfix
+++ b/CHANGES/1209.bugfix
@@ -1,0 +1,1 @@
+Fixed signing task that could skip some image signing.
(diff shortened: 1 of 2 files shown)


mdellweg (approved):
I see there is no constraint that you could only have a single signature per manifest. So +1.

PR #1220, opened 2023-02-27
Stream only blobs.

ipanova:
Manifest artifact is always expected to be found, even in the on_demand sync. If artifact is missing then repo was corrupted. Repair should be called then.

[noissue]

### DIFF
--- a/pulp_container/app/registry.py
+++ b/pulp_container/app/registry.py
@@ -155,8 +155,7 @@ class Registry(Handler):
 
     async def dispatch_tag(self, request, tag, response_headers):
         """
-        Finds an artifact associated with a Tag and sends it to the client, otherwise tries
-        to stream it.
+        Finds an artifact associated with a Tag and sends it to the client
 
         Args:
             request(:class:`~aiohttp.web.Request`): The request to prepare a response for.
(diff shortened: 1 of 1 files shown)


ipanova:
in light of recent found issue https://github.com/pulp/pulpcore/issues/3606 we should use streaming only for blobs because those are not headers sensitive

pulp_container/app/registry.py
```diff
@@ -249,10 +242,9 @@ async def get_by_digest(self, request):
             raise PathNotResolved(path)
         else:
             artifact = ca.artifact
-            if artifact:
-                return await Registry._dispatch(artifact, headers)
-            else:
+            if not artifact and isinstance(ca_content, Blob):
```
lubosmj, line 245:
```suggestion
            if not artifact and media_type == BLOB_CONTENT_TYPE:
```

lubosmj (dismissed):
Thanks for submitting this change!

pulp_container/app/registry.py
```diff
@@ -249,10 +242,9 @@ async def get_by_digest(self, request):
...
                 return await self._stream_content_artifact(request, web.StreamResponse(), ca)
+            return await Registry._dispatch(artifact, headers)
```
lubosmj, line 247: Do we need to consider any fallback in the case when there is no artifact present and yet the ca_content variable is not of the `Blob` type?

lubosmj: We might end up returning HTTP 500 in such a case.

lubosmj: How about returning an error like "the manifest could not be located, please, run the repair command"? I am not sure how container clients will cope with this.

ipanova: While there is an expectation that manifest would always have artifact downloaded even in the on_demand case, i forgot to remember our lovely 'reclaim space' feature that would potentially delete the artifact from the storage and db.

In that case we should probably focus then on how to get fixed the pulpcore downloader bug and close this PR.

ipanova: We could also add a safeguard check and raise 404 for when artifact is not found for type != blob ( that is to circumvent the downloader bug mostly..)

lubosmj: I am all in for the safeguard check until we resolve the downloader issue.

ipanova: I will add path not resolved, but now I am not sure myself I like this, since behavior for blobs and manifests will differ.

ipanova:
@lubosmj yes let's close it, we should whether address this issue with artifactless manifest or fixing the one in core

PR #1226, opened 2023-03-02
Use extract_pk, raise_for_unknown_content_units from pulpcore.plugin.…

ipanova:
…util

[noissue]


requirements.txt
```diff
@@ -1,6 +1,6 @@
 # remember to also update unittest_requirements.txt when updating this file
 ecdsa>=0.14,<=0.18.0
 jsonschema>=4.4,<4.18
-pulpcore>=3.20.0.dev,<3.25
+pulpcore>=3.23.0.dev,<3.25
```
lubosmj, line 4: Is it worth doing this? We wanted to have broader versioning to not enforce updates of particular plugins.

```suggestion
pulpcore>=3.25.0.dev,<3.30
```

ipanova: sure, this was temporarily bumped to see the ci green.

mdellweg: Unless we stop adding deprecations for 3.25 in pulpcore today, we cannot move the upper bound.
So i think the requirements are right that way and should actually be `pulpcore>=3.23.0,<3.25` when we merge obviously after releaseing pulpcore 3.23.

lubosmj (approved):

PR #1255, opened 2023-04-11
Handle unauthorized exceptions in a custom handler

### LINKED ISSUE #1254: Can't configure authentication on token API
**Version**
2.14.3

**Describe the bug**
Automation Hub relies on being able to configure the default DRF authentication classes in order to be able to provide podman and docker authentication when keycloak is configured. 

**Expected behavior**
The token API on the container registry is hard coded to use basic auth and can't be reconfigured

lubosmj:
closes #1254

### DIFF
--- a/pulp_container/app/settings.py
+++ b/pulp_container/app/settings.py
@@ -1,4 +1,12 @@
+from copy import deepcopy
+from django.conf import settings
+
 DRF_ACCESS_POLICY = {
     "dynaconf_merge_unique": True,
     "reusable_conditions": ["pulp_container.app.global_access_conditions"],
 }
+
+REST_FRAMEWORK = deepcopy(settings.REST_FRAMEWORK)
+REST_FRAMEWORK.update(
+    {"EXCEPTION_HANDLER": "pulp_container.app.exceptions.unauthorized_exception_handler"}
+)
--- a/CHANGES/1254.bugfix
+++ b/CHANGES/1254.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug that disallowed users to configure custom authentication classes for the token server.
--- a/pulp_container/app/exceptions.py
+++ b/pulp_container/app/exceptions.py
@@ -1,4 +1,19 @@
-from rest_framework.exceptions import NotFound, ParseError
+from rest_framework import status, views
+from rest_framework.exceptions import (
+    AuthenticationFailed,
+    NotAuthenticated,
+    NotFound,
+    ParseError,
+)
+
+
+def unauthorized_exception_handler(exc, context):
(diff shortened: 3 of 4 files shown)


pulp_container/app/settings.py
```diff
@@ -1,4 +1,11 @@
+from django.conf import settings
+
 DRF_ACCESS_POLICY = {
     "dynaconf_merge_unique": True,
     "reusable_conditions": ["pulp_container.app.global_access_conditions"],
 }
+
+REST_FRAMEWORK = settings.REST_FRAMEWORK
+REST_FRAMEWORK.update(
+    {"EXCEPTION_HANDLER": "pulp_container.app.exceptions.unauthorized_exception_handler"}
```
newswangerd, line 10: Won't this set the exception handler for all plugins?

lubosmj: Hmm, that is a good question. I think it probably will. 😭

lubosmj: I can try cloning this to make it available only to pulp_container.

pulp_container/app/settings.py
```diff
@@ -1,4 +1,12 @@
...
+from django.conf import settings
+
 DRF_ACCESS_POLICY = {
     "dynaconf_merge_unique": True,
     "reusable_conditions": ["pulp_container.app.global_access_conditions"],
 }
+
+REST_FRAMEWORK = deepcopy(settings.REST_FRAMEWORK)
+REST_FRAMEWORK.update(
+    {"EXCEPTION_HANDLER": "pulp_container.app.exceptions.unauthorized_exception_handler"}
```
rochacbruno, line 11:
```suggestion
DRF_ACCESS_POLICY__reusable_conditions = "@merge_unique pulp_container.app.global_access_conditions"

REST_FRAMEWORK__EXCEPTION_HANDLER = "pulp_container.app.exceptions.unauthorized_exception_handler"
```

pulp_container/app/settings.py
```diff
@@ -1,4 +1,12 @@
+from copy import deepcopy
+from django.conf import settings
```
rochacbruno, line 2: I don' t recommend importing `django.conf.settings` here, you can use dynaconf inline merging with `@ ` and  `__` (as in my other comment)

Or you can add a `pulp_container/app/dynaconf_hooks.py` and then do any conditionals or updates there.

`dynaconf_hooks.py`
```py
def post(settings):
    data = {"dynaconf_merge": True}
    data["REST_FRAMEWORK"]["EXCEPTION_HANDLER"] = "..."
    return data
```

The reason is that as settings is formed by the merging of multiple sources the final state is not the one you get grom django.conf at this point here.

PR #1265, opened 2023-04-17
Sign manifests asynchronously

### LINKED ISSUE #1208: Sign manifests concurrently
Manifest signing is serial now. 

**Describe the solution you'd like**
Manifest signing can be sped up by increasing concurrency.

**Additional context**
https://github.com/pulp/pulp_deb/pull/683
https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/tasks/signature.py#L131

lubosmj:
closes #1208

### DIFF
--- a/pulp_container/app/tasks/sign.py
+++ b/pulp_container/app/tasks/sign.py
@@ -1,8 +1,9 @@
+import asyncio
 import base64
 import hashlib
-import os
-import tempfile
 
+from aiofiles import tempfile
+from asgiref.sync import sync_to_async
 from django.conf import settings
 
 from pulpcore.plugin.models import Repository
@@ -19,6 +20,8 @@ from pulp_container.constants import (
     SIGNATURE_TYPE,
 )
 
+semaphore = asyncio.Semaphore(settings.MAX_PARALLEL_SIGNING_TASKS)
+
 
 def sign(repository_pk, signing_service_pk, reference, tags_list=None):
     """
@@ -27,8 +30,8 @@ def sign(repository_pk, signing_service_pk, reference, tags_list=None):
     Create signature for each manifest that is specified and add it to the repo.
     If no manifests were specified, then sign all manifests in the repo.
 
-    What manifests to sign is identified by tag.
-    Manifest lists are not signed. Image manifests from manifest list are signed by digest.
+    What manifests to sign is identified by tags.
+    Manifest lists are signed too. Image manifests from the manifest lists are signed by tags.
 
     Args:
         repository_pk (uuid): A pk for a Repository for which a new Repository Version should be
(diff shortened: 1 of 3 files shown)


pulp_container/app/tasks/sign.py
```diff
@@ -81,20 +100,23 @@ def create_signature(manifest, reference, signing_service):
...
+    async with semaphore:
         # download and write file for object storage
+        artifact = await sync_to_async(manifest._artifacts.get)()
         if settings.DEFAULT_FILE_STORAGE != "pulpcore.app.models.storage.FileSystem":
-            manifest_file = tempfile.NamedTemporaryFile(dir=working_directory, delete=False)
-            artifact = manifest._artifacts.get()
-            manifest_file.write(artifact.file.read())
-            manifest_file.flush()
+            async with tempfile.NamedTemporaryFile(dir=".", mode="wb", delete=False) as tf:
+                await tf.write(artifact.file.read())
```
mdellweg, line 108:
```suggestion
                await tf.write(await sync_to_async(artifact.file.read)())
```
This should probably not be blocking...

lubosmj:
I cannot really test the changes locally because of multiple permission errors in oci_env. Is someone with a vagrant box volunteering to test the changes? Or, do we trust our CI tests?

ipanova:
I can test, my vagrant box seems to be alive.

pulp_container/app/tasks/sign.py
lubosmj: Based on your latest commit, I am reverting the suggested change:
https://github.com/pulp/pulp_container/pull/1285#discussion_r1196595234

Can you sign this off?

ipanova:
can you update this line https://github.com/pulp/pulp_container/pull/1265/files#diff-94c6436dad7eb9a689d63e5b6e6ceb0d18ddac41ce565c1aa91313c1be8ccfc2R35 we sign manifest lists and children manifests are signed by tag too

pulp_container/app/tasks/sign.py
```diff
@@ -1,8 +1,8 @@
+import asyncio
 import base64
 import hashlib
-import os
-import tempfile
 
+from aiofiles import tempfile
```
mdellweg, line 5: Oh... That name is a bit deceiving. But ok.

pulp_container/app/tasks/sign.py
```diff
@@ -47,26 +51,40 @@ def sign(repository_pk, signing_service_pk, reference, tags_list=None):
...
+        async for tag in latest_repo_tags.aiterator():
+            tagged_manifest = tag.tagged_manifest
+            docker_reference = ":".join((reference, tag.name))
+            signature_pk = await create_signature(
+                tagged_manifest, docker_reference, signing_service
+            )
+            added_signatures.append(signature_pk)
+            if tagged_manifest.media_type in MANIFEST_MEDIA_TYPES.LIST:
+                # parse ML and sign per-arches
+                added_signatures = []
```
mdellweg, line 71: Why are you clearing that list here?

pulp_container/app/tasks/sign.py
mdellweg: Trying to understand what i meant back then. I still think that django storages does not yet know about async. So "artifact.file.read" (which is probably a blocking call,) should not be called from an async context directly, but `sync_to_async` as i suggested.

pulp_container/app/tasks/sign.py
lubosmj: I should not. Oopsie woopsie. 👀

pulp_container/app/tasks/sign.py
```diff
@@ -81,20 +99,23 @@ def create_signature(manifest, reference, signing_service):
...
         else:
-            manifest_path = manifest._artifacts.get().file.path
-        sig_path = os.path.join(working_directory, "signature")
+            manifest_path = artifact.file.path
+
+        async with tempfile.NamedTemporaryFile(dir=".", prefix="signature") as tf:
+            sig_path = tf.name
 
-        signed = signing_service.sign(
+        signed = await signing_service.asign(
```
ipanova, line 118: where is asign?

pulp_container/app/tasks/sign.py
```diff
@@ -19,6 +20,10 @@
     SIGNATURE_TYPE,
 )
 
+SIGNING_TASKS_COUNTER = 10
```
ipanova, line 23: can we expose this to the settings so it is configurable?
I know, this PR then won't be backportable, which is ok given that we don't have now any BZs opened and if we will then backporting without the settings is always an option.

ipanova (approved):
thanks!

PR #1269, opened 2023-04-21
Ephemeral blobs

mdellweg:
depending on this one https://github.com/pulp/pulp_container/pull/1249


pulp_container/app/registry_api.py
```diff
@@ -599,34 +597,12 @@ def mount_blob(self, request, path, repository):
         digest = request.query_params["mount"]
         try:
             blob = models.Blob.objects.get(digest=digest, pk__in=version.content)
+            blob.touch()
```
ipanova, line 600: it's in the repo version, no need to touch?

pulp_container/app/registry_api.py
```diff
@@ -978,7 +948,8 @@ def put(self, request, path, pk=None):
                 blobs.add(digest)
 
             blobs_qs = models.Blob.objects.filter(
-                digest__in=blobs, pk__in=repository.latest_version().content
+                digest__in=blobs,
```
ipanova, line 951: you need to do same for config blob in line 919

pulp_container/app/registry_api.py
```diff
@@ -1009,6 +980,8 @@ def put(self, request, path, pk=None):
         else:
             add_content_units = [str(manifest.pk)]
             remove_content_units = []
+        # TODO: Is this necessary, or does the the manifest pull it's blobs automatically?
```
ipanova, line 983: if i get your question right - it is necessary,  you need to add these blob pks as part of add_remove_task, so you add tag, manifest, blobs in one go

pulp_container/app/registry_api.py
```diff
@@ -761,20 +737,8 @@ def put(self, request, path, pk=None):
...
-            kwargs={
-                "repository_pk": str(repository.pk),
-                "add_content_units": [str(blob.pk)],
-                "remove_content_units": [],
-            },
-        )
-
-        if has_task_completed(dispatched_task):
-            return BlobResponse(blob, path, 201, request)
+        repository.pending_blobs.add(blob)
```
ipanova, line 740: since you should removed the task dispatch you should remove line 677 and instead of those check repo.pending.blobs containing the blob digest in question

lubosmj:
I am closing this PR in favour of https://github.com/pulp/pulp_container/pull/1273, where this and the referenced PR is merged. Thanks for working on this. It is a huge improvement!

PR #1273, opened 2023-05-02
Use immediate task for manifest and signature push

### LINKED ISSUE #1212: Make manifest upload less error prone
**Version**
main

**Describe the bug**
Image push can fail with 429 too many requests.
Client can properly backoff during 429 blob upload 1s,2s,4s,8s,16s (tot 31s) and wait for the task to finish, client does not back off during manifest push. Api is locked only for 3 seconds and if  within that time tasks did not complete, client will re-try ( multiple times, podman/skopeo re-tries 3 times by default )by making another request which will lead to another task creation. Only first task will be operational, other subsequent tasks will be no-op.
The overall podman push command will have a non 0 exit code, however the image manifest will be properly uploaded to the registry. It means that task will complete after client will fail

I am not sure what would be best to do.
a) we can lock api for 3+ seconds but this will fail  anyway if workers are under load and task will be picked up after 3+ seconds, locking api for long time is not good approach anyway https://github.com/pulp/pulp_container/blob/main/pulp_container/app/utils.py#L130
b) we could check in the existing repo version whether such manifest and tag already are there and not dispatch the task ( which will anyway be no-op). This is also a half solution because first try will anyway fail( task will be dispatched/finish after 3 secs, however on the second try no task will be dispatched and 201 will be returned to the client  https://github.com/pulp/pulp_container/blob/main/pulp_container/app/registry_api.py#L1005
c) dispatch task for the manifest upload but do not check it result/wait for it and return 201 https://github.com/pulp/pulp_container/blob/main/pulp_container/app/registry_api.py#L1021 . This is not best either because task might not be picked up for X time and client will not be able to pull from that repo and in adidition will not know why because of misleading 201
d) other

**To Reproduce**
```
$ skopeo copy --retry-times 5    docker://registry.redhat.io/ansible-automation-platform-23/ee-minimal-rhel8 docker://pulp3-source-fedora36.puffy.example.com/ee-minimal-rhel8:latest --remove-signatures --dest-creds admin:password
Copying blob 370fc4ac3f1f skipped: already exists  
Copying blob 7ecdec58f172 skipped: already exists  
Copying blob 272c29e2f36c skipped: already exists  
Copying config 5a85baceb7 done  
WARN[0005] Failed, retrying in 1s ... (1/5). Error: copying system image from manifest list: writing manifest: uploading manifest latest to pulp3-source-fedora36.puffy.example.com/ee-minimal-rhel8: unsupported: Request was throttled. 
Copying blob 370fc4ac3f1f skipped: already exists  
Copying blob 7ecdec58f172 skipped: already exists  
Copying blob 272c29e2f36c skipped: already exists  
Copying config 5a85baceb7 done  
WARN[0010] Failed, retrying in 2s ... (2/5). Error: copying system image from manifest list: writing manifest: uploading manifest latest to pulp3-source-fedora36.puffy.example.com/ee-minimal-rhel8: unsupported: Request was throttled. 
Copying blob 7ecdec58f172 skipped: already exists  
Copying blob 370fc4ac3f1f skipped: already exists  
Copying blob 272c29e2f36c skipped: already exists  
Copying config 5a85baceb7 done  
WARN[0016] Failed, retrying in 4s ... (3/5). Error: copying system image from manifest list: writing manifest: uploading manifest latest to pulp3-source-fedora36.puffy.example.com/ee-minimal-rhel8: unsupported: Request was throttled. 
Copying blob 370fc4ac3f1f skipped: already exists  
Copying blob 7ecdec58f172 skipped: already exists  
Copying blob 272c29e2f36c skipped: already exists  
Copying config 5a85baceb7 done  
WARN[0024] Failed, retrying in 8s ... (4/5). Error: copying system image from manifest list: writing manifest: uploading manifest latest to pulp3-source-fedora36.puffy.example.com/ee-minimal-rhel8: unsupported: Request was throttled. 
Copying blob 272c29e2f36c skipped: already exists  
Copying blob 7ecdec58f172 skipped: already exists  
Copying blob 370fc4ac3f1f skipped: already exists  
Copying config 5a85baceb7 done  
WARN[0036] Failed, retrying in 16s ... (5/5). Error: copying system image from manifest list: writing manifest: uploading manifest latest to pulp3-source-fedora36.puffy.example.com/ee-minimal-rhel8: unsupported: Request was throttled. 
Copying blob 370fc4ac3f1f skipped: already exists  
Copying blob 272c29e2f36c skipped: already exists  
Copying blob 7ecdec58f172 skipped: already exists  
Copying config 5a85baceb7 done  
FATA[0056] copying system image from manifest list: writing manifest: uploading manifest latest to pulp3-source-fedora36.puffy.example.com/ee-minimal-rhel8: unsupported: Request was throttled. 
```

**Expected behavior**

**Additional context**
https://github.com/pulp/pulp_container/issues/1211 This behavior exacerbates the overall situation  which leads to additional tasks created.

lubosmj:
In this commit, ephemeral blobs and manifests are introduced to simplify
the concept of the push workflow from the Registry's perspective. The
ephemeral blobs are gradually added to the repository without committing.
A new repository version is created only after pushing the final tagged
manifest.

closes #1212


pulp_container/app/registry_api.py
```diff
@@ -738,15 +701,15 @@ def put(self, request, path, pk=None):
                 raise Exception("The digest did not match")
             try:
                 artifact = Artifact.init_and_validate(uploaded_file)
-                artifact.save()
+                artifact.touch()
```
mdellweg, line 704: I think the original way was correct. (This is creating a new artifact, no? Because if it already exists, we go down the ecxept path where touching is actually desired.)

pulp_container/app/registry_api.py
```diff
@@ -738,15 +701,15 @@ def put(self, request, path, pk=None):
...
             except IntegrityError:
                 artifact = Artifact.objects.get(sha256=artifact.sha256)
                 artifact.touch()
 
         with transaction.atomic():
             try:
                 blob = models.Blob(digest=digest)
-                blob.save()
+                blob.touch()
```
mdellweg, line 712: same here.

pulp_container/app/registry_api.py
lubosmj: So, are not we saving the object either way in this case?

https://github.com/pulp/pulpcore/blob/ad319b1963fd7462bbf9e5562ead263a42481db8/pulpcore/app/models/content.py#L389

pulp_container/app/registry_api.py
lubosmj: Okay, I can revert this.

pulp_container/app/registry_api.py
mdellweg: touch is limited to the timestamp field. I'm not sure how this behaves with new objects, but at least touch has a different intention that save.

lubosmj: Okay, thanks for the clarification. I will change it back.

pulp_container/app/registry_api.py
```diff
@@ -691,32 +667,19 @@ def put(self, request, path, pk=None):
...
-            # Upload has been deleted => task has started or even finished
-            task = Task.objects.filter(
-                name__endswith="add_and_remove",
-                reserved_resources_record__contains=[f"shared:upload:{pk}"],
-            ).last()
-            if not task:
-                # No upload and no task for it => the upload probably never existed
-                # return 404
+            try:
+                blob = repository.pending_blobs.get(digest=digest)
```
ipanova, line 677: i wonder if we should also look for the blob in the latest repo version. what about a situation when there 2 concurrent pushes of the same content and this line of code will hit in first push right when second push is removing blob from pending and adding those to repo?

pulp_container/app/registry_api.py
```diff
@@ -947,8 +914,10 @@ def put(self, request, path, pk=None):
 
             config_digest = config_layer.get("digest")
             try:
+                blobs_in_latest_version = repository.latest_version().content.values_list("pk")
```
ipanova, line 917: these 2 lines are repeated twice, you can re-use them and have just in one place, for example move them up right after config_digest line

pulp_container/app/registry_api.py
```diff
@@ -992,6 +963,8 @@ def put(self, request, path, pk=None):
                 objs=thru, ignore_conflicts=True, batch_size=1000
             )
 
+            add_content_units += [str(blob.pk) for blob in blobs_qs] + [str(config_blob.pk)]
+
         # a manifest cannot tagged by its digest - an identifier specified in the 'pk' parameter
         if not pk.startswith("sha256:"):
```
ipanova, line 969: I know that we agreed to first nail down the pending blobs, however, if i am not mistaken on the sequence of the manifest list API calls, it pushes children manifests by digest and not by tag - this way you could distinguish whether this PUT call is all about tagged image manifest or image manifest that is meant to be part of a list. Then you could add those to pending manifests and then clear out pending blob and manifests before creating the add_remove task.
Give it a try in a separate commit?

CHANGES/1212.bugfix
```diff
@@ -0,0 +1 @@
+Use pending blobs to not create unnecessary repository versions.
```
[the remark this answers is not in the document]
mdellweg: You know after this change we can think about the big repo merger again...

pulp_container/app/registry_api.py
mdellweg: Absolutely and we should do so first, i think.

pulp_container/app/registry_api.py
```diff
@@ -947,8 +914,10 @@ def put(self, request, path, pk=None):
 
             config_digest = config_layer.get("digest")
             try:
+                blobs_in_latest_version = repository.latest_version().content.values_list("pk")
+                pending_blobs = repository.pending_blobs.values_list("pk")
                 config_blob = models.Blob.objects.get(
-                    digest=config_digest, pk__in=repository.latest_version().content
+                    digest=config_digest, pk__in=chain(blobs_in_latest_version, pending_blobs)
```
mdellweg, line 920: Chain will evaluate these queries instead of creating subqueries. I think it should be
```python
available_blobs = repository.latest_version().content || repository.pending_blobs
config_blob = ... get(..., pk__in=avaliable_blobs.values_list("pk"))
```

CHANGES/1212.bugfix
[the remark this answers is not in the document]
ipanova: @mdellweg yes, looking forward to that :)

pulp_container/app/registry_api.py
lubosmj: You cannot do this in that way.

```
In [8]: Blob.objects.filter(pk__in=repo.pending_blobs || repo.latest_version().content)
  Cell In[8], line 1
    Blob.objects.filter(pk__in=repo.pending_blobs || repo.latest_version().content)
                                                   ^
SyntaxError: invalid syntax
```

```
----> 1 Blob.objects.filter(pk__in=repo.pending_blobs.all() | repo.latest_version().content)

File /usr/local/lib/python3.8/site-packages/django/db/models/query.py:345, in QuerySet.__or__(self, other)
    343 if not other.query.can_filter():
    344     other = other.model._base_manager.filter(pk__in=other.values('pk'))
--> 345 combined.query.combine(other.query, sql.OR)
    346 return combined

File /usr/local/lib/python3.8/site-packages/django/db/models/sql/query.py:586, in Query.combine(self, rhs, connector)
    577 def combine(self, rhs, connector):
    578     """
    579     Merge the 'rhs' query into the current one (with any 'rhs' effects
    580     being applied *after* (that is, "to the right of") anything in the
   (...)
    584     'rhs' query.
    585     """
--> 586     assert self.model == rhs.model, \
    587         "Cannot combine queries on two different base models."
    588     assert not self.is_sliced, \
    589         "Cannot combine queries once a slice has been taken."
    590     assert self.distinct == rhs.distinct, \
    591         "Cannot combine a unique query with a non-unique query."

AssertionError: Cannot combine queries on two different base models.
```

lubosmj: Ha, this helped:
```
In [15]: repo.latest_version().content.values_list("pk").union(repo.pending_blobs.values_list("pk"))
```

CHANGES/1212.bugfix
```diff
@@ -0,0 +1,2 @@
+Began to use ephemeral blobs when processing requests during the push operation to not create
+redundant repository versions.
```
mdellweg, line 2: I think it's not "redundant" but "unnecessary".

pulp_container/app/registry_api.py
mdellweg: That sounds great!

pulp_container/app/registry_api.py
lubosmj: I am doing it in one commit to reduce the number of separate migrations. Also, it is not that complex to implement and review it in one go.

lubosmj:
I introduced one change that needs attention. I made immediate tasks "deferrable". This will be useful in highly concurrent environments where users push images to repositories simultaneously. "Throttled" responses will be returned to users less frequently while DB calls will not be re-executed.

pulp_container/app/registry_api.py
```diff
@@ -689,34 +661,26 @@ def put(self, request, path, pk=None):
         _, repository = self.get_dr_push(request, path)
 
         digest = request.query_params["digest"]
-        # Try to see if the client came back after we told it to backoff with the ``Throttled``
-        # exception. In that case we answer based on the task state, or make it backoff again.
-        # This mechanism seems to work with podman but not with docker. However we let the task run
-        # anyway, since all clients will look with a HEAD request before attemting to upload a blob
-        # again.
+        # check if the client came back after the artifact was already uploaded; in that case, we
```
ipanova, line 664: you are raising here upload 404, client does not come back, it comes back only if you issue 429.

pulp_container/app/registry_api.py
```diff
@@ -936,7 +916,10 @@ def put(self, request, path, pk=None):
                 manifests_to_list, ignore_conflicts=True, batch_size=1000
             )
             manifest = manifest_list
+            config_blob = None
```
ipanova, line 919: initialize these variables before if/else

pulp_container/app/registry_api.py
```diff
@@ -977,12 +953,18 @@ def put(self, request, path, pk=None):
                 digest = layer.get("digest")
                 blobs.add(digest)
 
-            blobs_qs = models.Blob.objects.filter(
-                digest__in=blobs, pk__in=repository.latest_version().content
-            )
+            pending_blobs = repository.pending_blobs.values_list("pk")
+            blobs_pks = content_in_latest_version.union(pending_blobs)
```
ipanova, line 957: this one can be replaced with `content_pks` i mentioned earlier

pulp_container/app/registry_api.py
```diff
@@ -977,12 +953,18 @@ def put(self, request, path, pk=None):
...
+
+            blobs_qs = models.Blob.objects.filter(digest__in=blobs, pk__in=blobs_pks)
```
ipanova, line 959: please keep the naming consistent and maybe consider using `found_blobs`

pulp_container/app/registry_api.py
```diff
@@ -1001,26 +985,58 @@ def put(self, request, path, pk=None):
                 tag = models.Tag.objects.get(name=tag.name, tagged_manifest=manifest)
                 tag.touch()
 
+            add_content_units += [str(tag.pk), str(manifest.pk)]
+
+            if is_manifest_list:
```
ipanova, line 990: please add here else branch for image manifest

ipanova (commented):
not all comments might be entirely right but  the code is very confusing

ipanova:
> I introduced one change that needs attention. I made immediate tasks "deferrable". This will be useful in highly concurrent environments where users push images to repositories simultaneously. "Throttled" responses will be returned to users less frequently while DB calls will not be re-executed.

i think i am ok with defferable as i was suggesting here https://github.com/pulp/pulp_container/pull/1249#discussion_r1157531411

pulp_container/app/registry_api.py
```diff
@@ -1001,26 +985,58 @@ def put(self, request, path, pk=None):
...
+                    "remove_content_units": remove_content_units,
+                },
+                immediate=True,
+                deferred=True,
+            )
 
-        if has_task_completed(dispatched_task):
+            if immediate_task.state == "completed":
+                if found_manifests:
+                    repository.pending_manifests.remove(*found_manifests)
```
ipanova, line 1021: if you make this task deferrable and it will not complete within the allocated 3 secs waiting time, it will then be passed to worker. who will make sure the pending content is removed from this list? we should probably first remove and then dispatch the task, otherwise once the task is picked up by worker and it completes there will be noone who'd remove pending content

lubosmj: Good catch. I had not thought about that.

lubosmj: In the end, it would be an orphan cleanup that will handle this situation. But, I still think removing it beforehand is much better.

pulp_container/app/registry_api.py
```diff
@@ -575,7 +575,7 @@ def create(self, request, path):
     @staticmethod
     def tries_to_mount_blob(request):
         """Check if a client is trying to perform cross repository blob mounting."""
-        return (request.query_params.keys()) == {"from", "mount"}
+        return request.query_params.keys() == {"from", "mount"}
```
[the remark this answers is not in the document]
lubosmj: We need to fix it. I opened this issue after skimming through the code: https://github.com/pulp/pulp_container/issues/1286.

pulp_container/app/registry_api.py
lubosmj: Yeah, it makes sense now to move things like "add_content_units += [str(blob.pk) for blob in blobs_qs] + [str(config_blob.pk)]" here then.

pulp_container/app/registry_api.py
```diff
@@ -1001,26 +985,58 @@ def put(self, request, path, pk=None):
...
+
+            if is_manifest_list:
+                if not pending_blobs:
+                    pending_blobs = repository.pending_blobs.values_list("pk")
+
+                blobs_qs = models.Blob.objects.filter(
+                    pk__in=pending_blobs.union(found_manifests.values_list("blobs__pk"))
+                )
+                # we need to add pending blobs referenced by pending manifests
+                add_content_units += [str(blob.pk) for blob in blobs_qs]
```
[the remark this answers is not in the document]
lubosmj: Config blobs should be included within the qs since we were adding them one by one to the `pending` list. I believe I can even remove "[str(config_blob.pk)]" from "add_content_units += [str(blob.pk) for blob in blobs_qs] + [str(config_blob.pk)]".

pulp_container/app/registry_api.py
ipanova: definitely, also because orphan clean up is not regularly running task and god who know when it will be triggered

lubosmj: @mdellweg, do you suggest calling `repository.pending_[manifests|blobs].clear()` in the `finalize_new_version` hook? Can we rule out any problems with removing pending manifests/blobs when other users are pushing data to the same repository?

mdellweg: No, of course we should only remove the ones we added in that repository version.
Concurrent pushes is definitely a thing to consider.

pulp_container/app/registry_api.py
```diff
@@ -689,34 +662,11 @@ def put(self, request, path, pk=None):
...
-                blob = models.Blob.objects.get(digest=digest)
-                return BlobResponse(blob, path, 201, request)
-            elif task.state in ["waiting", "running"]:
-                raise Throttled()
-            else:
-                error = task.error
-                task.delete()
-                raise Exception(str(error))
+            # return 404
+            raise e_upload
```
mdellweg, line 669: Why reraising it now? I think the try block is not useful anymore now.

lubosmj: Would calling https://docs.djangoproject.com/en/4.2/topics/http/shortcuts/#get-object-or-404 be more favourable?

pulp_container/tests/functional/api/test_rbac_repo_versions.py
```diff
@@ -217,25 +214,13 @@ def test_cross_repository_blob_mount(
     # Cleanup
     add_to_cleanup(container_namespace_api, distribution.namespace)
 
-    # Test that admin can cross mount
-    for i, blob in enumerate(blobs, start=1):
+    # Test that an admin can cross mount but do not perform any further assertions because
+    # the blobs are committed only after uploading the final tagged manifest
```
mdellweg, line 218: I think you should be able to do a get on the mounted blob, as it's now considered to be part of the target repo.

mdellweg (approved):
One suggestion for a test, if you like.

pulp_container/tests/functional/api/test_rbac_repo_versions.py
lubosmj: And, I have found out I need to update the registry.py code because we query `ContentArtifact` against "repository_version.get_content" only (See the `get_by_digest` method).

pulp_container/app/registry_api.py
```diff
@@ -947,9 +914,7 @@ def put(self, request, path, pk=None):
 
             config_digest = config_layer.get("digest")
             try:
-                config_blob = models.Blob.objects.get(
-                    digest=config_digest, pk__in=repository.latest_version().content
-                )
+                config_blob = models.Blob.objects.get(digest=config_digest, pk__in=content_pks)
```
ipanova, line 917: how about you change this to:
config_blob_qs =  models.Blob.objects.filter(digest=config_digest, pk__in=content_pks)
if not config_blob_qs.exists():
    raise BlobInvalid

and reuse this qs by extending found_blobs after line 993

pulp_container/app/registry_api.py
```diff
@@ -1001,26 +964,45 @@ def put(self, request, path, pk=None):
                 tag = models.Tag.objects.get(name=tag.name, tagged_manifest=manifest)
                 tag.touch()
 
+            add_content_units = [str(tag.pk), str(manifest.pk)]
+
+            if is_manifest_list:
+                add_content_units.extend(
```
ipanova, line 970: you could get rid of if/else and purely have `[str(content.pk) for content in chain(found_blobs, found_manifests)]`

if you include in the image manifest `found_blobs` variable also the config blob.

pulp_container/tests/functional/api/test_rbac_repo_versions.py
ipanova: I don't think this is correct. You are not supposed to stream the content to the client if it is not fully committed yet to the repo which happens in the finalize_new_version.
It is ok and correct to give a 200 on handle_safe_methods though.

pulp_container/app/models.py
```diff
@@ -550,6 +552,15 @@ def finalize_new_version(self, new_version):
         """
         remove_duplicates(new_version)
         validate_repo_version(new_version)
+        self.remove_pending_content(new_version)
+
+    def remove_pending_content(self, repository_version):
+        """Remove pending blobs and manifests when committing the content to the repository."""
+        added_content = repository_version.added(
```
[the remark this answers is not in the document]
lubosmj: The method is included only within the push repository model. No need to check that once again.

CHANGES/1212.bugfix
```diff
@@ -0,0 +1,2 @@
+Enhanced push operation efficiency by implementing the utilization of ephemeral blobs and
```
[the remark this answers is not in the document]
lubosmj: Yes!

ipanova (approved):
:tada:

PR #1291, opened 2023-05-26
Remove the ansible installation section

lubosmj:
[noissue]

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -2,28 +2,11 @@
 User Setup
 ==========
 
-Ansible Installer (Recommended)
--------------------------------
+Containerized Installation
+**************************
 
-We recommend that you install `pulpcore` and `pulp-container` together using the `Ansible installer
-<https://github.com/pulp/pulp_installer/blob/master/README.md>`_. If you install this way, pulpcore
-installation and all the following steps will be done for you.
-
-Install ``pulpcore``
---------------------
-
-Follow the `installation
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -2,28 +2,11 @@
 User Setup
 ==========
 
-Ansible Installer (Recommended)
--------------------------------
+From OCI Images
```
ipanova, line 5: Containerized installation

docs/installation.rst
```diff
@@ -2,28 +2,11 @@
...
-
-Install plugin
---------------
-
-This document assumes that you have
-`installed pulpcore <https://docs.pulpproject.org/installation/instructions.html>`_
-into a the virtual environment ``pulpvenv``.
-
-Users should install from **either** PyPI or source.
+Follow the `podman compose instructions <https://pulpproject.org/podman-compose>`_ to deploy Pulp by
```
ipanova, line 8: I would not mention compose as the first thing here. Yes it does leverage oci images but this it is intended as a proof-of-concept, because user will likely need to modify the Compose file based on his needs.

Out of the box ready to use installations is whether single container or pulp-operator

ipanova (commented):
I remember one of the ideas were to create self-contained plugin docs. In this particular case, to me if feels like having the installation docs in pulpcore and point from plugins to it is best solution https://docs.pulpproject.org/pulpcore/installation/instructions.html

ipanova (approved):

Issue #507, opened 2021-12-15
As a user I can pull-through cache container images when remote is defined on distribution


pulpbot:
Author: @ipanova (ipanova@redhat.com)

Redmine Issue: 9560, https://pulp.plan.io/issues/9560

---

see pulp_python PR for more details https://github.com/pulp/pulp_python/pull/384/files
pulp_container won't have this as easy because of multiple content types needed to be downloaded as well as relations created between them

lubosmj:
In **Proposal 2**, we will enable a user to download/forward/stream content from a remote registry considering no matter which repository she is talking to.

Considering that, users could (intentionally) flood Pulp and make it unavailable for some time. It might be convenient to take a look at whitelisting upstreams, as proposed in https://github.com/pulp/pulp_container/issues/459.

ipanova:
We will revisit this once it becomes a higher priority

benedikt-bartscher:
This would be awesome to easily prevent dockerhub rate-limits. Is there anything the community could do to code/test/sponsor this?

ipanova:
@benedikt-bartscher thank you for showing your interest in this feature. There is already a mechanism in place in pulp-container to prevent dockerhub rate-limits. It would consist in creating and mirroring repo locally with on_deman policy  beforehand of client pull.
This feature specifically, would remove the need of creating repo on pulp side in advance, pulp would transparently create all needed objects based on the incoming client pull request.
Would existing workflow cover your current needs?

benedikt-bartscher:
Hey @ipanova thanks for you reply. I know about that mechanism, i am currently using it. However it's not very convenient to setup every repo manually, thats why i pushed this issue.

ipanova:
@benedikt-bartscher gotcha, yeah that's the inconvenience compared to the pull-through cache. We will try to resume  the work on this, there are some design challenges we need to wrap the head around ;)

ipanova:
https://github.com/pulp/pulp_container/issues/732

mdellweg:
Braindump of an idea:
Can we use a registry subdirectory (maybe a namespace) (like `/quay-pull-through/`) to attach a PullThroughRemote to it that will have all the necessary configuration to create all the requested repositories with remotes and distributions?
Having it be a namespace would allow to sort out the RBAC stuff for it.
Related question: Is is possible to configure `registry.json` in a way that a user issuing `podman pull busybox` will do the equivalent of `podman pull pulp-server.io/quay-pull-through/busybox`?

[...]

Issue #1268, opened 2023-04-21
As an admin/user I can maintain a controlled registry that doesn't continually grow in space


ipanova:
**Is your feature request related to a problem? Please describe.**
While for synced repos there is an option to sync content in a mirrored mode( not just additive), and then trigger orphan cleanup(still half solution). For pushed repos the content always gets accumulated.

**Describe the solution you'd like**
Introduce a concept of image garbage collection. Garbage collection will ensure efficient use of resources for active objects by removing objects that occupy sizeable amounts of disk space

**Describe alternatives you've considered**
1. set retain_repo_versions  to 1 on the push repo
2. manually remove undesired images by digest. It will delete its tags and blobs(in case they are not referenced in other images)
3. Trigger orphan cleanup. Might need to adjust orphan_protection_time based on needs.

For synced repos one can also consider and combine the disk reclamation feature in case repo versioning is needed.

**Additional context**
Design and criteria of garbage collection - TBD
Disclaimer - this feature might be incompatible with repo versions, and might require, or will take into account to garbage collect only those repos that do not have versioning enabled aka repo_version_retention=1

mdellweg:
When we implement #1212, I think we can move through all existing repository_versions and at least remove all the dangling blobs that got uploaded, but never added to a manifest. I know it is not exactly what this ticket is talking about, but also somehow related.

lubosmj:
I imagine having an endpoint that will accept a list of repositories/tags (include/exclude_list) which should be preserved, and everything else will be purged.

Also, we can introduce the concept of "stale repositories". Repositories will be flagged as *stale* if they have not been hit by container clients for a longer period of time. Then, adjusted orphan cleanup will remove these repositories along with their content.

ipanova:
ideas on tag retention policy   https://goharbor.io/docs/2.2.0/working-with-projects/working-with-images/create-tag-retention-rules/

lubosmj:
Nice find.

Based on the above comment, we can consider scheduling periodic cleanup tasks. We already have a facility for dispatching the periodic analytics task:
https://github.com/pulp/pulpcore/blob/b5dbcc0fda6a8b46b1ef8efe73cf71056fb67017/pulpcore/app/util.py#L298

From the scheduled task, we can traverse repositories or Pulp settings. If we allow users to set up a retention configuration that attaches to specific repositories, we could make the garbage collection more granular.

git-hyagi:
Some notes/thoughts about this feature:
[...]

Issue #459, opened 2021-12-15
Feature request: whitelist upstreams


pulpbot:
Author: joeydumont (joeydumont)

Redmine Issue: 5545, https://pulp.plan.io/issues/5545

---

I open this feature request after a short conversation on the mailing list: see the "pulp_docker: whitelist upstreams" thread.

I would like to be able to provide a list of whitelisted upstreams and create a single repository version with all of these upstreams. For example, I would like for users to be able to do:

~~~text
docker pull my_docker_distribution/alpine/git
docker pull my_docker_distribution/postgresql
~~~

but not

~~~text
docker pull my_docker_distribution/super_shady_container_image_hosted_on_Docker_Hub
~~~

where the underlying repositories would point to Docker Hub, or other container registries.

Thank you for any information!

lubosmj:
For the sake of consistency, we may want to create fields "includes" and "excludes" on remotes as we did in pulp_gem.

https://github.com/pulp/pulp_gem/blob/4e72167769f1a6ce11b4119b8a609d138e9e9fe2/docs/workflows/sync.rst?plain=1#L69

PR #1299, opened 2023-06-04
Add Pull-through caching

lubosmj:
- [x] Devise a workflow for adding content to a single repository version (adding content one by one will not follow the planned path for having repositories with consolidated repository versions)
- [x] Write a functional test (creating a pull-through cache remote and distribution and pulling content via the Pulp Container Registry)
- [x] Bump up the pulpcore requirement correspondingly because of migrations
- [x] Clean up the code (refactoring the code to remove duplicates and speeding up the operation)

closes https://github.com/pulp/pulp_container/issues/507


lubosmj:
@ipanova, @mdellweg, would you mind reviewing this PR? Focus on the underlying logic.

Things to consider:
1. The pull-through cache logic is placed exclusively in the live API (within a synchronous context). I could not lock a repository from the content app.
2. Repository/distributions' names/base_paths are created from remotes' upstream names. This might cause problems with existing repository/distributions' names/base_paths.
3. Similarly to the push workflow, I adopted the concept of pending blobs/manifests and extended it to classic repositories. The content is added to a repository only when the latest remaining blob from the manifest's layers is requested to prevent a situation in which the content is scattered in multiple repository versions.
4. In some cases (when some manifest layers are stored in the local storage), `podman` does not pull all blobs from Pulp. Because of this, we are not committing the content to the repository since we are waiting for a user to request it first. Is this fine? I thought of removing the need for adding content to the repository. In pulp-to-pulp sync scenarios where one Pulp instance acts as a master (with enabled pull-through-caching), the content will be committed to the repository because the master will serve it to another instance.

lubosmj:
Oh, the reason why I did not include sending head requests beforehand is that "docker-content-digest" is not a required header and is probably not present in other non-docker registries.

mdellweg:
> 1. The pull-through cache logic is placed exclusively in the live API (within a synchronous context). I could not lock a repository from the content app.

Can't you dispatch the add_content task from the content app? In the end, the content app does not even need to wait for it to finish, right?

mdellweg:
Sorry, wrong button

lubosmj:
I believe the problem was due to the asynchronous context. I could not dispatch the task because of it.

lubosmj:
@ipanova and I concluded that we should preserve the idea of adding content to a repository (the exact opposite of what we are doing in other plugins). The 4th bullet point is no longer a concern if we assume that there is a user who does not have cached layers on his system and will eventually download all pending blobs (this leads to committing the repository version). Repositories/distributions, created from special distributions, will be visible to users because we allow the pull operation.

Besides that, we identified two flaws in the current implementation:
1. We need to address garbage collection. At this time, it is sufficient to mimic the behaviour of `mirror=True` synced repositories (add_and_remove - delete content which is not needed + `retain_repo_version=1`). Later we can handle the garbage collection with: https://github.com/pulp/pulp_container/issues/1268.
2. The name of a distribution and a repository should have the following format to mitigate the conflicts: {special_cache_distribution_base_path}/{upstream_name}

Things to work on next:
- [ ] delete manifests and blobs from a caching repository, set `retain_repo_version=1`
- [ ] update the base_path/name initialization for distribution/remote/repository in `get_pull_through_drv`
- [ ] issue head/get requests to fetch the latest version of cached content
- [ ] add support for namespaced caching (allow caching a specific organization besides the whole registry)

lubosmj:
Namespace caching works out of the box without more changes:
```
http :5001/pulp/api/v3/distributions/container/pull-through/ name=library base_path=docker-cache/library remote=${PULL_THROUGH_REMOTE_HREF}
```
```
[lmjachky@localhost pulp_container]$ podman pull localhost:5001/docker-cache/library/busybox:latest --tls-verify=false
Trying to pull localhost:5001/docker-cache/library/busybox:latest...
Copying blob 3f4d90098f5b skipped: already exists  
Copying config a416a98b71 done  
Storing signatures
a416a98b71e224a31ee99cff8e16063554498227d2b696152a9c3e0aa65e5824
[lmjachky@localhost pulp_container]$ podman pull localhost:5001/docker-cache/libraru/busybox:latest --tls-verify=falseTrying to pull localhost:5001/docker-cache/libraru/busybox:latest...
Error: initializing source docker://localhost:5001/docker-cache/libraru/busybox:latest: reading manifest latest in localhost:5001/docker-cache/libraru/busybox: name unknown: Repository not found.
```

pulp_container/app/models.py
```diff
@@ -565,6 +642,25 @@ def remove_pending_content(self, repository_version):
         self.pending_manifests.remove(*Manifest.objects.filter(pk__in=added_content))
 
 
+class ContainerPullThroughDistribution(Distribution, AutoAddObjPermsMixin):
+    """
+    TODO: Add permissions.
+    """
```
lubosmj, line 648: May I use the same permissions that are currently used by `ContainerDistribution`? The result would look like this:
```
        permissions = [
            ("pull_containerdistribution", "Can pull from a registry repo"),
            (
                "manage_roles_containerdistribution",
                "Can manage role assignments on container distribution",
            ),
        ]
```

lubosmj: Same stands for `ContainerRemote`:
```
        permissions = [
            (
                "manage_roles_containerremote",
                "Can manage role assignments on container remote",
            ),
        ]
```

pulp_container/app/registry_api.py
```diff
@@ -272,7 +275,8 @@ def get_drv_pull(self, path):
         try:
             distribution = models.ContainerDistribution.objects.get(base_path=path)
         except models.ContainerDistribution.DoesNotExist:
-            raise RepositoryNotFound(name=path)
+            # get a pull-through cache distribution whose base_path is a substring of path
+            return self.get_pull_through_drv(path)
```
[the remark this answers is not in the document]
lubosmj: Another question is whether we want to use Redis caching in the first place. I am open to enhancing handle_safe_method() to detect if we are doing pull-through caching. If not, the decorator will remain; otherwise, we will proceed to real-time pull-through caching.

pulp_container/app/models.py
lubosmj: And, of course, the same for ViewSets.

lubosmj:
Additionally, I can resolve this issue: https://github.com/pulp/pulp_container/issues/459. Whitelisting upstreams might be worth it since all repositories can be downloaded from the remote registry. I will add a new optional parameter to either a remote or distribution with a list of blacklisted or whitelisted upstreams.

mdellweg:
> Additionally, I can resolve this issue: #459. Whitelisting upstreams might be worth it since all repositories can be downloaded from the remote registry. I will add a new optional parameter to either a remote or distribution with a list of blacklisted or whitelisted upstreams.

In pulp_gem/python we call these includes and excludes. I'd prefer them to be consistently named.

pulp_container/app/serializers.py
```diff
@@ -309,10 +325,16 @@ class ContainerDistributionSerializer(DistributionSerializer):
     repository_version = RepositoryVersionRelatedField(
         required=False, help_text=_("RepositoryVersion to be served"), allow_null=True
     )
+    remote = DetailRelatedField(
+        required=False,
+        help_text=_("Remote that can be used to fetch content when using pull-through caching."),
+        view_name_pattern=r"remotes(-.*/.*)?-detail",
+        queryset=models.ContainerRemote.objects.all(),
+    )
```
mdellweg, line 333: Should this be limited to pull-through remotes?
Is this even applicable to standard container distributions?

pulp_container/app/serializers.py
```diff
@@ -309,10 +325,16 @@ class ContainerDistributionSerializer(DistributionSerializer):
...
 
     def validate(self, data):
         """
-        Validate the ContainterDistribution.
+        Validate the ContainerDistribution.
```
mdellweg, line 337: :stuck_out_tongue_closed_eyes:

docs/workflows/host.rst
```diff
@@ -117,3 +117,36 @@ Docker Output::
...
+a transparent caching layer.
+
+By incorporating the Pull-Through Caching feature, administrators can **reduce external network
+dependencies**, and ensure a more reliable and responsive container deployment system in production
+environments.
+
+.. note::
+    Pulp automatically creates a new repository-remote-distribution triplet per every image pulled
+    by a user. The created repository retains a single repository version containing only the latest
+    image data. The said entities can be managed by standard API endpoints.
```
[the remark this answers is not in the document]
lubosmj: If you pull `debian:10` first, you should get a new repository `debian` and a new tagged manifest `10`. If you pull `debian:11`, there should be a new tagged manifest `11` added to the existing repository `debian`.

pulp_container/app/downloaders.py
```diff
@@ -95,14 +100,47 @@ async def _run(self, handle_401=True, extra_data=None):
...
+
+        Returns:
+            :class:`~pulpcore.plugin.download.DownloadResult`
+                or :class:`~aiohttp.client.ClientResponse`
+
+        Raises:
+            Exception: Any fatal exception emitted during downloading
+        """
+        return asyncio.get_event_loop().run_until_complete(self.run(extra_data=extra_data))
+
```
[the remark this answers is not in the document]
lubosmj: `fetch()` was copied here for the sake of CI because its updated version was not yet released back then.

Concerning `_ensure_writer_has_open_file()` (original version here: https://github.com/pulp/pulpcore/blob/3cb7ef4369fd612d17e2be7863845f6cc74c141f/pulpcore/download/base.py#L112), Pulp is by default downloading content from api-app to an inaccessible folder. Visit the docstring for more details.

pulp_container/app/models.py
```diff
@@ -487,6 +547,13 @@ class ContainerRepository(
         ManifestSigningService, on_delete=models.SET_NULL, null=True
     )
 
+    # temporary relations used for uncommitted pull-through cache operations
+    pending_tags = models.ManyToManyField(Tag)
+    pending_manifests = models.ManyToManyField(Manifest)
+    pending_blobs = models.ManyToManyField(Blob, related_name="pending_blobs")
+    # digests of remaining blobs to be attached to pending manifests
+    remaining_blobs = models.ManyToManyField(Blob, related_name="remaining_blobs")
```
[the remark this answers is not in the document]
lubosmj: While downloading/caching layers from a manifest, we track how many of them were already downloaded and how many are still missing. Once there are no remaining blobs to download, we commit the whole box of content, consisting of a tag, manifest, and blobs, to a new repository version: https://github.com/lubosmj/pulp_container/blob/f6e057f3ae5c44868155cb5657daf072b24f9ec4/pulp_container/app/registry_api.py#L982.

This relation was introduced to omit file access to the manifest's artifact each time an associated blob is downloaded. We read the manifest's artifact once at the beginning of the pull-through caching and then record all blobs that will be downloaded next (tag -> manifest -> blob1,blob2,blob3). During the download, we do not have a manifest object containing all FK references to layers/blobs. We build this relation just before committing the content to the repository, when everything is downloaded in place.

If we make the Manifest model artifactless, the relation will be redundant.

pulp_container/app/serializers.py
lubosmj: This remote is a standard remote we are aware of in the pulp_container ecosystem. Under the hood, this remote contains `remote.url = pull_through_remote.url` and `remote.upstream_name = requested_upstream_name`. The remote is attached to a newly created distribution and it is used for pull-through cache downloading (https://github.com/pulp/pulp_container/blob/f6e057f3ae5c44868155cb5657daf072b24f9ec4/pulp_container/app/registry_api.py#L1569).

docs/workflows/host.rst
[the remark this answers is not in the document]
lubosmj: Regarding the latter question, Pulp is supposed to create a new repository version on top of the existing repository if the tag points to a manifest with a different digest. But, I should test it first somewhere. 👁️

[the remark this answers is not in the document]
mdellweg: This all matches my expectations, great. The text here, however left me confused.

pulp_container/app/downloaders.py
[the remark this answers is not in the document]
mdellweg: What is wrong with always using dir_path="."? The task should have cd'd into it's working directory (btw a subdirectory of settings.WORKING_DIRECTORY).

pulp_container/app/models.py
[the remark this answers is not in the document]
mdellweg: We are not going to make manifests artifact-less, are we?
Maybe getting ahead of myself now, but do we need ephemeral manifests that blobs can link to, but that cannot be added to repositories before being converted into real ones? (A bit like how an Upload is an ephemeral/incomplete Artifact.)

pulp_container/app/downloaders.py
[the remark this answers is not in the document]
lubosmj: We download the content from api-app. The task is not involved in pull-through caching. This is what I see without the change, just using the pulpcore's facility:
```
  File "/src/pulpcore/pulpcore/download/http.py", line 258, in download_wrapper
    return await self._run(extra_data=extra_data)
  File "/src/pulp_container/pulp_container/app/downloaders.py", line 95, in _run
    return await self._run(handle_401=False, extra_data=extra_data)
  File "/src/pulp_container/pulp_container/app/downloaders.py", line 104, in _run
    to_return = await self._handle_response(response)
  File "/src/pulpcore/pulpcore/download/http.py", line 209, in _handle_response
    await self.finalize()
  File "/src/pulpcore/pulpcore/download/base.py", line 166, in finalize
    self._ensure_writer_has_open_file()
  File "/src/pulpcore/pulpcore/download/base.py", line 130, in _ensure_writer_has_open_file
    self._writer = tempfile.NamedTemporaryFile(dir=".", suffix=suffix, delete=False)
  File "/usr/lib64/python3.8/tempfile.py", line 540, in NamedTemporaryFile
    (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
  File "/usr/lib64/python3.8/tempfile.py", line 250, in _mkstemp_inner
    fd = _os.open(file, flags, 0o600)
PermissionError: [Errno 13] Permission denied: './tmpg9hy1dqd-latest'
```

pulp_container/app/models.py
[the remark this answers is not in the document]
lubosmj: We are going to make the manifests artifact-less soon-ish. By using real "ephemeral" manifests, we re-use what is already provided, right? The problem could be the migration that should be managed without downtime. Do you think having the `ManifestUpload` model is the way to go?

pulp_container/app/downloaders.py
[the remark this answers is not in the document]
mdellweg: Ah, i misread the logic. So you are saying the api-worker needs to start writing stuff to disk? This opens an new pandora box... (and probably some repercussions with SELinux)
Is there any reasonable way to move this out to the content-app?

pulp_container/app/models.py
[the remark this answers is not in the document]
mdellweg: oh, artifactless is different than Blob less, right?
Can we still state that a Manifest can only be added to the database with all the blob relations it will ever have in the same transaction?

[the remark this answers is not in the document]
mdellweg: I believe we cannot do that. The sync pipeline used to do it and we got into a lot of trouble with failed, incomplete syncs. Adding a blob to a manifest is like modifying it (while we guarantee unmutability).

The scenario i know that can go wrong is: A user adds an unfinished Manifest to their repository, and the task will carry only the by then associated blobs with it, leading to an invalid repository version. And the best solution I see is to prevent unfinished Manifests to exist in the first place.

pulp_container/app/downloaders.py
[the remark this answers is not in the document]
lubosmj: The problem with dispatching tasks from content-app is gone after calling `set_guid(generate_guid())`. Therefore, we can revive the idea of streaming the content from content-app directly. With this, we will likely not store the content within a repository. This is the subject for discussion.

Attaching the suggestion about a task that will handle blobs downloading in the background:
```
That task will take the artifact, read it, draft on-demand blobs from it and creates and adds a manifest to the repository. (The Manifest having healthy relationships to its blobs.)
In the meantime the user will probably download a certain number of the blobs in that manifest.
Again, pulp-content will stream download them to the client if not yet in pulp, and (maybe) create ephemeral blobs from them.
Once the task ran, we now have on-demand blobs, and the serving goes through the old channels of pulp.
```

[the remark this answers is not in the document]
mdellweg: You see me confused. The fact that we can dispatch that task means that we _can_ add the content from requests solely talking to the content_app.

pulp_container/app/cache.py
```diff
@@ -71,7 +72,15 @@ def find_base_path_cached(request, cached):
                 "repository", "repository_version"
```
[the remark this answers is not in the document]
lubosmj: But, is not the returned queryset cached by Django in this case? Then, if we access `distribution.repository_version` from `get_drv_pull`, it is already there for us (https://github.com/pulp/pulp_container/blob/47ee11184b4dbaa73cab62e56111dea27d970235/pulp_container/app/registry_api.py#L277). This is just an assumption. I can remove it if you insist.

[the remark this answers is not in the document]
ipanova: `find_base_path_cached` returns a path, it does not return whole distribution object so i don't understand why the query would use select_related if  in the end only base_path is returned https://github.com/pulp/pulp_container/blob/main/pulp_container/app/cache.py#L52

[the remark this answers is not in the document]
ipanova: This is no really related to your PR but just hit my eye while looking at it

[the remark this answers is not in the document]
lubosmj: Ah, gotcha. I thought the result is cached somewhere by django and re-used once there is another request for the same entity from another function running within the same process.

pulp_container/app/models.py
```diff
@@ -595,6 +691,13 @@ class ContainerDistribution(Distribution, AutoAddObjPermsMixin):
     )
     description = models.TextField(null=True)
 
+    pull_through_distribution = models.ForeignKey(
+        ContainerPullThroughDistribution,
+        related_name="distributions",
+        on_delete=models.CASCADE,
+        null=True,
+    )
```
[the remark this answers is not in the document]
lubosmj: How do you imagine having permissions assigned to these objects? Users who are in charge of pull-through cache distributions should be able to preview the created subdistributions. Are you suggesting adjusting default roles and viewsets in a manner that will disallow the users from seeing specific distributions (possible) or getting to pull-through cache distributions from reverse relation (not possible from API)? I would leave the permissions untouched as we have between Tasks and TaskGroups.

pulp_container/app/downloaders.py
```diff
@@ -24,13 +27,14 @@ class RegistryAuthHttpDownloader(HttpDownloader):
...
 
     registry_auth = {"bearer": None, "basic": None}
-    token_lock = asyncio.Lock()
 
     def __init__(self, *args, **kwargs):
         """
         Initialize the downloader.
         """
         self.remote = kwargs.pop("remote")
+        self.token_lock = asyncio.Lock()
```
lubosmj, line 36: This change was made because of this: https://github.com/pulp/pulp_container/actions/runs/6686219325/job/18165335541#step:12:2059. Not sure what is the performance impact.

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
lubosmj: We decided to leave the caching in place. Shortly, we will remove this since TTL is a globally configurable option that might negatively impact the feature.

pulp_container/app/serializers.py
lubosmj: Regarding the testing, this is limited to the distributions/remotes created during the initialization of pull-through caching entities in `get_pull_through_drv::if not root_cache_distribution`. Thus, we do not need to worry about standard distributions downloading remote data from attached remotes.

ipanova:
From @lubosmj , the following procedure is implemented:

1.     a user requests new content via pull-through
2.     Pulp checks if the digest was updated on the remote with a HEAD request, returns/redirects cached content if not, or issues a redirect directly to content-app
3.     Pulp checks in content-app if there is a distribution with a remote it can download content from, downloads a manifest by tag in memory, dispatches a task to initialize related content (like blobs), and returns the downloaded in-memory manifest to the user
4.     listed manifests and blobs are then streamed to the user as podman/docker asks for

lubosmj:
@ipanova, @mdellweg, does it make sense to flag this feature as tech-preview? We can change the Redis caching effect soon and perhaps overhaul some stuff once we have artifactless manifests.

ipanova:
+1 to tech-preview. this is quite a big feature and might take some time until we stabilize it.

pulp_container/app/serializers.py
```diff
@@ -277,6 +277,22 @@ class Meta:
         model = models.ContainerRemote
 
 
+class ContainerPullThroughRemoteSerializer(RemoteSerializer):
```
ipanova, line 280: we should do something against cross type usage e.g.  container repo + remote pull-through

pulp_container/app/serializers.py
```diff
@@ -360,12 +382,36 @@ class Meta:
         fields = tuple(set(DistributionSerializer.Meta.fields) - {"base_url"}) + (
             "repository_version",
             "registry_path",
+            "remote",
```
ipanova, line 385: why it needs remote?

pulp_container/app/serializers.py
```diff
@@ -360,12 +382,36 @@ class Meta:
...
+        many=True,
+        help_text="Distributions created after pulling content through cache",
+        view_name="distributions-detail",
+        queryset=models.ContainerDistribution.objects.all(),
+        required=False,
+    )
+
+    class Meta:
+        model = models.ContainerPullThroughDistribution
+        fields = DistributionSerializer.Meta.fields + ("remote", "distributions")
```
ipanova, line 412: we do not need base_url here as well

pulp_container/app/registry.py
```diff
@@ -117,7 +130,67 @@ async def get_tag(self, request):
...
+                    kwargs={
+                        "repository_pk": repository_version.repository.pk,
+                        "remote_pk": remote.pk,
+                        "tag_name": tag_name,
+                        "response_data": response.data,
+                    },
+                )
+
+                # waiting shortly for the task to be completed since a container client could
+                # request related content (i.e., manifests and blobs) and halt the pull operation
```
ipanova, line 156: i am not sure this is good practice during pull operation.. did you test if the client backs off if you raise 429? or it just fails?

docs/workflows/host.rst
```diff
@@ -117,3 +117,38 @@ Docker Output::
...
+    http ${BASE_ADDR}/pulp/api/v3/distributions/container/pull-through/ remote=${REMOTE_HREF} name=docker-cache base_path=docker-cache
+
+Downloading content::
+
+    podman pull localhost:24817/docker-cache/library/busybox
+
+In the example above, the image "busybox" is pulled from the "docker-cache" distribution, acting as
+a transparent caching layer.
+
+By incorporating the Pull-Through Caching feature, administrators can **reduce external network
```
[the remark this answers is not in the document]
lubosmj: Yes, I agree with you to explicitly state that pre-configuring each repository is no longer needed with this.

docs/workflows/host.rst
```diff
@@ -117,3 +117,38 @@ Docker Output::
...
+dependencies**, and ensure a more reliable and responsive container deployment system in production
```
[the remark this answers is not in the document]
lubosmj: When saying reliable, I mean that once the content is in your local server, your data are more available than in a remote registry that might have some outages in the future and you cannot do anything with it.

pulp_container/app/models.py
```diff
@@ -565,6 +688,23 @@ def remove_pending_content(self, repository_version):
         self.pending_manifests.remove(*Manifest.objects.filter(pk__in=added_content))
 
 
+class ContainerPullThroughDistribution(Distribution, AutoAddObjPermsMixin):
+    """
+    A distribution for pull-through caching, referencing normal distributions.
+    """
+
+    TYPE = "pull-through"
```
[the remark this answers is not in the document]
lubosmj: Good catch. I did not consider this.

pulp_container/app/serializers.py
```diff
@@ -309,10 +325,16 @@ class ContainerDistributionSerializer(DistributionSerializer, GetOrCreateSeriali
     repository_version = RepositoryVersionRelatedField(
         required=False, help_text=_("RepositoryVersion to be served"), allow_null=True
     )
+    remote = DetailRelatedField(
```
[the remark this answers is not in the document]
lubosmj: I can make this field read-only.

This remote is used for downloading content from the remote registry, so we will save one more retrieval for distro->pull-through-distro->remote->URL (+ URL building).

pulp_container/app/tasks/download_image_data.py
```diff
@@ -0,0 +1,147 @@
...
+from pulp_container.constants import MEDIA_TYPE
+
+from .sync_stages import ContainerContentSaver, ContainerFirstStage
+
+log = logging.getLogger(__name__)
+
+
+def download_image_data(repository_pk, remote_pk, tag_name, response_data):
+    repository = ContainerRepository.objects.get(pk=repository_pk)
+    remote = ContainerRemote.objects.get(pk=remote_pk).cast()
```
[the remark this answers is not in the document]
lubosmj: It is not needed.

pulp_container/app/tasks/download_image_data.py
```diff
@@ -0,0 +1,147 @@
...
+                await listed_manifest["manifest_dc"].resolution()
+            await self.put(manifest_list_dc)
+        self.manifest_list_dcs.clear()
+
+        tagged_manifest_dc = tag_dc.extra_data["tagged_manifest_dc"]
+        tag_dc.content.tagged_manifest = await tagged_manifest_dc.resolution()
+        await self.put(tag_dc)
+
+
+class ContainerPullThroughCacheDeclarativeVersion(DeclarativeVersion):
```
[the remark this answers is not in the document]
lubosmj: I think you are right. I had issues with imports, but this looks to be unnecessary residues.

pulp_container/app/viewsets.py
```diff
@@ -1302,6 +1382,103 @@ def destroy(self, request, pk, **kwargs):
...
+            {
+                "action": ["list", "my_permissions"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": ["create"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": "has_namespace_model_perms",
```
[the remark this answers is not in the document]
lubosmj: Oh. I should add a similar validation to the `PullThroughDistribution` serializer: https://github.com/pulp/pulp_container/blob/552998fd29f89c4cf0c5a59240b22e3177ef3206/pulp_container/app/serializers.py#L336-L341. I would say that `namespace` will be always equal to base_path. Is it correct?

pulp_container/app/serializers.py
lubosmj: Same reasoning: https://github.com/pulp/pulp_container/pull/1299/files#r1399046429.

pulp_container/app/serializers.py
lubosmj: You cannot combine remotes because of https://github.com/pulp/pulp_container/blob/552998fd29f89c4cf0c5a59240b22e3177ef3206/pulp_container/app/models.py#L483:
```
[lmjachky@localhost ostree-perf]$ http :5001/pulp/api/v3/repositories/container/container/ remote=/pulp/api/v3/remotes/container/pull-through/018bec87-8fdb-7ca7-bde0-7d2fad4778d0/ name=cache
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 65
Correlation-ID: a4ec7f0ddd8c432ca9658c380d066030
Cross-Origin-Opener-Policy: same-origin
Date: Mon, 20 Nov 2023 11:40:03 GMT
Referrer-Policy: same-origin
Server: nginx/1.22.1
Vary: Accept
X-Frame-Options: DENY

{
    "remote": [
        "Type for Remote 'cache' does not match Repository."
    ]
}
```

pulp_container/app/registry.py
lubosmj: I tested it with raising 429 just now. It helped.

I was not raising anything (even 429) in the beginning, so I thought waiting might be a better approach because of:
```
DEBU[0003] Error pulling candidate localhost:5001/docker-cache/library/busybox:latest: copying system image from manifest list: determining manifest MIME type for docker://localhost:5001/docker-cache/library/busybox:latest: reading manifest sha256:023917ec6a886d0e8e15f28fb543515a5fcd8d938edb091e8147db4efed388ee in localhost:5001/docker-cache/library/busybox: StatusCode: 404, 404: Not Found 
```

Yes, another option is to tell the client to wait explicitly with 429. Podman handled this case flawlessly and waited 2 seconds:
```
DEBU[0000] GET http://localhost:5001/v2/docker-cache/library/busybox/manifests/latest 
DEBU[0001] Too many requests to http://localhost:5001/v2/docker-cache/library/busybox/manifests/latest: sleeping for 2.000000 seconds before next attempt 
DEBU[0003] GET http://localhost:5001/v2/docker-cache/library/busybox/manifests/latest 
```

But, should I start returning 429 in every case? What if it takes more than 2 seconds (podman default backoff) to finish the task? We may dispatch another task unnecessarily. This is hard to reproduce. I was not able to find any podman option to change this default.

pulp_container/app/downloaders.py
mdellweg: What is the lock even good for when you have it per instance instead of per class. I don't think this is how it was intended.

pulp_container/app/models.py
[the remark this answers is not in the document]
mdellweg: I think that the owner of the pull-through distribution should also own all the auto-created stuff related to it.

pulp_container/app/serializers.py
ipanova: yes, but try creating repo and remote separately and provide the remote during sync call

pulp_container/app/registry.py
lubosmj: Yes, we can background the task, by returning 429.

pulp_container/app/tasks/download_image_data.py
```diff
@@ -0,0 +1,147 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def download_image_data(repository_pk, remote_pk, tag_name, response_data):
+    repository = ContainerRepository.objects.get(pk=repository_pk)
+    remote = ContainerRemote.objects.get(pk=remote_pk).cast()
+    log.info("Pulling cache: repository={r} remote={p}".format(r=repository.name, p=remote.name))
+    first_stage = ContainerPullThroughFirstStage(remote, tag_name, response_data)
+    dv = ContainerPullThroughCacheDeclarativeVersion(first_stage, repository, mirror=False)
```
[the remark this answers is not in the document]
ipanova: I am looking at our old notes, I thought we discussed to actually use mirror=True so we do not store unnecessary data locally that is no longer available remotely

pulp_container/app/registry.py
ipanova: I see your point @mdellweg however I am a little bit uncomfortable returning 429 right away. It means on every first pull operation the client will be throttled.

ipanova: can this task be immediate?

ipanova: In other instances we raise Throttled, can you check what's the difference? I  know that with Throttled there is exponential back off

pulp_container/app/registry.py
```diff
@@ -117,7 +130,67 @@ async def get_tag(self, request):
...
+                try:
+                    manifest_data = json.loads(response.data)
+                except json.decoder.JSONDecodeError:
+                    raise PathNotResolved(tag_name)
+                else:
+                    encoded_data = response.data.encode("utf-8")
+                    digest = calculate_digest(encoded_data)
+                    media_type = determine_media_type(manifest_data, response)
+
+                response_headers = {
```
ipanova, line 182: i think we need this header too Docker-Distribution-API-Version

pulp_container/app/registry.py
mdellweg: Why do we need to wait for the result of the test is what I don't understand.
If we need to then immediate is probably the solution. An that would be immediate only and if it failed we return 429, but the task would have been canceled at that time.

lubosmj: Even with immediate tasks, this might take 2 seconds, which is blocking in the end.

lubosmj: We need to ensure that the next time podman requests blobs, they are already present in the repository, otherwise 404 is returned. That is why we need to wait a bit, but not too long. Raising 429 is a safe bet because we explicitly tell podman to wait.

ipanova: What we need is some sort of atomic operation that will ensure that manifest, blobs and repo version will exist by the time we respond to the client with a manifest. Otherwise the client risks to ask for the blob which we have not yet created/added to the repo and he will get a 404 in the middle of the pull operation.
we have another option. We do all the work here in the app, and the work would consist in creating blobs/manifest content, remote artifacts, content associations, assigned created content to pending content and then dispatch and forget(i think) add_and_remove (immediate) task that will actually create repo version.

ipanova: next during pull operation we'd need looking besides the available content in the repo version also in the pending_content for the case when task has not finished yet

ipanova: there is one caveat - we do not have pending_* on the regular repo, just on the push repo model

ipanova: yet another reason to merge those types...

lubosmj: I implemented pending_* on a regulare repo before, when I was handling the pull-through workflow from api-app. So, I do not think this is a dealbreaker. I am more worried about downloading listed manifests from content-app now.

pulp_container/app/registry.py
```diff
@@ -117,7 +130,67 @@ async def get_tag(self, request):
...
+                else:
+                    raise HTTPTooManyRequests()
+
+                try:
+                    manifest_data = json.loads(response.data)
+                except json.decoder.JSONDecodeError:
+                    raise PathNotResolved(tag_name)
+                else:
+                    encoded_data = response.data.encode("utf-8")
+                    digest = calculate_digest(encoded_data)
```
ipanova, line 179: can you use reponse.artifact_arrtibutes?

pulp_container/app/registry.py
```diff
@@ -169,13 +241,8 @@ async def dispatch_tag(self, request, tag, response_headers):
...
 
         """
-        try:
-            artifact = await tag.tagged_manifest._artifacts.aget()
-        except ObjectDoesNotExist:
-            ca = await sync_to_async(lambda x: x[0])(tag.tagged_manifest.contentartifact_set.all())
-            return await self._stream_content_artifact(request, web.StreamResponse(), ca)
-        else:
-            return await Registry._dispatch(artifact, response_headers)
+        artifact = await sync_to_async(tag.tagged_manifest._artifacts.get)()
```
ipanova, line 244: please revert back this change. You will get in trouble for the reclaimed repos.

pulp_container/app/registry_api.py
```diff
@@ -280,6 +284,44 @@ def get_drv_pull(self, path):
...
+            raise RepositoryNotFound(name=path)
+
+        try:
+            with transaction.atomic():
+                cache_repository, _ = models.ContainerRepository.objects.get_or_create(
+                    name=path, retain_repo_versions=1
+                )
+
+                upstream_name = path.split(root_cache_distribution.base_path, maxsplit=1)[1]
+                cache_remote, _ = models.ContainerRemote.objects.get_or_create(
```
ipanova, line 304: I need help here to understand. With this change you made a regular sync repo pull-through cache as well. I don't remember if this was intended or not.

I thought you will have pull-through-remote associate to special distribution only, and that remote will be used during pull-through requests.
Now i see the in-memory-download on the regular remote and not on special one?

pulp_container/app/registry_api.py
```diff
@@ -951,13 +993,30 @@ def handle_safe_method(self, request, path, pk):
             try:
                 tag = models.Tag.objects.get(name=pk, pk__in=repository_version.content)
             except models.Tag.DoesNotExist:
-                raise ManifestNotFound(reference=pk)
+                if distribution.remote:
+                    remote = distribution.remote.cast()
+                    repository = distribution.repository.cast()
+                    manifest = self.fetch_manifest(remote, repository_version, repository, pk)
```
ipanova, line 999: please add a comment why we fetch here with HEAD ---> to avoid 302---> possible 404

lubosmj: Also, I will return 429 from content-app in case we reach the pull limit from the dockerhub registry after issuing a GET request there.

pulp_container/app/registry_api.py
```diff
@@ -951,13 +993,30 @@ def handle_safe_method(self, request, path, pk):
             try:
                 tag = models.Tag.objects.get(name=pk, pk__in=repository_version.content)
```
ipanova, line 994: somewhere we need to add a check: if tag exists locally is it up-to-date with the remote source? ( issue a head request)

pulp_container/app/registry_api.py
```diff
@@ -951,13 +993,30 @@ def handle_safe_method(self, request, path, pk):
...
             except models.Tag.DoesNotExist:
-                raise ManifestNotFound(reference=pk)
+                if distribution.remote:
```
ipanova, line 996: here we need to add a check if this distro is part of pull-through-distro otherwise we enable pull-through caching for regular synced repos.

pulp_container/app/registry_api.py
lubosmj: Yes, a regular remote is used for the pull-through caching. The special remote comes from the idea that we want a dedicated remote managed inside pulp-container, but with no reference to `upstream-name`. This remote is meant to be attached to a special pull-through distribution that helps us build regular remotes on the fly when processing incoming requests for pull-throughs.

pulp_container/app/registry.py
lubosmj: Nice idea.

pulp_container/app/serializers.py
lubosmj: Correct. This case was not considered. Thanks for pointing out:
```
        "description": "ContainerRemote matching query does not exist.",
        "traceback": "  File \"/src/pulpcore/pulpcore/tasking/tasks.py\", line 60, in _execute_task\n    result = func(*args, **kwargs)\n  File \"/src/pulp_container/pulp_container/app/tasks/synchronize.py\", line 36, in synchronize\n    remote = ContainerRemote.objects.get(pk=remote_pk)\n  File \"/usr/local/lib/python3.8/site-packages/django/db/models/manager.py\", line 87, in manager_method\n    return getattr(self.get_queryset(), name)(*args, **kwargs)\n  File \"/usr/local/lib/python3.8/site-packages/django/db/models/query.py\", line 637, in get\n    raise self.model.DoesNotExist(\n"
    },

```

lubosmj:
As per our latest mini-team discussion, we all agreed upon improving the pull-through caching facility a bit more.

The final process involves the following steps:
1. Download a manifest list from a remote registry. (works now)
2. Dispatch a task to create manifests, blobs, and relations, adding these units to a new repository version. (works now)
3. Stream the downloaded manifest list back to the client. (works now)
4. Respond to client requests for specific manifests by downloading the required data, creating a new manifest unit, remote blobs, and streaming the new manifest back to the client. (needs to be implemented, one if-else branch)

Key points:
- All mentioned content units are saved to the "pending_*" relation stored under the repository. (needs to be implemented)
- The task is dispatched upon the first request, and its completion may take a while, depending on factors like the number of listed manifests and internet connectivity. (needs to be implemented, one if-else branch)
- Content resolution and querying are handled within pipeline stages in the dispatched task, eliminating the need for additional concerns for race conditions. (works now, we initialize units from content-app like CU.get_or_create)
- A new pipeline stage called "RemovePendingContent" to remove pending content from the locked repository will be added at the end of the pipeline. (needs to be implemented)

pulp_container/app/tasks/download_image_data.py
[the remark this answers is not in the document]
lubosmj: We decided that this pipeline will be a bit different than the original `ContainerDeclarativeVersion` by having a stage that deletes pending_* relations. So, I am leaving it here.

pulp_container/app/registry.py
lubosmj: I would say that only the V2 endpoint should serve this header.

https://github.com/openshift/docker-distribution/blob/master/docs/spec/api.md#api-version-check

lubosmj: To stay consistent with other endpoints, I am returning this header too.

ipanova: client might fall back to v1 api if it will not find v2 registry header

pulp_container/app/registry_api.py
lubosmj: I will move this check to content-app since that one would deal with content that should be downloaded. I find it redundant to fetch a manifest from api-app because we cannot share the information about updated content with content-app.

lubosmj:
@ipanova, @mdellweg, could you please take a look at this PR? I refactored the pipeline in a new commit so you could easily spot the changes after the last review.

Things to consider:
1. Immediate tasks cannot be executed from content-app (I tried adding `immediate=True`). I am getting "There is no current event loop in thread ThreadPoolExecuor_0_0" errors.
2. Setting `mirror=True` in the staging resulted in older tags being removed from the repository. I am using the default `mirror=False`.
3. To make things work, I had to remove the `asyncio.as_completed` call from `ContainerPullThroughFirstStage.run` because I was constantly seeing the following error after moving `asyncio.Lock()` back to the class definition. I use a standard for loop as a replacement: `for manifest_data in content_data.get("manifests")`.
```
pulp [9b53bb4a24074482ad396adc26513134]: pulp_container.app.tasks.download_image_data:INFO: Pulling cache: repository=e55a6861-15d4-4dbe-a573-37ffa93db19c/pulp/hello-world remote=e55a6861-15d4-4dbe-a573-37ffa93db19c/pulp/hello-world
[pulp]  | ('pulp [55f43f750e314533a18c52ce0ac50a24]: ::ffff:127.0.0.1 - - [13/Dec/2023:00:28:47 +0000] "GET /v2/e55a6861-15d4-4dbe-a573-37ffa93db19c/pulp/hello-world/manifests/sha256:239de6dd745ed1a211123322865b4c342c706e7c1e01644a1bbefe8f8846c5ff HTTP/1.0" 302 0 "-" "containers/5.26.2 (github.com/containers/image)"',)
pulp [9b53bb4a24074482ad396adc26513134]: pulp_container.app.downloaders:INFO: Updating bearer token
pulp [9b53bb4a24074482ad396adc26513134]: asyncio:ERROR: Task was destroyed but it is pending!
task: <Task pending name='Task-47' coro=<Event.wait() done, defined at /usr/lib64/python3.8/asyncio/locks.py:296> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f8b5ecabc40>()]>>
pulp [9b53bb4a24074482ad396adc26513134]: pulpcore.tasking.tasks:INFO: Task 018c6093-7ba9-79cb-b922-715950bb8d11 failed (Task <Task pending name='Task-14' coro=<ContainerFirstStage.create_listed_manifest() running at /src/pulp_container/pulp_container/app/tasks/sync_stages.py:481> cb=[as_completed.<locals>._on_completion() at /usr/lib64/python3.8/asyncio/tasks.py:606]> got Future <Future pending> attached to a different loop)
pulp [9b53bb4a24074482ad396adc26513134]: pulpcore.tasking.tasks:INFO:   File "/src/pulpcore/pulpcore/tasking/tasks.py", line 60, in _execute_task
    result = func(*args, **kwargs)

  File "/src/pulp_container/pulp_container/app/tasks/download_image_data.py", line 25, in download_image_data
    return dv.create()

  File "/src/pulpcore/pulpcore/plugin/stages/declarative_version.py", line 161, in create
    loop.run_until_complete(pipeline)

  File "/usr/lib64/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()

  File "/src/pulpcore/pulpcore/plugin/stages/api.py", line 220, in create_pipeline
    await asyncio.gather(*futures)

  File "/src/pulpcore/pulpcore/plugin/stages/api.py", line 41, in __call__
    await self.run()

  File "/src/pulp_container/pulp_container/app/tasks/download_image_data.py", line 61, in run
    listed_manifest = await listed_manifest_task

  File "/usr/lib64/python3.8/asyncio/tasks.py", line 619, in _wait_for_one
    return f.result()  # May raise f.exception().

  File "/src/pulp_container/pulp_container/app/tasks/sync_stages.py", line 481, in create_listed_manifest
    saved_artifact, content_data, _, response = await self._download_and_save_artifact_data(

  File "/src/pulp_container/pulp_container/app/tasks/sync_stages.py", line 71, in _download_and_save_artifact_data
    response = await downloader.run(extra_data={"headers": V2_ACCEPT_HEADERS})

  File "/src/pulpcore/pulpcore/download/http.py", line 269, in run
    return await download_wrapper()

  File "/usr/local/lib/python3.8/site-packages/backoff/_async.py", line 151, in retry
    ret = await target(*args, **kwargs)

  File "/src/pulpcore/pulpcore/download/http.py", line 254, in download_wrapper
    return await self._run(extra_data=extra_data)

  File "/src/pulp_container/pulp_container/app/downloaders.py", line 96, in _run
    await self.update_token(response_auth_header, this_token, repo_name)

  File "/src/pulp_container/pulp_container/app/downloaders.py", line 122, in update_token
    async with self.token_lock:

  File "/usr/lib64/python3.8/asyncio/locks.py", line 97, in __aenter__
    await self.acquire()

  File "/usr/lib64/python3.8/asyncio/locks.py", line 203, in acquire
    await fut

pulp [None]: pulp_container.app.downloaders:INFO: Updating bearer token
[pulp]  | ::ffff:127.0.0.1 [13/Dec/2023:00:28:47 +0000] "GET /pulp/container/e55a6861-15d4-4dbe-a573-37ffa93db19c/pulp/hello-world/manifests/sha256:239de6dd745ed1a211123322865b4c342c706e7c1e01644a1bbefe8f8846c5ff HTTP/1.0" 200 762 "http://localhost:5001/v2/e55a6861-15d4-4dbe-a573-37ffa93db19c/pulp/hello-world/manifests/sha256:239de6dd745ed1a211123322865b4c342c706e7c1e01644a1bbefe8f8846c5ff" "containers/5.26.2 (github.com/containers/image)"
```
I do not know what could be the root cause...

pulp_container/app/models.py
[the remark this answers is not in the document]
lubosmj: I would like to focus on delivering an MVP to see if we get any feedback for this feature.

mdellweg (commented):
I think immediate tasks that run async code called from other async code will attempt to create their own async event loop and that is the setup for the error you are seeing. But i thought we were dispatching the task to run independently in the background anyway never to be awaited.

pulp_container/app/registry.py
```diff
@@ -123,74 +122,144 @@ async def get_tag(self, request):
...
 
                 relative_url = "/v2/{name}/manifests/{tag}".format(
                     name=remote.namespaced_upstream_name, tag=tag_name
                 )
                 tag_url = urljoin(remote.url, relative_url)
-                downloader = remote.get_in_memory_downloader(url=tag_url)
-                response = await downloader.run(extra_data={"headers": V2_ACCEPT_HEADERS})
+                downloader = remote.get_downloader(url=tag_url)
+                try:
+                    response = await downloader.run(extra_data={"headers": V2_ACCEPT_HEADERS})
```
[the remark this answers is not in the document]
lubosmj: No, we want to ensure that we initialize the manifest and remote blobs first and add them to the pending_* content before dispatching the task and streaming data back to a client. The client can be faster than the task in this matter.

pulp_container/app/registry.py
```diff
@@ -123,74 +122,144 @@ async def get_tag(self, request):
...
+                            "manifest_artifact_pk": saved_artifact.pk,
+                            "tag_name": tag_name,
+                        },
+                    )
+
+                    manifest_data = json.loads(raw_data)
+                    media_type = determine_media_type(manifest_data, response)
+                    if media_type not in (MEDIA_TYPE.MANIFEST_LIST, MEDIA_TYPE.INDEX_OCI):
+                        await self.save_manifest_and_blobs(
+                            digest, manifest_data, media_type, remote, repository, saved_artifact
```
ipanova, line 251: the digest you got from docker-content-digest header is not reliable because that is not a required header. you should resort to calcualt_digest as in the except branch on line 176

pulp_container/app/registry.py
```diff
@@ -226,9 +295,103 @@ async def get_tag(self, request):
         # convert if necessary
         return await Registry.dispatch_converted_schema(tag, accepted_media_types, path)
 
+    async def save_manifest_and_blobs(
+        self, digest, manifest_data, media_type, remote, repository, artifact
+    ):
+        config_digest = manifest_data["config"]["digest"]
```
ipanova, line 301: how are you sure this is not a schema1 manifest?

pulp_container/app/registry.py
```diff
@@ -318,7 +486,54 @@ async def get_by_digest(self, request):
...
+                with open(response.path) as f:
+                    raw_data = f.read()
+
+                response.artifact_attributes["file"] = response.path
+                saved_artifact = await save_artifact(response.artifact_attributes)
+
+                try:
+                    manifest_data = json.loads(raw_data)
+                except json.decoder.JSONDecodeError:
+                    raise PathNotResolved(digest)
```
ipanova, line 516: should be path here

pulp_container/app/registry.py
```diff
@@ -318,7 +486,54 @@ async def get_by_digest(self, request):
                 "Docker-Content-Digest": ca_content.digest,
             }
         except ObjectDoesNotExist:
-            raise PathNotResolved(path)
+            distribution = await distribution.acast()
```
ipanova, line 489: you need to do something about this code being repeated 3 times in 3 places

pulp_container/app/registry_api.py
```diff
@@ -1017,23 +1039,21 @@ def handle_safe_method(self, request, path, pk):
...
-                        manifest = self.fetch_manifest(remote, repository_version, repository, pk)
-                        if manifest is None:
-                            return redirects.redirect_to_content_app("manifests", pk)
+                # the manifest might be a part of listed manifests currently being uploaded
+                # or saved during the pull-through caching
+                try:
+                    manifest = repository.pending_manifests.get(digest=pk)
+                    manifest.touch()
+                except models.Manifest.DoesNotExist:
+                    pass
```
ipanova, line 1048: why not raising the error?

ipanova:
@lubosmj 
>     2. Setting `mirror=True` in the staging resulted in older tags being removed from the repository. I am using the default `mirror=False`.

I thought this is the desired behavior. Since this is a pull-through-cache repository, it should match exactly the content remotely. So if a tag was removed remotely I would not know why we should keep it locally.

pulp_container/app/models.py
```diff
@@ -473,58 +416,11 @@ class Meta:
...
-            ValueError: If neither remote_artifact and url are passed, or if both are passed.
-
-        Returns:
-            subclass of :class:`~pulpcore.plugin.download.BaseDownloader`: A downloader that
-            is configured with the remote settings.
 
-        """
-        kwargs["remote"] = self
-        return super().get_downloader(remote_artifact=remote_artifact, url=url, **kwargs)
+    This remote is used for instantiating new regular container remotes with the upstream name.
```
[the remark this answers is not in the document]
lubosmj: Regular remotes come with the upstream name.

pulp_container/app/registry.py
```diff
@@ -123,74 +122,144 @@ async def get_tag(self, request):
...
+            if distribution.remote_id and distribution.pull_through_distribution_id:
                 remote = await distribution.remote.acast()
 
                 relative_url = "/v2/{name}/manifests/{tag}".format(
                     name=remote.namespaced_upstream_name, tag=tag_name
                 )
                 tag_url = urljoin(remote.url, relative_url)
-                downloader = remote.get_in_memory_downloader(url=tag_url)
-                response = await downloader.run(extra_data={"headers": V2_ACCEPT_HEADERS})
+                downloader = remote.get_downloader(url=tag_url)
```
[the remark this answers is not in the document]
lubosmj: A regular one, derived from a pull-through remote, initialized in registry_api.

pulp_container/app/registry.py
```diff
@@ -123,74 +122,144 @@ async def get_tag(self, request):
...
+                try:
+                    response = await downloader.run(
+                        extra_data={"headers": V2_ACCEPT_HEADERS, "http_method": "head"}
+                    )
+                except ClientResponseError:
+                    raise PathNotResolved(path)
+
+                digest = response.headers.get("docker-content-digest")
+                if tag.tagged_manifest.digest != digest:
+                    downloader = remote.get_downloader(url=tag_url)
```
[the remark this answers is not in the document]
lubosmj: In that case, we check for the digest with the HEAD request.

pulp_container/app/registry.py
lubosmj: I am not. 😭

lubosmj:
> I thought this is the desired behavior. Since this is a pull-through-cache repository, it should match exactly the content remotely. So if a tag was removed remotely I would not know why we should keep it locally.

But, using `mirror=True` causes the sync pipeline to think that the old tag, which was previously pulled by the user, had gone. It removes it even though it might be present on the remote. See the test case `test_pull_manifest_list` where I pull two different tags. With `mirror=True`, the test fails because of the missing `latest` tag.

How can I forcefully tell the sync pipeline to not remove the existing tag? Also, how do we know if the tag was removed from the remote registry if the user never asks for it and thus we never realize that?

pulp_container/app/registry_api.py
lubosmj: There is still a chance that the fired pull-through download task has not finished and the user is trying to get a new listed manifest. We do not pre-record all listed manifests and their blobs in content-app. So, the manifest does not exist in `pending_manifests` and is still not associated with any repository.

pulp_container/app/models.py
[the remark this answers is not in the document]
ipanova: ahh i see now

ipanova:
@lubosmj since we using `ContainerPullThroughCacheDeclarativeVersion` that never checks rest of content as it would usually check in the normal DeclarativeVersion, it will always keep only the latest pulled tag with mirror=True. So you're right we should use mirror=False because we have no choice

pulp_container/app/registry.py
lubosmj: I extracted a new class.

Issue #1296, opened 2023-06-01
Establish a new artifactless Blob Config model


lubosmj:
https://github.com/pulp/pulp_container/issues/1288#issuecomment-1560638504

Ultimately, we will have two models -- one for config blobs, and another for regular blobs. Only the former one will be artifactless since it stores less information.

git-hyagi:
After considering the benefits this change would bring and analyzing and identifying some complexities with it, the team decided to keep the current implementation as is (more details in https://github.com/pulp/pulp_container/pull/1603#issuecomment-2084939231 discussion).

Issue #1288, opened 2023-05-23
Make Manifest content type artifactless


ipanova:
Currently, `Manifests` have an artifact which is stored in the artifact storage - the `manifest.json` file.

The proposal is move out the `content_artifact.artifact` of a `Manifest` content into a field onto its DB model. `The manifest.json` is designed to be small so storing it in the DB, unlike the blobs ( which are binary data that can be large) should not be a concern.

Reasoning behind:
1. `manifest.json` file contains a lot of information and we  store just some of it on the model in the db. Everytime when we discover the need in the information that we do not have on the model we a) whether need to read this file from disk and parse it b) add the missing information onto the model that as a result brings a migration too.
2. `OCI manifest.json` specs change. At lease for the past year there were 2 spec change which have added new fields or renamed them.

By storing the json file in the DB it will be much cheaper the load it from the DB then from FS.

**Note** Out of scope but question is of the similar nature - `Config blobs` which are also json files could be stored in the DB as well, since this is where `Labels` of a container are defined, however, our model `Blob` covers also regular blobs ( large binary data) so we will whether need to split `ConfigBlob` into its separate model ( meh solution :/) or add a new field on the existing model which deepening on the `media_type` of the blob will whether have it populated or have the blob.content_artifact.artifact ( also meh solution :/)

mdellweg:
Re config blobs: Having the Config blob as a separate content type would keep the number of artifacts per Blob constant (at 1). This is probably the solution with less surprises. At that point I would, for flexibility, have an ArtifactBlob and a DBBlob (both inheriting the generic Blob) so we can onboard future blob types easily.
However _just_ making the (single) Blob type capable of inlining next to artifact storage also has some nice aspects to it. I think we could make that even in a zero downtime manner.

lubosmj:
Two things to consider. We will need to change the sync pipeline because the artifact will no longer be present in the filesystem. We should also handle the reclamation task: Remove: PROTECTED_FROM_RECLAIM = False.

For the blobs thingy, we could open a separate issue to track the progress.

lubosmj:
I opened a new issue to track the blob model's refactorization: https://github.com/pulp/pulp_container/issues/1296

lubosmj:
Do we want to have this data saved into `JSONField` or saved as a string? https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.JSONField

ipanova:
+1 to jsonfield

PR #1301, opened 2023-06-06
Implement flatpak index view

### LINKED ISSUE #1315: As a user I host and expose flatpack content via flatpack index
Flatpack content can be packaged as a regular container image.
Flatpack clients usually send request with different query params to the /index/static or /index/dynamic endpoint.

This index contains  flatpack container images.

See specs for more https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md
These specs are not part of official docker registry API, hence we are going to introduce a feature setting that will enable Flatpack index. It will be off by default.

stbergmann:
This implements the protocol documented at <https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md>, with the following notes:
* It does not announce any Annotations, only Labels.
* It does not announce any Lists of images, only individual Images.

Both of those simplifications are in line with how upstream flatpak indexers (e.g., <https://github.com/owtaylor/flatpak-indexer> serving <https://registry.fedoraproject.org/>) behave and with what clients actually expect and make use of.

(This fixes <https://issues.redhat.com/browse/SAT-4416> "Create a JSON index to allow installing via Flatpak client".)

closes #1315

### DIFF
--- a/pulp_container/app/cache.py
+++ b/pulp_container/app/cache.py
@@ -6,6 +6,7 @@ from pulp_container.app.models import ContainerDistribution
 from pulp_container.app.exceptions import RepositoryNotFound
 
 ACCEPT_HEADER_KEY = "accept_header"
+QUERY_KEY = "query"
 
 
 class RegistryCache:
@@ -72,3 +73,19 @@ def find_base_path_cached(request, cached):
         except ObjectDoesNotExist:
             raise RepositoryNotFound(name=path)
         return distro.base_path
+
+
+class FlatpakIndexStaticCache(SyncContentCache):
+    def __init__(self, expires_ttl=None, auth=None):
+        updated_keys = (QUERY_KEY,)
+        super().__init__(
+            base_key="/index/static", expires_ttl=expires_ttl, keys=updated_keys, auth=auth
+        )
+
+    def make_key(self, request):
+        """Make a key composed of the request's query."""
+        all_keys = {
+            QUERY_KEY: request.query_params.urlencode(),
+        }
+        key = ":".join(all_keys[k] for k in self.keys)
+        return key
--- a/pulp_container/app/models.py
+++ b/pulp_container/app/models.py
@@ -11,7 +11,9 @@ from django.db import models
 from django.conf import settings
 from django.contrib.postgres import fields
 from django.shortcuts import redirect
+from django_lifecycle import hook, AFTER_CREATE, AFTER_DELETE, AFTER_UPDATE
 
+from pulpcore.plugin.cache import SyncContentCache
 from pulpcore.plugin.download import DownloaderFactory
 from pulpcore.plugin.models import (
     Artifact,
@@ -611,6 +613,23 @@ class ContainerDistribution(Distribution, AutoAddObjPermsMixin):
         if self.content_guard:
             url = self.content_guard.cast().preauthenticate_url(url)
         return redirect(url)
+
+    @hook(AFTER_CREATE)
+    @hook(AFTER_DELETE)
+    @hook(
+        AFTER_UPDATE,
+        when_any=[
+            "base_path",
+            "private",
+            "repository",
+            "repository_version",
+        ],
+        has_changed=True,
+    )
+    def invalidate_flatpak_index_cache(self):
+        """Invalidates the cache for /index/static."""
+        if settings.CACHE_ENABLED:
+            SyncContentCache().delete(base_key="/index/static")
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -46,7 +46,11 @@ from rest_framework.views import APIView
 
 from pulp_container.app import models, serializers
 from pulp_container.app.authorization import AuthorizationService
-from pulp_container.app.cache import find_base_path_cached, RegistryApiCache
+from pulp_container.app.cache import (
+    find_base_path_cached,
+    FlatpakIndexStaticCache,
+    RegistryApiCache,
+)
 from pulp_container.app.exceptions import (
     unauthorized_exception_handler,
     InvalidRequest,
@@ -471,6 +475,134 @@ class CatalogView(ContainerRegistryApiMixin, ListAPIView):
 
         accessible_repositories = repositories_by_distribution & repositories_by_namespace
         return (public_repositories | accessible_repositories).distinct()
+
+
+class FlatpakIndexDynamicView(APIView):
+    """
+    Handles requests to the /index/dynamic endpoint
+    """
+
+    authentication_classes = []
+    permission_classes = []
+
+    def recurse_through_manifest_lists(self, tag, manifest, oss, architectures, manifests):
+        if manifest.media_type in (models.MEDIA_TYPE.MANIFEST_V2, models.MEDIA_TYPE.MANIFEST_OCI):
+            manifests.setdefault(manifest, set()).add(tag)
+        elif manifest.media_type in (models.MEDIA_TYPE.MANIFEST_LIST, models.MEDIA_TYPE.INDEX_OCI):
+            mlms = manifest.listed_manifests.through.objects.filter(image_manifest__pk=manifest.pk)
+            if oss:
+                mlms.filter(os__in=oss)
+            if architectures:
+                mlms.filter(architecture__in=architectures)
+            for mlm in mlms:
+                self.recurse_through_manifest_lists(
+                    tag, mlm.manifest_list, oss, architectures, manifests
+                )
+
+    def get(self, request):
+        req_repositories = None
+        req_tags = None
+        req_oss = None
+        req_architectures = None
+        req_label_exists = set()
+        req_label_values = {}
+        for key, values in request.query_params.lists():
+            if key == "repository":
+                req_repositories = values
+            elif key == "tag":
+                req_tags = values
+            elif key == "os":
+                req_oss = values
+            elif key == "architecture":
+                req_architectures = values
+            elif key.startswith("label:"):
+                if key.endswith(":exists"):
+                    if any(v != "1" for v in values):
+                        raise ParseError(detail=f"{key} must have value 1.")
+                    label = key[len("label:") : len(key) - len(":exists")]
+                    req_label_exists.add(label)
+                else:
+                    label = key[len("label:") :]
+                    req_label_values[label] = values
+            else:
+                # In particularly, this covers any annotation:... parameters, which this
+                # implementation does not support:
+                raise ParseError(detail=f"Unsupported {key}.")
+
+        if "org.flatpak.ref" not in req_label_exists:
+            raise ParseError(detail="Missing label:org.flatpak.ref:exists=1.")
+
+        distributions = models.ContainerDistribution.objects.filter(private=False).only("base_path")
+
+        if req_repositories:
+            distributions = distributions.filter(base_path__in=req_repositories)
+
+        results = []
+        for distribution in distributions:
+            images = []
+            if distribution.repository:
+                repository_version = distribution.repository.latest_version()
+            else:
+                repository_version = distribution.repository_version
+            tags = models.Tag.objects.select_related("tagged_manifest").filter(
+                pk__in=repository_version.content
+            )
+            if req_tags:
+                tags = tags.filter(name__in=req_tags)
+            manifests = {}  # mapping manifests to sets of encountered tag names
+            for tag in tags:
+                self.recurse_through_manifest_lists(
+                    tag.name, tag.tagged_manifest, req_oss, req_architectures, manifests
+                )
+            for manifest, tagged in manifests.items():
+                with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
+                    raw_data = file.read()
+                config_data = json.loads(raw_data)
+                labels = config_data.get("config", {}).get("Labels")
+                if not labels:
+                    continue
+                if any(label not in labels.keys() for label in req_label_exists):
+                    continue
+                os = config_data["os"]
+                if req_oss and os not in req_oss:
+                    continue
+                architecture = config_data["architecture"]
+                if req_architectures and architecture not in req_architectures:
+                    continue
+                if any(
+                    labels.get(label) not in values for label, values in req_label_values.items()
+                ):
+                    continue
+                images.append(
+                    {
+                        "Tags": tagged,
+                        "Digest": manifest.digest,
+                        "MediaType": manifest.media_type,
+                        "OS": os,
+                        "Architecture": architecture,
+                        "Labels": labels,
+                    }
+                )
+            if images:
+                results.append({"Name": distribution.base_path, "Images": images})
+
+        return Response(data={"Registry": settings.CONTENT_ORIGIN, "Results": results})
+
+
+class FlatpakIndexStaticView(FlatpakIndexDynamicView):
+    """
+    Handles requests to the /index/static endpoint
+    """
+
+    @FlatpakIndexStaticCache()
+    def get(self, request):
+        response = super().get(request)
+        # Avoid django.template.response.ContentNotRenderedError:
+        response.accepted_renderer = JSONRenderer()
+        response.accepted_media_type = JSONRenderer.media_type
+        response.renderer_context = {}
+        response.render()
+        return response
 
 
 class ContainerTagListSerializer(ModelSerializer):
--- a/pulp_container/tests/functional/api/test_flatpak.py
+++ b/pulp_container/tests/functional/api/test_flatpak.py
@@ -1,0 +1,72 @@
+"""Tests that verify Flatpak support"""
+import pytest
+import subprocess
+
+from django.conf import settings
+
+from pulp_container.tests.functional.constants import REGISTRY_V2
+
+
+def test_flatpak_install(
+    add_to_cleanup,
+    registry_client,
+    local_registry,
+    container_namespace_api,
+):
+    if not settings.FLATPAK_INDEX:
+        pytest.skip("This test requires FLATPAK_INDEX to be enabled")
+
+    image_path1 = f"{REGISTRY_V2}/pulp/oci-net.fishsoup.busyboxplatform:latest"
+    registry_client.pull(image_path1)
+    local_registry.tag_and_push(image_path1, "pulptest/oci-net.fishsoup.busyboxplatform:latest")
+    image_path2 = f"{REGISTRY_V2}/pulp/oci-net.fishsoup.hello:latest"
+    registry_client.pull(image_path2)
+    local_registry.tag_and_push(image_path2, "pulptest/oci-net.fishsoup.hello:latest")
+    namespace = container_namespace_api.list(name="pulptest").results[0]
+    add_to_cleanup(container_namespace_api, namespace.pulp_href)
+
+    # Install flatpak:
+    subprocess.check_call(
+        [
+            "flatpak",
+            "--user",
+            "remote-add",
+            "pulptest",
+            "oci+" + settings.CONTENT_ORIGIN,
+        ]
+    )
+    # See <https://pagure.io/fedora-lorax-templates/c/cc1155372046baa58f9d2cc27a9e5473bf05a3fb>
+    # "lorax-embed-flatpaks.tmpl: Run the flatpak-install under dbus-run-session" for the need for
+    # dbus-run-session to avoid "error: Cannot autolaunch D-Bus without X11 $DISPLAY":
+    subprocess.check_call(
+        [
+            "dbus-run-session",
+            "flatpak",
+            "--user",
+            "install",
+            "--noninteractive",
+            "pulptest",
+            "net.fishsoup.Hello",
+        ]
+    )
+
+    # Clean up flatpak:
+    subprocess.run(
+        [
+            "flatpak",
+            "--user",
+            "uninstall",
+            "--noninteractive",
+            "net.fishsoup.Hello",
+        ]
+    )
+    subprocess.run(
+        [
+            "flatpak",
+            "--user",
+            "uninstall",
+            "--noninteractive",
+            "net.fishsoup.BusyBoxPlatform",
+        ]
+    )
+    subprocess.run(["flatpak", "--user", "remote-delete", "pulptest"])
--- a/template_config.yml
+++ b/template_config.yml
@@ -62,8 +62,14 @@ pulp_settings:
   - /tmp
   allowed_import_paths:
   - /tmp
-pulp_settings_azure: null
+  flatpak_index: true
+pulp_settings_azure:
+  flatpak_index: true
 pulp_settings_gcp: null
+# Do not enable flatpak_index for s3 for now, as it would cause
+# pulp_container/tests/functional/api/test_flatpak.py to fail due to
+# <https://gitlab.gnome.org/GNOME/libsoup/-/issues/357> "Odd 400 response with https GET 302
+# redirecting to http":
 pulp_settings_s3: null
 pulp_settings_stream: null
 pulpprojectdotorg_key_id: aa499d7938ed
--- a/.github/workflows/scripts/install.sh
+++ b/.github/workflows/scripts/install.sh
@@ -68,7 +68,7 @@ services:
 VARSYAML
 
 cat >> vars/main.yaml << VARSYAML
-pulp_settings: {"allowed_content_checksums": ["sha1", "sha224", "sha256", "sha384", "sha512"], "allowed_export_paths": ["/tmp"], "allowed_import_paths": ["/tmp"]}
+pulp_settings: {"allowed_content_checksums": ["sha1", "sha224", "sha256", "sha384", "sha512"], "allowed_export_paths": ["/tmp"], "allowed_import_paths": ["/tmp"], "flatpak_index": true}
 pulp_scheme: https
 
 pulp_container_tag: https
@@ -107,7 +107,7 @@ if [ "$TEST" = "azure" ]; then
       - ./azurite:/etc/pulp\
     command: "azurite-blob --blobHost 0.0.0.0 --cert /etc/pulp/azcert.pem --key /etc/pulp/azkey.pem"' vars/main.yaml
   sed -i -e '$a azure_test: true\
-pulp_scenario_settings: null\
+pulp_scenario_settings: {"flatpak_index": true}\
 ' vars/main.yaml
 fi
 
--- a/.github/workflows/scripts/post_before_script.sh
+++ b/.github/workflows/scripts/post_before_script.sh
@@ -1,0 +1,4 @@
+if [[ " ${SCENARIOS[*]} " =~ " ${TEST} " ]]; then
+  # Needed by pulp_container/tests/functional/api/test_flatpak.py:
+  cmd_prefix dnf install -yq dbus-daemon flatpak
+fi
--- a/CHANGES/1315.feature
+++ b/CHANGES/1315.feature
@@ -1,0 +1,1 @@
+Added support for Flatpak index endpoints.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -28,6 +28,7 @@ Features
 * Host content either `locally or on S3 <https://docs.pulpproject.org/installation/storage.html>`_
 * De-duplication of all saved content
 * Support disconnected and air-gapped environments with the Pulp Import/Export facility for container repositories
+* Support for :ref:`hosting Flatpak content in OCI format <flatpak-workflow>`
 
 Tech Preview
 ------------
--- a/docs/tech-preview.rst
+++ b/docs/tech-preview.rst
@@ -4,3 +4,4 @@ Tech previews
 The following features are currently being released as part of a tech preview:
 
 * Build an OCI image from a Containerfile
+* Support for hosting Flatpak content in OCI format.
--- a/docs/workflows/flatpak-support.rst
+++ b/docs/workflows/flatpak-support.rst
@@ -1,0 +1,64 @@
+.. _flatpak-workflow:
+
+Hosting Flatpak Content in OCI Format
+=====================================
+
+Pulp can host Flatpak application and runtime images that are distributed in OCI format.  To make
+such content discoverable, it can provide ``/index/dynamic`` and ``/index/static`` endpoints as
+specified by `the Flatpak registry index protocol
+<https://github.com/flatpak/flatpak-oci-specs/blob/main/registry-index.md>`_.  This is not enabled
+by default.  To enable it, define ``FLATPAK_INDEX = True`` in the settings file.
+
+Clients like the ``flatpak`` command-line tool or the GNOME Software application will typically
+query the ``/index/static`` endpoint, which is intended to be called repeatedly with identical query
+parameters, and whose responses are meant to be cached.  The ``/index/dynamic`` endpoint serves
+exactly the same content, but is intended for one-off requests that should not be cached.  These
+endpoints can be accessed without authentication.  They only provide information about public
+repositories.
+
+The two endpoints support a number of query parameters (``architecture``, ``tag``, ``label``, etc.),
+see the protocol specification for details.  Two notes:
+
+* Every request must include a ``label:org.flatpak.ref:exists=1`` query parameter.  This acts as a
+  marker to only report Flatpak content, and to exclude other container content that may also be
+  provided by the Pulp instance.
+
+* This implementation does not support annotations.  Including any ``annotation`` query parameters
+  will result in a 400 failure response.  Use ``label`` query parameters instead.  (Existing clients
+  like the ``flatpak`` command-line tool never issue requests including any ``annotation`` query
+  parameters.)
+
+Install a Flatpak image from Pulp
+---------------------------------
+
+This section assumes that you have created at least one public distribution in your Pulp instance
+that serves a repository containing Flatpak content.  To do this, see the general :doc:`host`
+documentation.
+
+You can for example use the ``flatpak`` `command-line tool
+<https://docs.flatpak.org/en/latest/using-flatpak.html#the-flatpak-command>`_ to set up a Flatpak
+remote (named ``pulp`` here) that references your Pulp instance:
+
+.. code:: shell
+
+  flatpak remote-add pulp oci+"$BASE_ADDR"
+
+Then, use
+
+.. code:: shell
+
+  flatpak remote-ls pulp
+
+to retrieve a list of all Flatpak applications and runtimes that your Pulp instance serves.  (This
+queries the ``/index/static`` endpoint, as explained above.)  Finally, if your Pulp instance serves
+e.g. the ``org.gnome.gedit`` application, use
+
+.. code:: shell
+
+  flatpak install pulp org.gnome.gedit
+
+to install it and run it with
+
+.. code:: shell
+
+  flatpak run org.gnome.gedit
--- a/docs/workflows/index.rst
+++ b/docs/workflows/index.rst
@@ -78,3 +78,4 @@ OCI artifact support
    cosign-support
    helm-support
    oci-artifacts
+   flatpak-support
--- a/pulp_container/app/settings.py
+++ b/pulp_container/app/settings.py
@@ -46,3 +46,5 @@ ADDITIONAL_OCI_ARTIFACT_TYPES = {
         "application/vnd.wasm.content.layer.v1+wasm",
     ],
 }
+
+FLATPAK_INDEX = False
--- a/pulp_container/app/urls.py
+++ b/pulp_container/app/urls.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.urls import include, path
 from rest_framework.routers import Route, SimpleRouter
 from pulp_container.app.registry_api import (
@@ -5,6 +6,8 @@ from pulp_container.app.registry_api import (
     Blobs,
     BlobUploads,
     CatalogView,
+    FlatpakIndexDynamicView,
+    FlatpakIndexStaticView,
     Manifests,
     Signatures,
     TagsListView,
@@ -35,3 +38,10 @@ urlpatterns = [
     path("v2/<path:path>/tags/list", TagsListView.as_view()),
     path("", include(router.urls)),
 ]
+if settings.FLATPAK_INDEX:
+    urlpatterns.extend(
+        [
+            path("index/dynamic", FlatpakIndexDynamicView.as_view()),
+            path("index/static", FlatpakIndexStaticView.as_view()),
+        ]
+    )


pulp_container/app/registry_api.py
```diff
@@ -475,6 +475,144 @@ def get_queryset(self, *args, **kwargs):
...
+                    break
+            elif key == "architecture":
+                req_architecture = values[0]
+                if any(v != req_architecture for v in values[1:]):
+                    bad = True
+                    break
+            elif key.startswith("annotation:"):
+                # This implementation never announces any Annotations:
+                bad = True
+                break
```
lubosmj, line 512: Perhaps, we can remove this?

pulp_container/app/registry_api.py
```diff
@@ -475,6 +475,144 @@ def get_queryset(self, *args, **kwargs):
...
+                            "MediaType": manifest.media_type,
+                            "OS": os,
+                            "Architecture": architecture,
+                            "Labels": labels,
+                        }
+                    )
+                if images:
+                    results.append({"Name": distribution.base_path, "Images": images})
+
+        return Response(data={"Registry": "..", "Results": results})
```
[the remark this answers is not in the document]
stbergmann: What `settings.CONTENT_REGISTRY` would that be?  Doing that change as-is causes "AttributeError: 'Settings' object has no attribute 'CONTENT_REGISTRY'", and I don't find no `CONTENT_REGISTRY` at all when I grep pulpcore and pulp_container.

pulp_container/app/registry_api.py
```diff
@@ -475,6 +475,144 @@ def get_queryset(self, *args, **kwargs):
...
+            namespaces = models.ContainerNamespace.objects.all()
+            repositories_by_namespace = get_objects_for_user(
+                self.request.user, namespace_permission, namespaces
+            )
+            repositories_by_namespace = queryset.filter(namespace__in=repositories_by_namespace)
+            accessible_repositories = repositories_by_distribution & repositories_by_namespace
+            repositories = (public_repositories | accessible_repositories).distinct()
+
+            if req_repository is not None:
+                repositories = repositories.filter(base_path=req_repository)
```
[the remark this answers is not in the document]
stbergmann: No, that `req_repository` models what other languages would model with something like Haskell `Maybe String`, either `Nothing` (i.e., `None` here) or `Just s` (i.e., a string object here).

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
lubosmj: Oh, use CONTENT_ORIGIN. I am sorry.

pulp_container/app/registry_api.py
```diff
@@ -475,6 +475,144 @@ def get_queryset(self, *args, **kwargs):
...
+
+    def get(self, request):
+        bad = False
+        req_repository = None
+        req_tags = set()
+        req_os = None
+        req_architecture = None
+        req_label_exists = {"org.flatpak.ref"}
+        req_label_values = {}
+        for (key, values) in request.query_params.lists():
```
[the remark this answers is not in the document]
stbergmann: I see multiple issues:
(1)  <https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.QueryDict.__getitem__> states "If the key has more than one value, it returns the last value".  This is a no-go for `tag`.  And for other keys, we would silently ignore conflicting request parameters, and I think that's not a good idea (i.e., `?os=foo&os=bar` should arguably return an empty result set, not the result set for `?os=bar`).  And for `label:...[:exists]`, there is no way to guess all the different keys to `get()`, we actually need to somehow retrieve all of them.
(2) We would no longer check for unknown request parameters, would instead silently ignore them, and I don't know if that's a good idea.
(3) `bad` is not only about unknown request parameters, but about unsatisfiable requests in general (e.g., for an unsatisfiable `?os=foo&os=bar`).  I tried to clarify that now with <https://github.com/stbergmann/pulp_container/commit/726ca47fbb9e5b1c39a03f4afd99775bae5afaf1> "Clarify 'bad' -> 'unsatisfiable'".
(4) About extracting the validation into a separate method:  That would need to return in some way a 7-tuple `(bad, req_repository, req_tags, reg_os, reg_architecture, req_label_exists, req_label_values)`, and I'm not sure that wouldn't make the code unnecessarily verbose and less clear, overall?

stbergmann:
> Would you mind further squashing all the commits into a single one

@ipanova requested that "you can keep the commits with all the extra info you've put there since that will help".  Please let me know what's preferred.

> and fixing the lint errors: https://github.com/pulp/pulp_container/actions/runs/5201618714/jobs/9384837157?pr=1301?

done, <https://github.com/stbergmann/pulp_container/commit/2ce9aeb0c246f0e51e26cfab2f6cb2995a849063> "Manually reformat according to black 23.3.0"

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
lubosmj: Then, `getlist` might help.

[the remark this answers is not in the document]
stbergmann: Ah, hadn't spotted `getlist`.  But that would still not help with the `label:...[:exists]` keys.

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
gerrod3: Shouldn't `repositories` always return a query of 1 (or 0). `base_path` is a unique field for distributions so filtering by it will return a unique distribution.

pulp_container/app/cache.py
```diff
@@ -72,3 +73,19 @@ def find_base_path_cached(request, cached):
         except ObjectDoesNotExist:
             raise RepositoryNotFound(name=path)
         return distro.base_path
+
+
+class FlatpakCache(SyncContentCache):
```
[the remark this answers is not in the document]
gerrod3: The `base-key` for the cache should (normally) be based on the base-path of the distribution for the request being made. See https://github.com/pulp/pulpcore/blob/main/pulpcore/app/util.py#L424 for how we normally determine the `base-key`. And then the `key` should typically be based on the incoming request (path/headers/method/etc...). 

This case seems to be special since this new endpoint is not based on the `base_path` of a distribution and seems to be serving content from multiple distributions. I think having it set to '/static/index' is fine. You'll need to update the hook on the container distribution and repository models to delete this cache entry upon update of those models. [0] [1]

Also, since the response of this endpoint changes based on the permissions of the user I think you need to add `request.user` to the `key` in `make_key`.

[0] https://github.com/pulp/pulpcore/blob/3851472b3c9bb27578729943040fca57fc52182b/pulpcore/app/models/publication.py#L601
[1] https://github.com/pulp/pulpcore/blob/3851472b3c9bb27578729943040fca57fc52182b/pulpcore/app/models/repository.py#L278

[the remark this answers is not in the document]
stbergmann: @gerrod3, re "You'll need to update the hook on the container distribution and repository models":  Just to make sure:  The hook conditions on the added `ContainerDistribution` and `ContainerRepository`  `invalidate_cache` methods would need to cover at least all of the existing hook conditions on the base `Distribution` and `Repository` `invalidate_cache` methods, right?  (Or else `invalidate_cache` would stop being called for derived instances on the missing conditions, I assume.  Which looks a bit brittle, that the hooks in `pulp_container` would need to be adapted whenever the underlying hooks in `pulpcore` change.)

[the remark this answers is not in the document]
gerrod3: You should just be able to override the `invalidate_cache` method on those two models, call `super` first and then add your own cache invalidate call. The hooks should stay inplace (atleast that's how I think Python should work).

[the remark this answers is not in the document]
stbergmann: Ach, ignore me.  I can just add some completely unrelated `invalidate_flatpak_cache` methods to the derived classes, with their own sets of hook conditions...

pulp_container/app/models.py
```diff
@@ -11,7 +11,9 @@
 from django.conf import settings
 from django.contrib.postgres import fields
 from django.shortcuts import redirect
+from django_lifecycle import hook, AFTER_CREATE, AFTER_DELETE, AFTER_UPDATE
 
+from pulpcore.cache import Cache
```
stbergmann, line 16: This caused the <https://github.com/pulp/pulp_container/actions/runs/5291865234> lint job to fail with
```
ERROR: Detected bad imports from pulpcore:
./pulp_container/app/models.py:16:from pulpcore.cache import Cache
```
So we'd need to first export `Cache` from `pulpcore.plugin`?  Is there a protocol how to coordinate interdependent PRs to `pulpcore` and `pulp_container`?

lubosmj: Yes, you should be importing pulpcore facilities from the plugin namespace.

In case you want to make changes in pulpcore and then benefit from it in a plugin, follow this section: https://docs.pulpproject.org/pulpcore/contributing/git.html#requiring-other-pull-requests

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+            repositories_by_distribution = get_objects_for_user(
+                self.request.user, distribution_permission, queryset
+            )
+            namespaces = models.ContainerNamespace.objects.all()
+            repositories_by_namespace = get_objects_for_user(
+                self.request.user, namespace_permission, namespaces
+            )
+            repositories_by_namespace = queryset.filter(namespace__in=repositories_by_namespace)
+            accessible_repositories = repositories_by_distribution & repositories_by_namespace
+            repositories = (public_repositories | accessible_repositories).distinct()
```
ipanova, line 553: s/repositories/registry_repositories

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+
+            if req_repository is not None:
+                repositories = repositories.filter(base_path=req_repository)
+
+            for distribution in repositories:
```
ipanova, line 558: for example s/repositories/registry_repositories

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+                tags = models.Tag.objects.filter(pk__in=repository_version.content)
+                manifests = models.Manifest.objects.filter(
+                    pk__in=repository_version.content,
+                    media_type__in=(models.MEDIA_TYPE.MANIFEST_V2, models.MEDIA_TYPE.MANIFEST_OCI),
+                )
+                for manifest in manifests:
+                    with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
+                        raw_data = file.read()
+                    config_data = json.loads(raw_data)
+                    labels = config_data.get("config", {}).get("Labels")
```
ipanova, line 573: config is required for these media_types

ipanova (commented):
As discussed on slack with @stbergmann  there is some undocumented/outdated information on the specs that will need to be fixed.
Another question i have, is about specs themselves, they do not really seem to be official or finalized. They are not accepted by registry specs either. Last mention I found about them is that they are in DRAFT https://opencontainers.org/posts/blog/2018-11-07-bringing-oci-images-to-the-desktop-with-flatpak/

With this reasoning we need to introduce a so called feature flag, probably a pulp setting that will enabling or disabling the flatpack index support in the pulp-container registry. By default it should be disabled.

@stbergmann  can you provide multiple various examples of how client request would look like?
We also will need to add docs as part of this PR.

stbergmann:
> As discussed on slack with @stbergmann there is some undocumented/outdated information on the specs that will need to be fixed.

I had created <https://github.com/owtaylor/flagstate/pull/6> "protocol.md: Fix note formatting and add a note about annotations and image lists".  I'm not aware there's further "undocumented/outdated information" apart from that "note that annotations and image lists are currently unused" (and thus not implemented by this pulp_container PR).  Let me know if you think more is missing.

> Another question i have, is about specs themselves, they do not really seem to be official or finalized. They are not accepted by registry specs either. Last mention I found about them is that they are in DRAFT https://opencontainers.org/posts/blog/2018-11-07-bringing-oci-images-to-the-desktop-with-flatpak/

Can you be more specific what exactly you mean with "registry specs"?

On a RH-internal slack channel, @owtaylor mentioned that the "docs [...] at https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md are as official as it gets.  I perhaps should move that protocol doc to a separate repository under github.com/flatpak, or into the flatpak project, since the "flagstate" project is dead, but you can certainly count it as official"

> With this reasoning we need to introduce a so called feature flag, probably a pulp setting that will enabling or disabling the flatpack index support in the pulp-container registry. By default it should be disabled.

OK, I will look into implementing that.

> @stbergmann can you provide multiple various examples of how client request would look like? We also will need to add docs as part of this PR.

Clients here are the `flatpak` command line interface and (IIUC) integration into applications like GNOME Software through libflatpak (both the `flatpak` command line tool and libflatpak are provided by <https://github.com/flatpak/flatpak>).  A typical usage scenario is to set up an OCI-based remote,
```
$ flatpak remote-add some-name oci+https://...
```
pointing at the <https://...> base URL relative to which the `./index/static` endpoint is served.  Then,
```
$ flatpak remote-ls some-name
```
retrieves information from that endpoint via some <https://.../index/static?label:org.flatpak.ref:exists=1&architecture=...&os=linux&tag=latest>.  Then, given the identifier of some flatpak app advertised through the above,
```
$ flatpak install some-name identifier
```
pulls the relevant <https://.../base-path/blobs/...>.

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+    Handles requests to the /index/dynamic endpoint
+    """
+
+    def get(self, request):
+        unsatisfiable = False
+        req_repository = None
+        req_tags = set()
+        req_os = None
+        req_architecture = None
+        req_label_exists = {"org.flatpak.ref"}
```
ipanova, line 493: should the second label be added here too? https://github.com/owtaylor/flatpak-indexer/blob/main/flatpak_indexer/indexer.py#L85

ipanova:
I filed story to track this https://github.com/pulp/pulp_container/issues/1315

pulp_container/app/registry_api.py
stbergmann: With "second label" you refer to `k.startswith('org.freedesktop.')`, i.e., initialize `req_label_exists` to `{"org.flatpak.ref", "org.freedesktop.ref"}`?  That wouldn't work, as it would require that any images to be returned would need to have both some `org.flatpak.ref` and some `org.freedesktop.ref` label.  But e.g. <https://registry.fedoraroject.org/index/static> only reports `org.flatpak.ref` labels and no `org.freedesktop.ref` labels at all.  So if we import images from that flatpak registry into pulp, none of the imported images would have an `org.freedesktop.ref` label, and this endpoint would in turn not report any images to clients.
And I think the flatpak-indexer code you reference is about including both `org.flatpak...` and `org.freedesktop...` (like `org.freedesktop.appstream.appdata`) Koji labels, not about support for some hypothetical `org.freedesktop.ref` label.
(The code above, `req_label_exists = {"org.flatpak.ref"}`, is about presetting the set of requested `label:...:exists=1` query parameters with `label:org.flatpak.ref:exists=1`, in case a client would fail to include that in its query.  I thought that <https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md> mentions that that should always be included in queries, but it looks like it doesn't actually mention that, and that I only learned from private communication with @owtaylor that (a) clients should always include `label:org.flatpak.ref:exists=1` in their queries and (2) servers should always assume that `label:org.flatpak.ref:exists=1` is implicitly assumed for every query.  Maybe we should add something about that to <https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md>.)

pulp_container/app/registry_api.py
stbergmann: Sorry, I don't understand what you mean with that.

ipanova:
> > Another question, do flatpack clients account for the registry auth? In that case you need to make it work with token auth, the view needs to have `ContainerRegistryApiMixin`
> 
> I don't really know. Maybe @owtaylor can give some information here?
> 
> From how I implemented things in this PR and tried them out in a local oci-env instance, I naively assumed that per the design of Pulp the `/index/static` endpoint will by necessity be behind some kind of authentication. I found that in my local oci-env instance it can be reached via http://admin:changeme@localhost:5001/index/static (and I patched my local `flatpak` command-line client to make such URLs containing user:password work there, see [flatpak/flatpak#5446](https://github.com/flatpak/flatpak/issues/5446) "[Bug]: Support oci remotes with URLs containing user:password"). At https://matrix.to/#/!RfXaBjokqHAbzZrgHz:matrix.org/$reAbU4QuJaBhruvvjtAAZvJtp-3iIEHavOzFMYoudK4?via=matrix.org&via=gnome.org&via=kde.org, I noticed: "I see there's some --authenticator-name etc. arguments to flatpak remote-add, but can't find any documentation about that, so no idea it could help me. Any ideas?" But didn't get much of an authoritative response.
> 
> So, first question presumably is whether, per the design of Pulp and Satellite, these `/index/dynamic` and `/index/static` endpoints should be behind some kind of authentication in the first place?

Per pulp-container design, as long as this endpoint is considered a registry endpoint it should be backed up by token auth.
Satellite has it's own auth layer, so this is a separate unrelated question to answer, for now.

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+        req_label_values = {}
+        for key, values in request.query_params.lists():
+            if key == "repository":
+                req_repository = values[0]
+                if any(v != req_repository for v in values[1:]):
```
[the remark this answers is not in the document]
stbergmann: From private communication between @owtaylor and me:
"sbergman:  At <https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md#requests>, for an image to match the given query parameters means that it matches their conjunction, right?  (i.e., `?architecture=amd64&architecture=arm64` will necessarily return an empty list, not a list containing images that mach either of the architectures)
"Owen Taylor:  Yeah, I think so. I'm not sure I really considered repeated arguments."

(Yet something else to clarify in <https://github.com/owtaylor/flagstate/pull/6>.  Will do.)

So that's why I check in all these cases that if a query parameter like `repository` is given multiple times, that all the given values are the same, or else the request is unsatisfiable.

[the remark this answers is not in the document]
ipanova: gotcha, this info is helpful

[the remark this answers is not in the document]
ipanova: really depends on the client's expectation but what we could do instead of checking all the cases of repo query param to have same values, just pick the first value and ignore rest

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+                req_os = values[0]
+                if any(v != req_os for v in values[1:]):
+                    unsatisfiable = True
+                    break
+            elif key == "architecture":
+                req_architecture = values[0]
+                if any(v != req_architecture for v in values[1:]):
+                    unsatisfiable = True
+                    break
+            elif key.startswith("label:"):
```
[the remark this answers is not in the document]
stbergmann: Yes, my understanding is that that should be supported, and that the current code supports it correctly.

pulp_container/app/registry_api.py
owtaylor: > should the second label be added here too? https://github.com/owtaylor/flatpak-indexer/blob/main/flatpak_indexer/indexer.py#L85

This code is specifically about converting the "new" type of Flatpaks with metadata in labels to be compatible with older clients (older clients can be ignored at this point.) In terms of creating labels, there is no special handling for any labels - all should be included.

> (The code above, req_label_exists = {"org.flatpak.ref"}, is about presetting the set of requested label:...:exists=1 query parameters with label:org.flatpak.ref:exists=1, in case a client would fail to include that in its query. [...]  I only learned from private communication with @owtaylor that (a) clients should always include label:org.flatpak.ref:exists=1 in their queries and (2) servers should always assume that label:org.flatpak.ref:exists=1 is implicitly assumed for every query. Maybe we should add something about that to https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md.)

I think what I meant was more along the lines of (1) since the only client that talks this protocol is the Flatpak client, it *will* always include `label:org.flatpak.ref:exists=1`, and (2) since the only clients that talk this protocol is the Flatpak client, servers *can* assume that org.flatpak.ref:exists=1.

(For Fedora and RHEL, we pre-generate indexes for this case, and 404 if we get a query without the restriction.)

But unless the 2) actually simplifies things, you definitely don't need to make that assumption.

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
owtaylor: > From private communication between @owtaylor and me:
> "sbergman: At https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md#requests, for an image to match the given query parameters means that it matches their conjunction, right? (i.e., `?architecture=amd64&architecture=arm64` will necessarily return an empty list, not a list containing images that mach either of the architectures)
> "Owen Taylor: Yeah, I think so. I'm not sure I really considered repeated arguments."

So, I dug through the Flagstate code, and the interpretation for repeated arguments is OR rather than AND. Which is definitely also the more useful interpretation.  My preference would be:

 * Document the OR implementation in the protoocl
 * If it's easy to implement the OR in the Pulp code, do that
 * Otherwise, error out (with a 501 Not Implemented perhaps) if a repeated term is found

The Flatpak client will *not* pass in repeated terms, so there is no need to go through a lot of work to implement them.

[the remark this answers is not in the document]
stbergmann: So, just to clarify, the rule would be:  An image I matches a set S of query parameters iff:  For each K=V in S:  There exists a K'=V' in S with:  K==K' and I matches V'.

(That is, `?label:eat=apple&label:eat=orange&label:drink=water` matches all images that have both: (a) an `eat` label of either `apple` or `orange`; and (b) a `drink` label of `water`.)

stbergmann:
> So the ideal thing would be if we could just make the index unauthenticated - if that doesn't violate pulp principles too badly. If we are doing this, then requiring `label:org.freedesktop.exists=1` might make sense to not leak information about _non_ Flatpak containers in the registry.

@ipanova Would that be possible, to "make the index unauthenticated"?

@owtaylor I assume you meant `label:org.flatpak.ref:exists=1` instead of `label:org.freedesktop.exists=1`.

ipanova:
@owtaylor if it is expected to have index unauthenticated then the assumption that the index is going to expose only public repos. Public repos can be are accessible by everyone. It does not make sense to go through all the content repos( perf. overhead) , expose them( information leak) and then leave it to the user session to provide it's credentials which might or might not give a 200 on content pull.
If my read on this is same as your then we're on the same page.

pulp_container/app/cache.py
```diff
@@ -72,3 +74,20 @@ def find_base_path_cached(request, cached):
...
+    def __init__(self, expires_ttl=None, auth=None):
+        updated_keys = (USER_KEY, QUERY_KEY)
+        super().__init__(
+            base_key="/index/static", expires_ttl=expires_ttl, keys=updated_keys, auth=auth
+        )
+
+    def make_key(self, request):
+        """Make a key composed of the request's query."""
+        all_keys = {
+            USER_KEY: request.user.get_username(),
```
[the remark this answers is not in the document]
ipanova: from the recent info we got, since index is unauthenticated then probably it does not make sense to register the user key?

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+                    if any(v != values[0] for v in values[1:]):
+                        unsatisfiable = True
+                        break
+                    label = key[len("label:") :]
+                    req_label_values[label] = values[0]
+            else:
+                # This covers unknown query parameters (which we consider as unsatisfiable for
+                # simplicity, rather than causing some error response) as well as any annotation:...
+                # query parameters (whose presence necessarily cause the request to be
+                # unsatisfiable, as this implementation never announces any Annotations):
```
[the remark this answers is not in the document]
ipanova: @owtaylor PTAL

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+                unsatisfiable = True
+                break
+
+        results = []
+
+        if not unsatisfiable:
```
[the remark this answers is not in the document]
ipanova: @owtaylor PTAL

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
stbergmann: addressed with <https://github.com/pulp/pulp_container/pull/1301/commits/719a9f599ba647de371a21450ab45f8c2af52366> "Fix logic for repeated query parameters" now

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
stbergmann: addressed with <https://github.com/pulp/pulp_container/pull/1301/commits/d7706aa3f3545a6090a3cea55c6a4fe2e755549e> "Send 400 reply for bad requests" now

ipanova:
Here's my interpretation and summary of the outstanding work:

- [ ] Add docs
- [ ] Add new boolean setting FLATPACK_INDEX=False, it will make flatpack index url conditional
- [ ] move https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md to https://github.com/flatpack/flagstate/blob/master/docs/protocol.md
- [ ] update flagstate protocol.md with the recent info/changes
- [ ] tailor the code to the Flatpack client. i.e:
    1.  client sends always only *one* (not repeated query param): repo, tag, label, os, arch, annotation. If query param is repeated raise 400
    2. query param values are not input validated i.e. `os=blabla` would not raise error rather return a []
    3. query param value has always a single value (comma separated list is not understood), otherwise raise 400, not sure on this one, @owtaylor 
    4.  if `flatpack` label is not present in the query raise 400
    5.  for unknown query param raise 400 (only repo, tag, label, os, arch, annotation are understood)
 - [ ] index/static and index/dynamic are non authed urls. It means only public repo are exposed.
 - [ ] figure out how to treat `annotations` query param https://github.com/pulp/pulp_container/pull/1301#discussion_r1235440233

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+                    break
+            elif key == "tag":
+                req_tags.update(values)
+            elif key == "os":
+                req_os = values[0]
+                if any(v != req_os for v in values[1:]):
+                    unsatisfiable = True
+                    break
+            elif key == "architecture":
+                req_architecture = values[0]
```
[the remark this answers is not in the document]
stbergmann: So with the updated understanding about multiple `architecture=...` query parameters, my understanding here would be that: if no `architecture=...` query parameters are given then all images are returned unfiltered; otherwise, all images are returned that match one of the given `architecture=...` query parameters.  (As is now implemented by <https://github.com/pulp/pulp_container/pull/1301/commits/719a9f599ba647de371a21450ab45f8c2af52366> "Fix logic for repeated query parameters".)

@owtaylor ?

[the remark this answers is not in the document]
ipanova: yes, that would be my understanding of OR logic in the repeated query params

ipanova:
> > ```
> >   1. client sends always only _one_ (not repeated query param): repo, tag, label, os, arch, annotation. If query param is repeated raise 400
> > ```
> 
> It was easy to address that properly (i.e., by not raising 400 but rather implementing the intended semantics) with [719a9f5](https://github.com/pulp/pulp_container/commit/719a9f599ba647de371a21450ab45f8c2af52366) "Fix logic for repeated query parameters", so I at least would be fine with leaving the code that way.

I am fine either way as long as we all agree what to follow. Updated my previous comment, check it out

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
owtaylor: > reading explanation from Owen i read it as you whether ignore `annotations` query param without making the request invalid or treat interchangeably annotations as labels if they are present in the request
> 
> > So we switche
> > dover to using labels for the metadata. We have the ability in our pipeline to
> > generate either or both, so you might see images with both - but the only
> > thing you need to support is labels.
> 
> can we get more clarification here?

There are two options I'd consider valid:

 * Treat requests with annotation parameters as invalid (501 or 400)
 * Fully handle annotation parameters based on the actual annotations in OCI images

I would not suggest ignoring annotation parameters or applying them to the labels.

If Pulp isn't currently indexing annotations in the database, then the first one seems easier and is perfectly sufficient for Flatpak. All support for annotations was dropped from Flatpak prior to the 1.6.0 release in Dec 2019.

pulp_container/app/registry_api.py
```diff
@@ -475,6 +479,161 @@ def get_queryset(self, *args, **kwargs):
...
+                # query parameters (whose presence necessarily cause the request to be
+                # unsatisfiable, as this implementation never announces any Annotations):
+                unsatisfiable = True
+                break
+
+        results = []
+
+        if not unsatisfiable:
+            # This part copied from CatalogView:
+            queryset = models.ContainerDistribution.objects.all().only("base_path")
```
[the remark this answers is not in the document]
stbergmann: <https://github.com/pulp/pulp_container/pull/1301/commits/2818e70ae3b782923b2ed77d3e92bdb37f14ea16> "Only index public repositories" removed this part of the code now

pulp_container/app/registry_api.py
stbergmann: <https://github.com/pulp/pulp_container/pull/1301/commits/2818e70ae3b782923b2ed77d3e92bdb37f14ea16> "Only index public repositories" removed this part of the code now

pulp_container/app/registry_api.py
stbergmann: see <https://github.com/pulp/pulp_container/pull/1301/commits/12b27910923d9b3b3fd57cdb70fb068ce20cf18f> "Rename repositories -> distributions for clarity"

pulp_container/app/models.py
```diff
@@ -612,6 +626,22 @@ def redirect_to_content_app(self, url):
...
+        AFTER_UPDATE,
+        when_any=[
+            "base_path",
+            "private",
+            "repository",
+            "repository_version",
+        ],
+        has_changed=True,
+    )
+    def invalidate_flatpak_index_cache(self):
```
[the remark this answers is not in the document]
stbergmann: done, <https://github.com/pulp/pulp_container/pull/1301/commits/17edb840431d0e322f6595141ea32f39a252d68f> "No need to invalidate cache on repository creation/deletion"

pulp_container/app/models.py
stbergmann: ...and `SyncContentCache` is derived from `Cache` and happens to be a usable substitute here; done with <https://github.com/pulp/pulp_container/pull/1301/commits/4a4ca14a36a015414f7a422f2e4c01009f6162bf> "Use derived class SyncContentCache"

stbergmann:
> [...]
> 
> > With this reasoning we need to introduce a so called feature flag, probably a pulp setting that will enabling or disabling the flatpack index support in the pulp-container registry. By default it should be disabled.
> 
> i.e.,
> 
> [...]
> 
> > Add new boolean setting FLATPACK_INDEX=False, it will make flatpack index url conditional
> 
> I assume I'd add a line `FLATPAK_INDEX = False` to `pulp_container`'s `pulp_container/app/settings.py`. But then, how would I set that to true in my local oci_env? Adding a line `PULP_FLATPAK_INDEX=true` to `oci_env`s `base/pulp_config.env` and/or `profiles/pulp_container_base/pulp_config.env` apparently doesn't enable it. (And https://github.com/pulp/oci_env#profiles leaves me clueless. What is a profile, exactly? How do I use one? Do I use one per plugin? ...)

done now with <https://github.com/pulp/pulp_container/pull/1301/commits/097f8cd8c922306336d9805586fc53413c8d5baa> "Make support for FLATPAK_INDEX conditional", and enabling it in oci_env by adding `PULP_FLATPAK_INDEX=true` to `base/pulp_config.env` works as intended (no idea why it didn't work for me initially)

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
stbergmann: * I do not know any existing `/index/static` flatpak indexer endpoint that reports Annotations. 
* At least <https://registry.fedoraproject.org/index/static?label:org.flatpak.ref:exists=1> does not report any.  However, for an image I that it reports, the corresponding <https://registry.fedoraproject.org/v2/$base_path/manifests/$digest> is either of type `application/vnd.docker.distribution.manifest.v2+json` or `application/vnd.oci.image.manifest.v1+json`.  And if it is of the latter type, then (Observation 1:) the manifest does contain Annotations, exactly duplicating the Labels of that image I as reported by the flatpak indexer. 
* So when such an image I is imported into the existing pulp_container implementation, the Annotations in the `application/vnd.oci.image.manifest.v1+json` are available in the manifest blob, and I could in turn report them as Annotations from my pulp_container flatpak indexer endpoint.  But as those Annotations are exactly the same (due to Observation 1) as the Labels already reported by my pulp_container flatpak indexer endpoint, I see little value in doing so.
* So my pulp_container flatpak indexer endpoint never reports any Annotations (apparently in line with other existing flatpak indexer endpoints, see above).
* So for an `?annotation:...` request, my pulp_container flatpak indexer endpoint could conceivably either (1) return an empty set of responses (as no images can match), or (2) return some 400 or similar error response.  I decided for (1) because that looks like the logically most plausible option to me, but if (2) is preferred for some reason, then we could use that.  @owtaylor ?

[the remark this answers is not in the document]
ipanova: I would treat this as invalid request:
1) All support for annotations was dropped from Flatpak prior to the 1.6.0 release in Dec 2019. I would even mention this in docs.
2) Flatpack client itself does not send annotations in the request.

If you're not parsing manifest for annotations (I did not find any logic in your code)  you can't really be sure whether there will be any images matched or not. Today you cannot find such examples, however, you cannot go forward with this assumption since in future this can change.

pulp_container/app/registry_api.py
```diff
@@ -475,6 +480,138 @@ def get_queryset(self, *args, **kwargs):
...
+                req_architectures = values
+            elif key.startswith("annotation:"):
+                if key.endswith(":exists") and any(v != "1" for v in values):
+                    raise ParseError()
+                req_annotations = True
+            elif key.startswith("label:"):
+                if key.endswith(":exists"):
+                    if any(v != "1" for v in values):
+                        raise ParseError()
+                    label = key[len("label:") : len(key) - len(":exists")]
```
ipanova, line 516: you can do this:
```
In [1]: x='label:<label>:exists'

In [2]: x.split(":")
Out[2]: ['label', '<label>', 'exists']

In [4]: x.split(":")[1]
Out[4]: '<label>'
```

pulp_container/app/registry_api.py
```diff
@@ -475,6 +480,138 @@ def get_queryset(self, *args, **kwargs):
...
+                    req_label_exists.add(label)
+                else:
+                    label = key[len("label:") :]
```
ipanova, line 519: same here key.split(":")[1]

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
stbergmann: > If you're not parsing manifest for annotations (I did not find any logic in your code)

That's correct.

> you can't really be sure whether there will be any images matched or not.

That depends on how you define matching and how you define the link between  `application/vnd.oci.image.manifest.v1+json` manifest annotations and `/index/static` annotations.  (Whether or not image I should match an `?annotation:...` request when the information returned by `/index/static` about image I: (1) does not list any annotations, but (2) references an `application/vnd.oci.image.manifest.v1+json` manifests that does contain annotations matching the `?annotation:...` request.) 

> Today you cannot find such examples,

That's not correct:  Both e.g. the existing <https://registry.fedoraproject.org/index/static>, as well as my implementation here, both do reference `application/vnd.oci.image.manifest.v1+json` manifests that contain annotations, see my "Observation: 1" above.  (They just both never include those manifest's annotations in their `/index/static` JSON replies.)

>  however, you cannot go forward with this assumption since in future this can change.

see above


But anyway, I'll just change my implementation to consider any annotation requests as invalid and be done with it.

pulp_container/app/registry_api.py
```diff
@@ -475,6 +480,138 @@ def get_queryset(self, *args, **kwargs):
...
+                    media_type__in=(models.MEDIA_TYPE.MANIFEST_V2, models.MEDIA_TYPE.MANIFEST_OCI),
+                )
+                for manifest in manifests:
+                    with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
+                        raw_data = file.read()
+                    config_data = json.loads(raw_data)
+                    labels = config_data.get("config", {}).get("Labels")
+                    if not labels:
+                        continue
+                    if any(label not in labels.keys() for label in req_label_exists):
```
ipanova, line 557: it seems like you require here all specified labels to be present compared to os,arch,tags were you satisfy at least 1 match, is my read correct? and why such difference in logic only for labels?

docs/workflows/flatpak.rst
```diff
@@ -0,0 +1,62 @@
...
+specified by `the Flatpak index protocol
+<https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md>`_.  This is not enabled by
+default.  To enable it, define ``FLATPAK_INDEX = True`` in the settings file.
+
+Clients like the ``flatpak`` command-line tool or the GNOME Software application will typically
+query the ``/index/static`` endpoint, which is intended to be called repeatedly with identical query
+parameters, and whose responses are meant to be cached.  The ``/index/dynamic`` endpoint serves
+exactly the same content, but is intended for one-off requests that should not be cached.  These
+endpoints can be accessed without authentication.  They only provide information about public
+distributions.
```
ipanova, line 17: s/distributions/repositories

docs/workflows/flatpak.rst
```diff
@@ -0,0 +1,62 @@
...
+
+The two endpoints support a number of query parameters (``architecture``, ``tag``, ``label``, etc.),
+see the protocol specification for details.  Two notes:
+
+* Every request must include a ``label:org.flatpak.ref:exists=1`` query parameter.  This acts as a
+  marker to only report Flatpak content, and to exclude other container content that may also be
+  provided by the Pulp instance.
+
+* This implementation does not support annotations, so including any ``annotation`` query parameter
+  will result in an empty response.  Use ``label`` query parameters instead.
```
ipanova, line 27: result in 400

ipanova: can you add a note that this is in line for flatpack clients works?

pulp_container/app/registry_api.py
ipanova: i misread, ignore

ipanova:
Apart from the last 1-2 questions I am done with the review. I worry about performance though, it would be good to populate pulp instance with some flatpack content and get some numbers when querying json indexer endpoint.
As next step we plan to get onto https://github.com/pulp/pulp_container/issues/1296 and https://github.com/pulp/pulp_container/issues/1288 so the data will be read and loaded from DB instead of file on disk.

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
stbergmann: > But anyway, I'll just change my implementation to consider any annotation requests as invalid and be done with it.

done, <https://github.com/pulp/pulp_container/pull/1301/commits/1f6b2cbe768b0c3e7871a3956385b8ab36d0d3b9> "Make annotation query parameters cause 400 response"

pulp_container/app/registry_api.py
stbergmann: But that does not work when `<label>` contains colons.  (<https://github.com/owtaylor/flagstate/blob/master/docs/protocol.md> is vague on what those labels can be; and there would of course be an ambiguity that `label:foo:exists=1` is interpreted as a check whether label `foo` exists, rather than that label `foo:exists` has the value `1`, but I think that is more a shortcoming of the protocol as defined there, rather than an indication that labels may not contain colons.  The closest we get to an actual definition of what those labels can be is probably <https://github.com/opencontainers/image-spec/blob/main/config.md> "Labels object, OPTIONAL: The field contains arbitrary metadata for the container. This property MUST use the [annotation rules](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules).", where the referenced rules merely state that "Keys SHOULD be named using a reverse domain notation - e.g. com.example.myKey", but does not rule out colons.)

pulp_container/app/registry_api.py
stbergmann: see above

pulp_container/app/registry_api.py
stbergmann: Oops, thanks for catching that.  Done with <https://github.com/pulp/pulp_container/commit/f5594407655fd97635edd1fdf9acd873853137fc> "Fix logic also for repeated label query parameters".

pulp_container/app/registry_api.py
ipanova: hmmm to be honest based on generic guidelines, I  never encountered labels/annotation with colons https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations

However , you are right, these are unwritten conventions and if someone really wants to put a colon into a label or annotation nothing will prevent him from doing so:
```
 podman build -t scriptcamp/nginx . --label "dynamic:label2=value1"  --annotation "dynamic:label1=value1"
STEP 1/3: FROM nginx:alpine
STEP 2/3: COPY index.html /usr/share/nginx/html/index.html
--> Using cache 696dde108fecd53a8aa0145f184dcad3f175a1f993c597b5f4d28909ceadd761
--> 696dde108fe
STEP 3/3: LABEL "dynamic:label2"="value1"
COMMIT scriptcamp/nginx
--> 1c3a9e454b7
Successfully tagged localhost/scriptcamp/nginx:latest
1c3a9e454b7e0aa1a0a6e8061af166bb813a46fd2f9d33c6fda857f3cae7c273

$ podman inspect scriptcamp/nginx:latest | jq -r .[].Annotations
{
  "dynamic:label1": "value1",
  "org.opencontainers.image.base.digest": "sha256:3c98d55fa19047f2f69159a0543661848738f64f637bb6e60c95c0ce0a2c081a",
  "org.opencontainers.image.base.name": "docker.io/library/nginx:alpine"
}
(pulp) [vagrant@pulp3-source-fedora36 nginx-image]$ podman inspect scriptcamp/nginx:latest | jq -r .[].Config.Labels
{
  "dynamic:label2": "value1",
  "io.buildah.version": "1.28.0",
  "maintainer": "NGINX Docker Maintainers <docker-maint@nginx.com>"
}

```

pulp_container/app/registry_api.py
stbergmann: Oh my, now I was confused.  According to <https://github.com/stbergmann/flagstate/commit/a0e3b3086db37044cd9052b417ed6fea80569e5f> "Clarify how query parameters are matched", the original code was exactly matching the clarified specification.  For `?label:l1:exists=1&label:l2:exitst=1&tag=t1&tag=t2` to match: (`label:t1:exists=1` must match "the image has a label `l1`") AND (`label:t2:exitsts=1` must match "the image has a lable `l2`") AND (`tag=t1&tag=t2` must match "the image has at least one of the tags `t1`, `t2`").

docs/workflows/flatpak.rst
stbergmann: done, <https://github.com/pulp/pulp_container/pull/1301/commits/73b728389570ba329b51a46afbad0e55498eec89> "Documentation improvements"

docs/workflows/flatpak.rst
stbergmann: done, <https://github.com/pulp/pulp_container/pull/1301/commits/1f6b2cbe768b0c3e7871a3956385b8ab36d0d3b9> "Make annotation query parameters cause 400 response" and <https://github.com/pulp/pulp_container/pull/1301/commits/73b728389570ba329b51a46afbad0e55498eec89> "Documentation improvements"

pulp_container/app/registry_api.py
```diff
@@ -475,6 +480,134 @@ def get_queryset(self, *args, **kwargs):
...
+            images = []
+            if distribution.repository:
+                repository_version = distribution.repository.latest_version()
+            elif distribution.repository_version:
+                repository_version = distribution.repository_version
+            tags = models.Tag.objects.filter(pk__in=repository_version.content)
+            manifests = models.Manifest.objects.filter(
+                pk__in=repository_version.content,
+                media_type__in=(models.MEDIA_TYPE.MANIFEST_V2, models.MEDIA_TYPE.MANIFEST_OCI),
+            )
```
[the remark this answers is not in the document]
owtaylor: I think this is the right approach even if if we are trying to get an accurate Tags list for the index - I'd first find the matching images, *then* figure out the Tags that apply to the image. But actually, I would suggest not providing an accurate Tags, if it requires extra database queries and work. Flatpak doesn't care about the value of Tags, and it's only a SHOULD in the spec.
  
> Tags: List of all tags applied to the image. This appears only for images appearing in the Images list of a repository, and not for images in an image List. All tags applied to the image SHOULD be returned, even if a tag parameter in the query limits the images returned by tag.

pulp_container/app/registry_api.py
stbergmann: Fixed again with <https://github.com/pulp/pulp_container/pull/1301/commits/a51d0e452fa0a0592d056f5709e43fb2a194fa82> "Revert 'Fix logic also for repeated label query parameters'".

pulp_container/app/registry_api.py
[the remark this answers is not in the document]
stbergmann: > I would note here that a query from a Flatpak client will _always_ include a tag= query parameter - and this can be the biggest limiter on the resulting content set.
> 
> My instinct, from a performance point of view, would be to accumulate a set of manifests that are tagged with the tags in the query (directly or indirectly), then only parse _those_ manifests to apply further filters such as label, os, architecture.
> 
> This could easily result in factors of 100 less config blobs being parsed, since it wouldn't be shocking to have hundreds of historical tags.

done with <https://github.com/pulp/pulp_container/pull/1301/commits/bd60171eb85a95d3a0f70d23dadea97ea334da14> "Optimize for common case of filtering on requested tags"

> The request will also always include an architecture. I think it would be fine to apply the os/architecture filters when querying for manifests tagged by manifest list, despite the wording in the spec that the os/architecture come from the config.json.
> 
> This would reduce parsing config blobs by up to a factor of 4 - if someone was literally mirroring manifest lists with all architectures from registry.redhat.io - I don't know if there's any provision for subsetting architectures when mirroring.

done with <https://github.com/pulp/pulp_container/pull/1301/commits/9c9ac62758ff19a15afbea2ecfd53a42530133f3> "Also filter on os/arch in recurse_through_manifest_lists"

> I think this is the right approach even if if we are trying to get an accurate Tags list for the index - I'd first find the matching images, _then_ figure out the Tags that apply to the image. But actually, I would suggest not providing an accurate Tags, if it requires extra database queries and work. Flatpak doesn't care about the value of Tags, and it's only a SHOULD in the spec.

done with <https://github.com/pulp/pulp_container/pull/1301/commits/0942e37ad485509bf28a51ad75f4e145b528dd82> "To improve performance, only report requested tags"

docs/workflows/flatpak.rst
```diff
@@ -0,0 +1,64 @@
+.. _flatpak-workflow:
+
+Hosting Flatpak Content in OCI Format
+=====================================
+
+Pulp can host Flatpak application and runtime images that are distributed in OCI format.  To make
```
lubosmj, line 6:
```suggestion
Pulp can host Flatpak applications and runtime images that are distributed in OCI format.  To make
```

docs/workflows/flatpak.rst
```diff
@@ -0,0 +1,64 @@
...
+see the protocol specification for details.  Two notes:
+
+* Every request must include a ``label:org.flatpak.ref:exists=1`` query parameter.  This acts as a
+  marker to only report Flatpak content, and to exclude other container content that may also be
+  provided by the Pulp instance.
+
+* This implementation does not support annotations.  Including any ``annotation`` query parameters
+  will result in a 400 failure response.  Use ``label`` query parameters instead.  (Existing clients
+  like the ``flatpak`` command-line tool never issue requests including any ``annotation`` query
+  paramters.)
```
lubosmj, line 29:
```suggestion
  parameters.)
```

pulp_container/app/cache.py
```diff
@@ -72,3 +73,19 @@ def find_base_path_cached(request, cached):
         except ObjectDoesNotExist:
             raise RepositoryNotFound(name=path)
         return distro.base_path
+
+
+class FlatpakIndexStaticCache(SyncContentCache):
+    def __init__(self, expires_ttl=None, auth=None):
+        updated_keys = (QUERY_KEY,)
+        super().__init__(
+            base_key="/index/static", expires_ttl=expires_ttl, keys=updated_keys, auth=auth
```
lubosmj, line 82: `EXPIRES_TTL` is set to 600 seconds by default if we pass there `None`. Is this expected?

pulp_container/app/registry_api.py
```diff
@@ -473,6 +477,135 @@ def get_queryset(self, *args, **kwargs):
...
+                        raise ParseError()
+                    label = key[len("label:") : len(key) - len(":exists")]
+                    req_label_exists.add(label)
+                else:
+                    label = key[len("label:") :]
+                    req_label_values[label] = values
+            else:
+                # In particularly, this covers any annotation:... parameters, which this
+                # implementation does not support:
+                raise ParseError()
```
lubosmj, line 529: Is it possible to return an error message as well (https://www.django-rest-framework.org/api-guide/exceptions/#parseerror)?

pulp_container/app/registry_api.py
```diff
@@ -473,6 +477,135 @@ def get_queryset(self, *args, **kwargs):
...
+    Handles requests to the /index/static endpoint
+    """
+
+    @FlatpakIndexStaticCache()
+    def get(self, request):
+        response = super().get(request)
+        response.accepted_renderer = JSONRenderer()
+        response.accepted_media_type = JSONRenderer.media_type
+        response.renderer_context = {}
+        response.render()
```
lubosmj, line 605: Is this necessary to do?

pulp_container/app/registry_api.py
```diff
@@ -473,6 +477,135 @@ def get_queryset(self, *args, **kwargs):
...
+            distributions = distributions.filter(base_path__in=req_repositories)
+
+        results = []
+        for distribution in distributions:
+            images = []
+            if distribution.repository:
+                repository_version = distribution.repository.latest_version()
+            else:
+                repository_version = distribution.repository_version
+            tags = models.Tag.objects.filter(pk__in=repository_version.content)
```
lubosmj, line 546:
```suggestion
            tags = models.Tag.objects.select_related("tagged_manifest").filter(pk__in=repository_version.content)
```

pulp_container/app/registry_api.py
```diff
@@ -473,6 +477,135 @@ def get_queryset(self, *args, **kwargs):
...
+        mlms = manifest.listed_manifests.through.objects.filter(image_manifest__pk=manifest.pk)
+        if oss:
+            mlms.filter(os__in=oss)
+        if architectures:
+            mlms.filter(architecture__in=architectures)
+        for mlm in mlms:
+            self.recurse_through_manifest_lists(
+                tag, mlm.manifest_list, oss, architectures, manifests
+            )
+        if manifest.media_type in (models.MEDIA_TYPE.MANIFEST_V2, models.MEDIA_TYPE.MANIFEST_OCI):
```
lubosmj, line 498: Do you think moving this `if` condition to the top of the method is more efficient?

pulp_container/app/registry_api.py
```diff
@@ -473,6 +477,135 @@ def get_queryset(self, *args, **kwargs):
...
+                    continue
+                if any(
+                    labels.get(label) not in values for label, values in req_label_values.items()
+                ):
+                    continue
+                if not req_tags:
+                    direct_tags = models.Tag.objects.filter(
+                        tagged_manifest__pk=manifest.pk
+                    ).values_list("name", flat=True)
+                    tagged.update(direct_tags)
```
lubosmj, line 577: Here we can return tags that are not present in the current repository. How about removing the need for updating `tagged` at all?

pulp_container/app/registry_api.py
```diff
@@ -473,6 +477,135 @@ def get_queryset(self, *args, **kwargs):
...
+            manifests = {}  # mapping manifests to sets of encountered tag names
+            for tag in tags:
+                self.recurse_through_manifest_lists(
+                    tag.name, tag.tagged_manifest, req_oss, req_architectures, manifests
+                )
+            for manifest, tagged in manifests.items():
+                with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
+                    raw_data = file.read()
+                config_data = json.loads(raw_data)
+                labels = config_data.get("config", {}).get("Labels")
```
lubosmj, line 558:
```suggestion
                labels = config_data.get("config", {}).get("Labels", [])
```

`labels` can be `None` here. On line 585 this might not be appropriate, right? If the protocol does not define its behaviour, we can omit the entry.

lubosmj (commented):
Great job! We are getting closer to the end of our journey. Please, ensure the feature is advertised as a tech preview (https://docs.pulpproject.org/pulp_container/tech-preview.html).

pulp_container/app/registry_api.py
lubosmj: Also, I am not quite sure if it is necessary to recursively go through manifests because there should be just one manifest list within the stack.

docs/workflows/flatpak.rst
stbergmann: that should be fine, short for "application images and runtime images"

pulp_container/app/cache.py
stbergmann: I have no idea what a good value would be, and naively assumed the default would fit.  Any suggestions?

pulp_container/app/registry_api.py
stbergmann: that's functionally equivalent and "just" a performance improvement?

pulp_container/app/registry_api.py
stbergmann: should be fine, see the following `if not labels:  continue`

pulp_container/app/cache.py
lubosmj: I would make it last even for 1 or 6 hours. It is about your preferences.

pulp_container/app/registry_api.py
lubosmj: This should be a performance improvement.

We hit the database after doing `for tag in tags` first. Then, we hit the database every iteration when we try to work with the tagged manifests. By selecting the related field, we load the referenced objects ahead of time.

docs/workflows/flatpak.rst
stbergmann: done in c6605fe350000eaa3c0de35d05f73af25453405d

pulp_container/app/cache.py
stbergmann: I think I'm completely incompetent to give any preferences here :)  (And my understanding would be that there doesn't need to be any expiration anyway if the invalidation logic is correct?)  But lets settle with an 8 hour expiration time as done in c6605fe350000eaa3c0de35d05f73af25453405d.

pulp_container/app/registry_api.py
stbergmann: yes, added comment in c6605fe350000eaa3c0de35d05f73af25453405d

pulp_container/app/registry_api.py
stbergmann: done in c6605fe350000eaa3c0de35d05f73af25453405d

pulp_container/app/registry_api.py
stbergmann: Moved the `if` part to the top in c6605fe350000eaa3c0de35d05f73af25453405d.
I don't know if it is somehow guaranteed that there will only ever be one level of manifest lists, but why not keep the code simple and generic and let it thus automatically cater for multiple levels of lists.

pulp_container/app/registry_api.py
stbergmann: This was a remnant of a previous solution which, on second thought, appears to indeed no longer be necessary.  Removed in c6605fe350000eaa3c0de35d05f73af25453405d.

stbergmann:
> Please, ensure the feature is advertised as a tech preview (https://docs.pulpproject.org/pulp_container/tech-preview.html).

done in c6605fe350000eaa3c0de35d05f73af25453405d

.github/workflows/scripts/before_script.sh
```diff
@@ -39,7 +39,7 @@ cmd_prefix bash -c "usermod -a -G wheel pulp"
 SCENARIOS=("pulp" "performance" "azure" "gcp" "s3" "stream" "generate-bindings" "lowerbounds")
 if [[ " ${SCENARIOS[*]} " =~ " ${TEST} " ]]; then
   # Many functional tests require these
-  cmd_prefix dnf install -yq lsof which
+  cmd_prefix dnf install -yq dbus-daemon flatpak lsof which
```
stbergmann, line 42: Should I be concerned about the
```
# WARNING: DO NOT EDIT!
#
# This file was generated by plugin_template, and is managed by it. Please use
# './plugin-template --github pulp_container' to update this file.
#
# For more info visit https://github.com/pulp/plugin_template
```
at the top of this file? Should these additional `dnf install` rather be done somewhere else?

lubosmj: Oh, nice catch! Yes, you should create a new file called `.github/workflows/scripts/post_before_script.sh` where you will install the additional packages in the same way. This script is called from `before_script.sh`.

stbergmann: done in 914492c284d764a42378cf968b50c93ea6c4a9a9

pulp_container/tests/functional/api/test_flatpak.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    # Cleanup:
+    subprocess.run(["flatpak", "--user", "uninstall", "--noninteractive", "net.fishsoup.Hello"])
+    subprocess.run(
+        ["flatpak", "--user", "uninstall", "--noninteractive", "net.fishsoup.BusyBoxPlatform"]
+    )
+    subprocess.run(["flatpak", "--user", "remote-delete", "pulptest"])
+    namespace = container_namespace_api.list(name="pulptest").results[0]
+    add_to_cleanup(container_namespace_api, namespace.pulp_href)
```
lubosmj, line 52: Please, move the cleanup right after `local_registry.tag_and_push(image_path2, "pulptest/oci-net.fishsoup.hello:latest")`. The cleanup is executed when the test finishes: https://github.com/pulp/pulpcore/blob/bbb30b7a5de3cab3ac13e1913facb33f1292e2fe/pulpcore/tests/functional/__init__.py#L926. In case we fail again when running the flatpak application, other tests shall not be affected.

stbergmann: done in a0ad4ac0a28e9c61dbf3cec17e3e298c96fe8b3f

pulp_container/app/registry_api.py
```diff
@@ -473,6 +477,134 @@ def get_queryset(self, *args, **kwargs):
...
+
+        return Response(data={"Registry": settings.CONTENT_ORIGIN, "Results": results})
+
+
+class FlatpakIndexStaticView(FlatpakIndexDynamicView):
+    """
+    Handles requests to the /index/static endpoint
+    """
+
+    @FlatpakIndexStaticCache(expires_ttl=60 * 60 * 8)  # somewhat arbitrary 8h expiration time
```
ipanova, line 597: As pointed out in previous comment the default cache expiration is 10mins, or the value that is explicitly configured in Pulp cache settings `EXPIRES_TTL`.
With this hardcoded approach any cache expiration setting will be ignored. This is not very explicit. I am suggesting a) follow the regular setting values or b) if we want to keep this to 8h exp. time then can we please at least document this, and also mention that this is not *changeable value* either. c) introduce a separate flatpak_cache_expires_ttl setting.

@owtaylor do you have any insights or suggestions of what would be a reasonable caching value for the static index?

owtaylor: 8h seems far too long to me - since it doesn't look like the cache is ever invalidated. It would be very annoying if changes to the content weren't seen for 8 hours. Even 10m seems a bit long, but I don't see a compelling reason to override the overall pulp settings.

Various improvements here come to mind:
 * Set a similar value as max-age on the response (or set no-cache). Flatpak defaults to 30m if no max-age is set - so a server admin might be confused that clients don't see changes for many times a configured EXPIRES_TTL.
 * Compute an ETag as hash(contents), so that in the typical case where the index is constant for days, we don't keep sending contents. Doesn't matter much for small numbers of Flatpak distributions in Pulp, which is the use case we are planning for - would matter a lot for large indexes.

I don't think either is necessary to land this PR.

lubosmj: Okay, I propose we leave it the way it was (default pulp): https://github.com/pulp/pulp_container/pull/1301#discussion_r1260332898

stbergmann: > 8h seems far too long to me - since it doesn't look like the cache is ever invalidated.

Just to clarify:  The cache _is_ invalidated, see `invalidate_flatpak_index_cache` in `pulp_container/app/models.py`.  (Hence my earlier comment at <https://github.com/pulp/pulp_container/pull/1301#discussion_r1261166968>:  "And my understanding would be that there doesn't need to be any expiration anyway if the invalidation logic is correct?")

And I'm fine with reverting this back to just use the default again.

stbergmann: > And I'm fine with reverting this back to just use the default again.

done in c384e8a3ddb22258c3cc70760b8520322560e57c

template_config.yml
```diff
@@ -61,9 +61,11 @@ pulp_settings:
   - /tmp
   allowed_import_paths:
   - /tmp
```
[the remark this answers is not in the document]
stbergmann: I don't know what "test flatpack index under pulp_settings" means.

Related to modifications in this file, there are apparently five CI jobs that could either run or skip `pulp_container/tests/functional/api/test_flatpak.py::test_flatpak_install`, depending on whether or not `flatpak_index: true` is enabled for them in some way: "test (pulp)", "test (azure)", "test (s3)", "test (stream)", and "test (lowerbounds)".  The comment at <https://github.com/pulp/pulp_container/blob/c384e8a3ddb22258c3cc70760b8520322560e57c/template_config.yml#L67> argues that "test (s3)" should skip it for now.  And "test (azure)" currently runs the test.

IIRC, the original idea (discussed outside this PR, IIRC) was to enable this for just a subset of CI jobs, to see that both variants (enabled/disabled) work.  But I'm just as fine with enabling it for all jobs but "test (s3)".  So:

(1) For which of the four remaining jobs "test (pulp)", "test (azure)", "test (stream)", and "test (lowerbounds)" shall it be enabled?
(2) If it shall be enabled for "test (pulp)", what modifications would it take to do so?
(3) If it shall be enabled for "teast (lowerbounds)", what modifications would it take to do so?  Add some
```
pulp_settings_lowerbounds:
  flatpak_index: true
```
section to `template_config.yml`?

.github/workflows/update_ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - name: Run update
         working-directory: pulp_container
         run: |
-          .github/workflows/scripts/update_ci.sh
+          ../plugin_template/scripts/update_ci.sh
```
stbergmann, line 56: The changes in this file are the result of running `plugin_template`'s `./plugin-template pulp_container --github`.  Are they good, bad, or benign; should I manually revert them?

lubosmj: I believe these are good. See https://github.com/pulp/plugin_template/commit/fa6c951e5645513b6abae6a2a04f8b0e613ec599.

template_config.yml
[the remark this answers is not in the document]
lubosmj: Leave it enabled for azure and pulp.

```
pulp_settings:
  flatpak_index: true

...

pulp_settings_azure:
   flatpak_index: true
```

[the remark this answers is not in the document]
stbergmann: done in 30c187d5019a5ad33c4c2781ee209985d5940cae

(I had feared that `pulp_settings` might be some kind of base case, and modifying it might also affect all the other `pulp_settings_*`.)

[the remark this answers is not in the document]
stbergmann: > (I had feared that `pulp_settings` might be some kind of base case, and modifying it might also affect all the other `pulp_settings_*`.)

So it looks like my fears were not unfounded, and at <https://github.com/pulp/pulp_container/actions/runs/5623208588/job/15237438136?pr=1301> "test (s3)" now started to execute `pulp_container/tests/functional/api/test_flatpak.py::test_flatpak_install` again (and, of course, failed).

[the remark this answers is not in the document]
stbergmann: So with the latest changes in f08a44fd1f71030874bad60257d4bff4a3c9373f, it runs the test in "test (pulp)", "test (azure)", and "test (lowerbounds)", and skips it in "test (s3)" and "test (stream), and all are green.

lubosmj (approved):
Thanks for working on this! Please, consider this comment once again: https://github.com/pulp/pulp_container/pull/1301#discussion_r1266523554.

ipanova:
@owtaylor do you want to give another look before this gets merged?

stbergmann:
> Please, consider this comment once again: [#1301 (comment)](https://github.com/pulp/pulp_container/pull/1301#discussion_r1266523554).

Which I had addressed with "But anyway, if it's better to be explicit here, done that now in d8ad7a6" at <https://github.com/pulp/pulp_container/pull/1301#discussion_r1269473573>, no?

PR #1330, opened 2023-07-11
Add test code for Flatpak support

stbergmann:
...following up on <https://github.com/pulp/pulp_container/pull/1301> "Implement flatpak index view".

TODO:
* For one, while `registry.fedoraproject.org/kcolorchooser:latest` is rather small (an `org.flatpak.download-size` of 77097 for `app/org.kde.kcolorchooser/x86_64/stable`), its runtime `registry.fedoraproject.org/f38/flatpak-kde5-runtime:latest` is rather large (an `org.flatpak.download-size` of 1200177442 for `runtime/org.fedoraproject.KDE5Platform/x86_64/f38`).  These should be replaced with a minimal runtime and app created just for testing purposes.  (They wouldn't even need to be executable, just need to be installable.)
* And for another, they should be pulled from `ghcr.io` rather than from `registry.fedoraproject.org` (cf. the values in `pulp_container/tests/functional/constants.py` as used by other existing tests).

### DIFF
--- a/.github/workflows/scripts/before_script.sh
+++ b/.github/workflows/scripts/before_script.sh
@@ -39,7 +39,7 @@ cmd_prefix bash -c "usermod -a -G wheel pulp"
 SCENARIOS=("pulp" "performance" "azure" "gcp" "s3" "stream" "generate-bindings" "lowerbounds")
 if [[ " ${SCENARIOS[*]} " =~ " ${TEST} " ]]; then
   # Many functional tests require these
-  cmd_prefix dnf install -yq lsof which
+  cmd_prefix dnf install -yq dbus-daemon flatpak lsof which
 fi
 
 if [[ "${REDIS_DISABLED:-false}" == true ]]; then
--- a/pulp_container/tests/functional/api/test_flatpak.py
+++ b/pulp_container/tests/functional/api/test_flatpak.py
@@ -1,0 +1,39 @@
+"""Tests that verify Flatpak support"""
+import subprocess
+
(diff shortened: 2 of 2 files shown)


owtaylor:
https://github.com/owtaylor/flatpak-oci-test-content/ has scripts to create multi-arch (amd64+aarch64) OCI test images. With both architectures, the sizes are:

```
$ du -hs oci*
3.9M	oci-net.fishsoup.BusyBoxPlatform
36K	oci-net.fishsoup.Hello
```

The time to create these is small (5 seconds, of which 4 seconds is downloading all architectures of docker.io/library/busybox), but I think it's probably better to upload these to the Pulp ghcr.io repoistory rather than create them on the fly in the test code:
 * There's some complexity in the creation  which doesn't need to be in the Pulp test cases (200 lines of Python, but fairly esoteric Flaptak and OCI stuff) 
 * To avoid network traffic to docker.io during the tests, which could be another source of intermittent failures

owtaylor:
Note: this is about small as you can make a *working* Flatpak runtime. We could simply create the runtime with an empty filesystem, and you'd still be able to do `flatpak install`, just `flatpak run` would break. I do think there's some virtue in being able to check that the install actually worked.

stbergmann:
folded into <https://github.com/pulp/pulp_container/pull/1301> "Implement flatpak index view" in <https://github.com/pulp/pulp_container/commit/56626c912d44b6773249cb40e95fdb72c0215f0e>

PR #1332, opened 2023-07-17
Fixed pulp-to-pulp failing sync with ``406 Not Acceptable``

### LINKED ISSUE #1329: 406 error when syncing between pulp instances
**Version**
Upstream pulp
Pulpcore 3.21.2
pulp_container 2.14.2

Downstream pulp from pulp-in-one container (version 3.24.1)
Pulpcore 3.24.1
pulp_container 2.14.6

**Describe the bug**
Pulling container content from the upstream pulp into the downstream pulp fails with 406 `Not Acceptable`.

**Additional context**
This error has been narrowed down to a change in the latest version of pulp_container.
It can be resolved by reverting the recent change to [line 49](https://github.com/pulp/pulp_container/compare/2.14.5...2.14.6#diff-ffafbe9964d2494a4b00c254f07c75e6a3b44a892775375b120e550cd648e638L49) of `pulp_container/app/downloaders.py`.
The change was itself brought in as a temporary fix for issue https://github.com/pulp/pulp_container/issues/1288

ipanova:
closes #1329

### DIFF
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -343,7 +343,9 @@ class ContainerFirstStage(Stage):
         while True:
             link = urljoin(self.remote.url, rel_link)
             list_downloader = self.remote.get_downloader(url=link)
-            await list_downloader.run(extra_data={"repo_name": repo_name})
+            # FIXME this can be rolledback after https://github.com/pulp/pulp_container/issues/1288
+            # tags/list endpoint does not like any unnecessary headers to be sent
+            await list_downloader.run(extra_data={"repo_name": repo_name, "headers": {}})
             with open(list_downloader.path) as tags_raw:
                 tags_dict = json.loads(tags_raw.read())
                 tag_list.extend(tags_dict["tags"])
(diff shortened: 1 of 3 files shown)


pulp_container/app/tasks/sync_stages.py
```diff
@@ -343,7 +343,9 @@ async def get_paginated_tag_list(self, rel_link, repo_name):
         while True:
             link = urljoin(self.remote.url, rel_link)
             list_downloader = self.remote.get_downloader(url=link)
-            await list_downloader.run(extra_data={"repo_name": repo_name})
+            # FIXME this can be rolledback after https://github.com/pulp/pulp_container/issues/1288
+            # tags/list endpoint does not like any unnecessary headers to be sent
+            await list_downloader.run(extra_data={"repo_name": repo_name, "headers": {}})
```
ipanova, line 348: FYI the proper solution would be to write a CustomRenderer, https://www.django-rest-framework.org/api-guide/renderers/#custom-renderers  like we have for Blobs and Manifests https://github.com/pulp/pulp_container/blob/main/pulp_container/app/registry_api.py#L89 and add it to every view/viewset

PR #1340, opened 2023-07-20
Check a list of scopes when verifying tokens

### LINKED ISSUE #1286: Users can mount blobs without having read access to the `from` repository
We verify just the access to the created `to` repository (https://github.com/pulp/pulp_container/blob/15aca2fa56e53a0b0c72cc299e468dcf0ad81545/pulp_container/app/token_verification.py#L205) and still allow performing blob mount without verifying read permissions to the `from` repository (https://github.com/pulp/pulp_container/blob/15aca2fa56e53a0b0c72cc299e468dcf0ad81545/pulp_container/app/registry_api.py#L578).

The fix will need to be backported.

lubosmj:
Before this commit, the registry checked one scope, ignoring other scopes that could relate to blob mounting operations. Due to that, users without sufficient permissions could mount blobs from other users unauthorized.

closes #1286

### DIFF
--- a/pulp_container/app/token_verification.py
+++ b/pulp_container/app/token_verification.py
@@ -39,16 +39,24 @@ def _decode_token(encoded_token, request):
     return decoded_token
 
 
-def _contains_accessible_actions(decoded_token, scope):
+def _contains_accessible_actions(decoded_token, scopes):
     """
-    Check if a client has permission to perform operations within the current scope
-    """
-    for access in decoded_token["access"]:
-        if scope.resource_type == access["type"] and scope.name == access["name"]:
-            if scope.action in access["actions"]:
-                return True
+    Check if a client has permission to perform operations within the current scope.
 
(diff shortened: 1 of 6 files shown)


pulp_container/app/token_verification.py
```diff
@@ -39,16 +39,24 @@ def _decode_token(encoded_token, request):
...
-    for access in decoded_token["access"]:
-        if scope.resource_type == access["type"] and scope.name == access["name"]:
-            if scope.action in access["actions"]:
-                return True
+    accessible_actions = []
+    for scope in scopes:
+        for access in decoded_token["access"]:
+            if scope.resource_type == access["type"] and scope.name == access["name"]:
+                if scope.action in access["actions"]:
+                    accessible_actions.append(True)
```
ipanova, line 54: i might have used here a boolean instead, but list will work ok too.

PR #1359, opened 2023-07-27
Update Changelog

pulpbot:


lubosmj (changes requested):
Why is this happening when I fixed the beta versioning in the main branch? Is this change created because of the other branches where the versioning has not been fixed/merged yet, @mdellweg?

ipanova:
i merged the manual changelog update, let's close this one and see if with the next release we get any issues.

PR #1362, opened 2023-08-01
Update Changelog

pulpbot:


mdellweg:
If you keep it open it will not create a new one every day.

ipanova:
you should try 'sudo, please stop it' ;)

PR #1364, opened 2023-08-02
Fix flatpak settings for the S3 test runner 

lubosmj:
The flatpak_index setting has to be explicitly set to false because of
the global pulp settings definition.

Enabling the indexing for the S3 test runner would cause
pulp_container/tests/functional/api/test_flatpak.py to fail due to
<https://gitlab.gnome.org/GNOME/libsoup/-/issues/357>.

[noissue]


template_config.yml
```diff
@@ -66,11 +66,8 @@ pulp_settings:
 pulp_settings_azure:
   flatpak_index: true
 pulp_settings_gcp: null
-# Do not enable flatpak_index for s3 for now, as it would cause
-# pulp_container/tests/functional/api/test_flatpak.py to fail due to
-# <https://gitlab.gnome.org/GNOME/libsoup/-/issues/357> "Odd 400 response with https GET 302
-# redirecting to http":
```
lubosmj, line 72: This was removed by plugin-template. Pulpbot will always try to remove this note too in a new "CI update" PR.

ggainey: Maybe put this note into the commit-msg for this PR? That'll preserve it (via git-blame/git-show) for someone 6 months from now wondering why it's set to false.

ipanova:
Can you change the commit to `| Fix flatpak settings for s3 test runner`

ipanova (approved):

PR #1376, opened 2023-08-25
Refactor the repository retrieval in get_dr_push

### LINKED ISSUE #489: Refactor get_dr_push so it is more readable

Redmine Issue: 8577, https://pulp.plan.io/issues/8577

---

https://github.com/pulp/pulp_container/blob/master/pulp_container/app/registry_api.py#L311

### LINKED ISSUE #1275: Investigate intermittent test failures in `PushManifestListTestCase`
```
=================================== FAILURES ===================================
____________ PushManifestListTestCase.test_push_manifest_list_v2s2 _____________

self = <pulp_container.tests.functional.api.test_push_content.PushManifestListTestCase testMethod=test_push_manifest_list_v2s2>

    def test_push_manifest_list_v2s2(self):
        """Push the created manifest list in the v2s2 format."""
        self.registry.login(
            "-u", self.user_admin["username"], "-p", self.user_admin["password"], self.registry_name
        )
>       self.registry.manifest_push(
            self.v2s2_tag,
            self.v2s2_image_path,
            "--all",
            "--format",
            "v2s2",
        )

usr/local/lib/python3.8/site-packages/pulp_container/tests/functional/api/test_push_content.py:432: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
usr/local/lib/python3.8/site-packages/pulp_smash/cli.py:975: in <lambda>
    self.manifest_push = lambda *args: self._dispatch_command(
usr/local/lib/python3.8/site-packages/pulp_smash/cli.py:1036: in _dispatch_command
    result = self._client.run(cmd)
usr/local/lib/python3.8/site-packages/pulp_smash/cli.py:322: in run
usr/local/lib/python3.8/site-packages/pulp_smash/cli.py:54: in code_handler
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = CompletedProcess(args=('podman', 'manifest', 'push', 'manifest_list', 'pulp.example.com:443/foo:manifest_list', '--all...: initiating layer upload to /v2/foo/blobs/uploads/ in pulp.example.com:443: name invalid: Invalid repository name.\n')

    def check_returncode(self):
        """Raise an exception if ``returncode`` is non-zero.
    
        Raise :class:`pulp_smash.exceptions.CalledProcessError` if
        ``returncode`` is non-zero.
    
        Why not raise ``subprocess.CalledProcessError``? Because stdout and
        stderr are not included when str() is called on a CalledProcessError
        object. A typical message is::
    
            "Command '('ls', 'foo')' returned non-zero exit status 2"
    
        This information is valuable. One could still make
        ``subprocess.CalledProcessError`` work by overloading ``args``:
    
        >>> if isinstance(args, (str, bytes)):
        ...     custom_args = (args, stdout, stderr)
        ... else:
        ...     custom_args = tuple(args) + (stdout, stderr)
        >>> subprocess.CalledProcessError(args, returncode)
    
        But this seems like a hack.
    
        In addition, it's generally good for an application to raise expected
        exceptions from its own namespace, so as to better abstract away
        dependencies.
        """
        if self.returncode != 0:
>           raise exceptions.CalledProcessError(
                self.args, self.returncode, self.stdout, self.stderr
            )
E           pulp_smash.exceptions.CalledProcessError: Command ('podman', 'manifest', 'push', 'manifest_list', 'pulp.example.com:443/foo:manifest_list', '--all', '--format', 'v2s2', '--tls-verify=true') returned non-zero exit status 125.
E           
E           stdout: 
E           
E           stderr: Getting image list signatures
E           Copying 3 of 3 images in list
E           Copying image sha256:d8fbbbf3fec1857c32c110292a9decf9744f9f97d7247019ae4776c241395221 (1/3)
E           Getting image source signatures
E           Copying blob sha256:463b9bfafbaf88a2dee0aa77fd293a7cb91558af9664bc23d4ed00830463615d
E           Copying blob sha256:d7ab49ce0b755a907e914d148cc8c1f437f7688048aea18a15d965f60c5d0a43
E           Error: copying image 1/3 from manifest list: writing blob: initiating layer upload to /v2/foo/blobs/uploads/ in pulp.example.com:443: name invalid: Invalid repository name.
```

lubosmj:
closes #1275
closes #489

### DIFF
--- a/CHANGES/1275.misc
+++ b/CHANGES/1275.misc
@@ -1,0 +1,2 @@
+Ensured repositories are correctly retrieved or initialized when handling client's parallel blob
+uploading.
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -25,14 +25,13 @@ from django.conf import settings
 from pulpcore.plugin.models import Artifact, ContentArtifact, UploadChunk
 from pulpcore.plugin.files import PulpTemporaryUploadedFile
 from pulpcore.plugin.tasking import add_and_remove, dispatch
-from pulpcore.plugin.util import get_objects_for_user
+from pulpcore.plugin.util import get_objects_for_user, get_url
 from rest_framework.exceptions import (
     AuthenticationFailed,
     NotAuthenticated,
     PermissionDenied,
     ParseError,
     Throttled,
-    ValidationError,
 )
 from rest_framework.generics import ListAPIView
 from rest_framework.pagination import BasePagination
@@ -291,32 +290,7 @@ class ContainerRegistryApiMixin:
             distribution = models.ContainerDistribution.objects.get(base_path=path)
         except models.ContainerDistribution.DoesNotExist:
             if create:
-                try:
-                    with transaction.atomic():
-                        repo_serializer = serializers.ContainerPushRepositorySerializer(
-                            data={"name": path}, context={"request": request}
-                        )
-                        repo_serializer.is_valid(raise_exception=True)
-                        repository = repo_serializer.create(repo_serializer.validated_data)
-                        repo_href = serializers.ContainerPushRepositorySerializer(
-                            repository, context={"request": request}
-                        ).data["pulp_href"]
-
-                        dist_serializer = serializers.ContainerDistributionSerializer(
-                            data={"base_path": path, "name": path, "repository": repo_href}
-                        )
-                        dist_serializer.is_valid(raise_exception=True)
-                        distribution = dist_serializer.create(dist_serializer.validated_data)
-                except ValidationError:
-                    raise RepositoryInvalid(name=path)
-                except IntegrityError:
-                    # Seems like another process created our stuff already. Retry fetching it.
-                    distribution = models.ContainerDistribution.objects.get(base_path=path)
-                    repository = distribution.repository
-                    if repository:
-                        repository = repository.cast()
-                        if not repository.PUSH_ENABLED:
-                            raise RepositoryInvalid(name=path, message="Repository is read-only.")
+                distribution, repository = self.create_dr(path, request)
             else:
                 raise RepositoryNotFound(name=path)
         else:
@@ -326,27 +300,31 @@ class ContainerRegistryApiMixin:
                 if not repository.PUSH_ENABLED:
                     raise RepositoryInvalid(name=path, message="Repository is read-only.")
             elif create:
-                try:
-                    with transaction.atomic():
-                        repo_serializer = serializers.ContainerPushRepositorySerializer(
-                            data={"name": path}, context={"request": request}
-                        )
-                        repo_serializer.is_valid(raise_exception=True)
-                        repository = repo_serializer.create(repo_serializer.validated_data)
-                        distribution.repository = repository
-                        distribution.save()
-                except IntegrityError:
-                    # Seems like another process created our stuff already. Retry fetching it.
-                    distribution = models.ContainerDistribution.objects.get(pk=distribution.pk)
-                    repository = distribution.repository
-                    if repository:
-                        repository = repository.cast()
-                        if not repository.PUSH_ENABLED:
-                            raise RepositoryInvalid(name=path, message="Repository is read-only.")
-                    else:
-                        raise RepositoryNotFound(name=path)
+                with transaction.atomic():
+                    repository = serializers.ContainerPushRepositorySerializer.get_or_create(
+                        {"name": path}
+                    )
+                    distribution.repository = repository
+                    distribution.save()
             else:
                 raise RepositoryNotFound(name=path)
+        return distribution, repository
+
+    def create_dr(self, path, request):
+        with transaction.atomic():
+            repository = serializers.ContainerPushRepositorySerializer.get_or_create({"name": path})
+            distribution = serializers.ContainerDistributionSerializer.get_or_create(
+                {"base_path": path, "name": path}, {"repository": get_url(repository)}
+            )
+
+            if distribution.repository:
+                dist_repository = distribution.repository.cast()
+                if not dist_repository.PUSH_ENABLED or repository != dist_repository:
+                    raise RepositoryInvalid(name=path, message="Repository is read-only.")
+            else:
+                distribution.repository = repository
+                distribution.save()
+
         return distribution, repository
 
 
--- a/CHANGES/489.misc
+++ b/CHANGES/489.misc
@@ -1,0 +1,1 @@
+Refactored ``get_dr_push()`` to be more readable.
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -196,7 +196,7 @@ class ContainerRepositorySerializer(RepositorySerializer):
         model = models.ContainerRepository
 
 
-class ContainerPushRepositorySerializer(RepositorySerializer):
+class ContainerPushRepositorySerializer(RepositorySerializer, GetOrCreateSerializerMixin):
     """
     Serializer for Container Push Repositories.
     """
@@ -277,7 +277,7 @@ class ContainerRemoteSerializer(RemoteSerializer):
         model = models.ContainerRemote
 
 
-class ContainerDistributionSerializer(DistributionSerializer):
+class ContainerDistributionSerializer(DistributionSerializer, GetOrCreateSerializerMixin):
     """
     A serializer for ContainerDistribution.
     """
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 ecdsa>=0.14,<=0.18.0
 jsonschema>=4.4,<4.20
-pulpcore>=3.40.1,<3.55
+pulpcore>=3.40.3,<3.55
 pyjwkest>=1.4,<=1.4.2
 pyjwt[crypto]>=2.4,<2.9


CHANGES/1275.misc
```diff
@@ -0,0 +1,2 @@
+Ensured repositories are correctly retrieved or initialized when handling concurrent requests
```
ipanova, line 1: when handling client parallel blob upload

pulp_container/app/registry_api.py
```diff
@@ -349,6 +325,32 @@ def get_dr_push(self, request, path, create=False):
...
+            raise RepositoryInvalid(name=path)
+
+        dist_serializer = serializers.ContainerDistributionSerializer(
+            data={"base_path": path, "name": path, "repository": get_url(repository)},
+            context={"request": request},
+        )
+        try:
+            dist_serializer.is_valid(raise_exception=True)
+            distribution = dist_serializer.create(dist_serializer.validated_data)
+        except (IntegrityError, ValidationError):
```
ipanova, line 343: you should catch these error separately because the message is different

pulp_container/app/registry_api.py
```diff
@@ -349,6 +325,32 @@ def get_dr_push(self, request, path, create=False):
                 raise RepositoryNotFound(name=path)
         return distribution, repository
 
+    @transaction.atomic
```
ipanova, line 328: there is a different wrapping into a transaction whole method vs just the try block

lubosmj: I removed this change. The logic should be similar to the previous one, where we catch exceptions outside the `transaction.atomic()`'s context manager.

ipanova (dismissed):

lubosmj:
@ipanova, I believe this is the best I can do. We no longer granularly handle the validation and creation of entities inside pulp-container, see https://github.com/pulp/pulpcore/blob/3594624a3556af2ad400c29871041252ae947495/pulpcore/app/serializers/base.py#L360.

pulp_container/app/registry_api.py
```diff
@@ -349,6 +324,24 @@ def get_dr_push(self, request, path, create=False):
...
+    def create_dr(self, path, request):
+        try:
+            with transaction.atomic():
+                repository = serializers.ContainerPushRepositorySerializer.get_or_create(
+                    {"name": path}
+                )
+                distribution = serializers.ContainerDistributionSerializer.get_or_create(
+                    {"base_path": path, "name": path}, {"repository": get_url(repository)}
+                )
+        except ObjectDoesNotExist:
```
ipanova, line 336: this seems to be unnecessary, we whether retrieve existing instance or create a new one

pulp_container/app/registry_api.py
```diff
@@ -349,6 +324,24 @@ def get_dr_push(self, request, path, create=False):
...
         return distribution, repository
 
+    def create_dr(self, path, request):
+        try:
+            with transaction.atomic():
+                repository = serializers.ContainerPushRepositorySerializer.get_or_create(
+                    {"name": path}
+                )
+                distribution = serializers.ContainerDistributionSerializer.get_or_create(
+                    {"base_path": path, "name": path}, {"repository": get_url(repository)}
```
ipanova, line 334: great, you figured it :+1:

ipanova: here you need to check whether returned distribution has a repo. and if it has it should be same repo-is as on line 330 otherwise we need to raise an error that repo is read-only
how about this workflow:
1. create repo(regular) foo and distribution foo
2. docker push foo

pulp_container/app/registry_api.py
```diff
@@ -349,6 +324,24 @@ def get_dr_push(self, request, path, create=False):
...
+                )
+        except ObjectDoesNotExist:
+            raise RepositoryInvalid(name=path)
+
+        repository = repository.cast()
+        if not repository.PUSH_ENABLED:
```
ipanova, line 340: yeah so this needs to be moved upper per my previous comment. now you're checking the repo you have created on line 330

pulp_container/app/registry_api.py
```diff
@@ -291,32 +291,7 @@ def get_dr_push(self, request, path, create=False):
...
-                    distribution = models.ContainerDistribution.objects.get(base_path=path)
-                    repository = distribution.repository
-                    if repository:
-                        repository = repository.cast()
-                        if not repository.PUSH_ENABLED:
-                            raise RepositoryInvalid(name=path, message="Repository is read-only.")
+                distribution, repository = self.create_dr(path, request)
             else:
                 raise RepositoryNotFound(name=path)
         else:
```
ipanova, line 297: Have you checked the else branch? I think it needs re-working too.

ipanova: line 328 makes little sense without trying to understand much the code, can you double check the logic if this is still needed?

pulp_container/app/registry_api.py
lubosmj: I think this is necessary. Once the validation fails (e.g., because of an invalid base_path), it tries to fetch an object from the database: https://github.com/pulp/pulpcore/blob/3594624a3556af2ad400c29871041252ae947495/pulpcore/app/serializers/base.py#L374. If the object does not exist after the last attempt, we should return `RepositoryNotFound`.

ipanova: I see what you're saying and you are right only because the get_or_create is not implemented fully as it should be. Get or create implies that on success an object will always be returned.
Here https://github.com/pulp/pulpcore/blob/3594624a3556af2ad400c29871041252ae947495/pulpcore/app/serializers/base.py#L372 we should not catch these together but separately, because if validation fails because of field being missing or invalid we need to raise it. And if it failed because 'this field must be unique; then we retrieve existing object.
Here I took a stab https://github.com/pulp/pulpcore/pull/4643

pulp_container/app/registry_api.py
ipanova: I revised the else branch and it also needs to switch the serializers to use get_or_create.

lubosmj:
Waiting for https://github.com/pulp/pulpcore/commit/bf0654bf144acb0a22c45057bfde7cdf2bd3b042 to be released.

pulp_container/app/registry_api.py
```diff
@@ -328,11 +302,9 @@ def get_dr_push(self, request, path, create=False):
...
-                        repo_serializer = serializers.ContainerPushRepositorySerializer(
-                            data={"name": path}, context={"request": request}
+                        repository = serializers.ContainerPushRepositorySerializer.get_or_create(
+                            {"name": path}
                         )
-                        repo_serializer.is_valid(raise_exception=True)
-                        repository = repo_serializer.create(repo_serializer.validated_data)
                         distribution.repository = repository
                         distribution.save()
                 except IntegrityError:
```
ipanova, line 310: this can be removed

pulp_container/app/registry_api.py
```diff
@@ -349,6 +321,19 @@ def get_dr_push(self, request, path, create=False):
...
         return distribution, repository
 
+    def create_dr(self, path, request):
+        with transaction.atomic():
+            repository = serializers.ContainerPushRepositorySerializer.get_or_create({"name": path})
+            distribution = serializers.ContainerDistributionSerializer.get_or_create(
+                {"base_path": path, "name": path}, {"repository": get_url(repository)}
+            )
+
+            dist_repository = distribution.repository.cast()
```
ipanova, line 331: you need to account for the case when distribution.repository will be None

lubosmj: Oh, yes.

pulp_container/app/registry_api.py
lubosmj: I was unsure if `save()` could raise any error. By the way, should we enforce saving the object just in time with `force_update=True`?

ipanova: i don't think so https://docs.djangoproject.com/en/dev/ref/models/instances/#ref-models-force-insert

pulp_container/app/registry_api.py
```diff
@@ -326,29 +300,39 @@ def get_dr_push(self, request, path, create=False):
...
         return distribution, repository
 
+    def create_dr(self, path, request):
+        with transaction.atomic():
+            repository = serializers.ContainerPushRepositorySerializer.get_or_create({"name": path})
+            distribution = serializers.ContainerDistributionSerializer.get_or_create(
+                {"base_path": path, "name": path}, {"repository": get_url(repository)}
+            )
+
+            if distribution.repository:
```
ipanova, line 329: it needs and else:
distribution.repository = repository
 distribution.save()

pulp_container/app/registry_api.py
```diff
@@ -326,29 +300,39 @@ def get_dr_push(self, request, path, create=False):
...
-                            raise RepositoryInvalid(name=path, message="Repository is read-only.")
-                    else:
-                        raise RepositoryNotFound(name=path)
+                with transaction.atomic():
+                    repository = serializers.ContainerPushRepositorySerializer.get_or_create(
+                        {"name": path}
+                    )
+                    distribution.repository = repository
+                    distribution.save()
+                # Seems like another process created our stuff already. Retry fetching it.
```
ipanova, line 309: line 309-315 can be entirely removed

pulp_container/app/registry_api.py
lubosmj: It is not needed if we create a new distribution. We want to check the names and references without updating anything.
```
In [1]: from pulp_container.app.serializers import ContainerPushRepositorySerializer

In [2]: from pulp_container.app.serializers import ContainerDistributionSerializer

In [3]: repo = ContainerPushRepositorySerializer.get_or_create({"name": "bla"})

In [4]: from pulp_container.app.models import ContainerPushRepository, ContainerDistribution

In [5]: ContainerPushRepository.objects.filter(name="bla")
Out[5]: <QuerySet [<ContainerPushRepository (pulp_type=container.container-push): pk=018bb01f-a735-79ab-a913-8878aec9226c>]>

In [6]: ContainerDistribution.objects.filter(name="bla")
Out[6]: <QuerySet []>

In [7]: from pulpcore.plugin.util import get_url

In [8]: dist = ContainerDistributionSerializer.get_or_create({"name": "bla", "base_path": "bla"}, {"repository": get_url(repo)})

In [9]: dist
Out[9]: <ContainerDistribution (pulp_type=container.container): pk=018bb022-33d0-723e-8562-576ce14e9ed8>

In [10]: dist.repository
Out[10]: <ContainerPushRepository (pulp_type=container.container-push): pk=018bb01f-a735-79ab-a913-8878aec9226c>

In [11]: repo
Out[11]: <ContainerPushRepository (pulp_type=container.container-push): pk=018bb01f-a735-79ab-a913-8878aec9226c>
```

lubosmj: The code path suggests that we are about to create a new repository:
```
            distribution = models.ContainerDistribution.objects.get(base_path=path)
        except models.ContainerDistribution.DoesNotExist:
            if create:
                distribution, repository = self.create_dr(path, request)
```
Do you agree?

ipanova: i agree that we are about to create new distribution, however to account for race condition during get_or_create ( and that's why we're using it here) another process could have created a distribution(without a repo) and the get_or_create would get the newly-created-dist by another process

lubosmj: Yes, you are right!

lubosmj:
I am closing https://github.com/pulp/pulp_container/issues/489 with this change.

ipanova (approved):
:clinking_glasses:

PR #1383, opened 2023-09-04
Fix the docs search bar

lubosmj:
[noissue]

### DIFF
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -1,4 +1,6 @@
 plantuml
-sphinx
-sphinx-rtd-theme
+sphinx<7.0
+sphinx-rtd-theme==1.2.0
+sphinxcontrib-jquery
+sphinxcontrib-openapi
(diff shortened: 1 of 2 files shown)


lubosmj:
Inspired by pulp_rpm (https://github.com/pulp/pulp_rpm/commit/9469ce1b4e0475d253df85456ccd6d7e9d9105fc) where the search bar works.

lubosmj:
This should finally resolve the issues with the search bar. I have tested it locally and it works.

mdellweg:
Have you tried using `ci_update_docs: true` and `doc_requirements_from_pulpcore: false` in the plugin template?

lubosmj:
I have not. What you have suggested is used in pulp_file and the search bar does not work.

https://github.com/pulp/pulp_file/blob/main/template_config.yml

mdellweg (approved):

PR #1401, opened 2023-10-12
Fix the file extension of exports

lubosmj:
[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_pulpimportexport.py
+++ b/pulp_container/tests/functional/api/test_pulpimportexport.py
@@ -32,6 +32,7 @@ def test_import_export_standard(
     importers_pulp_api_client,
     importers_pulp_imports_api_client,
     gen_object_with_cleanup,
+    has_pulp_plugin,
 ):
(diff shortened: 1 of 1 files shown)


lubosmj:
This is trying to fix CI errors caused by https://github.com/pulp/pulpcore/commit/c49624c0583b6f78d9067e7e8ab7713fe81d7ff5.

mdellweg (approved):
I like the sharp version dependent assertion.

But also i'm wondering, is it worth a test in pulp_container how the files are named? Is that maybe a pulpcore implementation detail we should not care about here?
Anyway, I'm approving.

ipanova (approved):

lubosmj:
To my knowledge, there is a limitation where we do not know the name of the exported tarball beforehand. That is the reason why this change was introduced in the first place. The same is done in pulp_rpm: https://github.com/pulp/pulp_rpm/blob/a1a3bdd051bdb263445a72bc4b652bec6db910fa/pulp_rpm/tests/functional/api/test_pulpimport.py#L269.

PR #1424, opened 2023-11-17
Use new fixture signing key

dralley:
[noissue]

### DIFF
--- a/pulp_container/tests/functional/api/test_push_signatures.py
+++ b/pulp_container/tests/functional/api/test_push_signatures.py
@@ -73,9 +73,7 @@ def test_assert_signed_image(
         decrypted = gpg.decrypt(raw_s)
 
         assert decrypted.key_id == keyid
-        assert decrypted.fingerprint == fingerprint
         assert decrypted.status == "signature valid"
-        assert decrypted.username == "Pulp QE"
 
         json_s = json.loads(decrypted.data)
 


pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -75,7 +75,7 @@ def test_assert_signed_image(
         assert decrypted.key_id == keyid
         assert decrypted.fingerprint == fingerprint
         assert decrypted.status == "signature valid"
-        assert decrypted.username == "Pulp QE"
+        assert decrypted.username == "pulp-fixture-signing-key"
```
mdellweg, line 78: Oh, this will introduce a bad version interdependence issue. If we cannot defer the userid of the key, we should not assert on it.

lubosmj: This test is useful for asserting data stored in Pulp. We verify whether we did not change anything along the way while parsing images. Am I reading it correctly that you want to remove this check?

mdellweg: Depending on the version of pulpcore used, this username will be the new one or the old one. So the best way would be to find a way to get the asserted name from the gpgkey fixture in pulpcore. If that is impossible, i'd really just remove the assertion (or assert only that it's not empty).

dralley: The fixture doesn't provide that information.  It does provide the KeyID, which IMO is good enough and effectively tests the same thing.  Actually the "fingerprint" is just a subset of the "key ID" too, yeah?

lubosmj: Sure, let's remove it!

dralley:
@lubosmj @mdellweg updated

pulp_container/tests/functional/api/test_push_signatures.py
```diff
@@ -73,9 +73,7 @@ def test_assert_signed_image(
         decrypted = gpg.decrypt(raw_s)
 
         assert decrypted.key_id == keyid
-        assert decrypted.fingerprint == fingerprint
```
mdellweg, line 76: The fingerprint comes from the bindings. I believe that should be fine. Maybe I'm missing something again.

dralley: See previous comment, fingerprint and keyid are effectively the same, as one (I believe keyid) is just the last X bits of the other.  Realistically just checking one is the same as checking both as they are both meant to be functionally unique values.

lubosmj (approved):
This should also fix nightly checks:
```
>           assert decrypted.username == "Pulp QE"
E           AssertionError: assert 'pulp-fixture...pproject.org>' == 'Pulp QE'
E             - Pulp QE
E             + pulp-fixture-signing-key <generated-signing-key@pulpproject.org>
```

https://github.com/pulp/pulp_container/actions/runs/6952312275/job/18948179340

PR #1508, opened 2024-02-13
Remove the optional "kid" parameter

### LINKED ISSUE #1485: Consider replacing the python-ecdsa dependency
## Minerva timing attack on P-256 in python-ecdsa (dependabot)

python-ecdsa has been found to be subject to a Minerva timing attack on the P-256 curve. Using the ecdsa.SigningKey.sign_digest() API function and timing signatures an attacker can leak the internal nonce which may allow for private key discovery. Both ECDSA signatures, key generation, and ECDH operations are affected. ECDSA signature verification is unaffected. The python-ecdsa project considers side channel attacks out of scope for the project and there is no planned fix.

Affected versions: <= 0.18.0

lubosmj:
This removes the key identifier stored in the JWK's header in the manifest v2 schema 1 signatures' payload.

The "kid" parameter is optional in JWK. This parameter might be also considered irrelevant because the key pair for ECDSA is generated each time the conversion from schema 2 to schema 1 happens. So, clients cannot verify the origin of the signature with the fingerprint/kid because the public key is created on the fly and then immediately trashed.

Ref: https://www.rfc-editor.org/rfc/rfc7517#section-4.5
Ref: https://docker-docs.uclv.cu/registry/spec/manifest-v2-1/

closes #1485

### DIFF
--- a/CHANGES/1485.removal
+++ b/CHANGES/1485.removal
@@ -1,0 +1,3 @@
+Removed the optional "kid" parameter stored inside the signatures' payload generated during
+docker manifest v2 schema 1 conversion. This change also removes the ``ecdsa`` dependency,
+which is vulnerable to Minevra timing attacks.
--- a/pulp_container/app/schema_convert.py
+++ b/pulp_container/app/schema_convert.py
@@ -1,9 +1,5 @@
-import base64
-import binascii
 import datetime
-import ecdsa
 import hashlib
-import itertools
 import json
 import logging
 
@@ -129,7 +125,6 @@ class Schema2toSchema1Converter:
         )
 
         key = jwk.ECKey().load_key(ecc.P256)
-        key.kid = getKeyId(key)
         manifest_data = _jsonDumps(manifest)
         signed_manifest_data = sign(manifest_data, key)
         return manifest_data, signed_manifest_data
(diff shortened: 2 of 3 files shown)


lubosmj:
I am wondering if it makes sense to deprecate the whole conversion machinery, see https://access.redhat.com/solutions/4346011.

CHANGES/1485.removal
```diff
@@ -0,0 +1,2 @@
+Removed the optional "kid" parameter stored inside the signatures' payload generated during
```
ipanova, line 1: it's going to look a bit weird because other incoming schema1 manifests (not converted by pulp)  would contain it

CHANGES/1485.removal
```diff
@@ -0,0 +1,2 @@
...
+manifest v2 schema 1 conversion.
```
ipanova, line 2: generated during docker manifest v2 schema 1 conversion

ipanova (approved):

CHANGES/1485.removal
lubosmj: And that is perfectly fine because users can verify the origin of the fingerprint. When it comes to the on-the-fly schema conversion, there is no way the users benefit from it, is it?

CHANGES/1485.removal
lubosmj: Okay. I am mentioning also the rationale behind the ecdsa removal.

CHANGES/1485.removal
ipanova: I meant that there might be incoming schema1 directly from remote registry without us doing any conversion. And that schema will contain kid. It was not a concern or anything, just an observation. I don't we should worry about this much because we've just agreed to deprecate the whole conversion mechanism

ipanova (approved):

Issue #1509, opened 2024-02-14
Remove the manifest schema conversion machinery


lubosmj:
This feature is no longer expected to be supported or maintained.

https://github.com/pulp/pulp_container/blob/4dfeb5cb71fc09aff23de3443b367e8bed5b0f8e/pulp_container/app/schema_convert.py#L23

https://github.com/pulp/pulp_container/blob/4dfeb5cb71fc09aff23de3443b367e8bed5b0f8e/docs/workflows/host.rst?plain=1#L116

lubosmj:
Additionally, we will remove the `pyjwkest` dependency from our requirements.

lubosmj:
In the end, we would like to return 404 in case the client is requesting docker manifest v2 schema 1 instead of accepting what is available (manifest v2 schema 2).

https://github.com/pulp/pulp_container/blob/59e06e591bd3e621401d83f417fd3fa2ecadbf0a/pulp_container/app/registry.py#L216

lubosmj:
Based on docker documentation (https://docker-docs.uclv.cu/registry/spec/manifest-v2-2/#backward-compatibility), we will no longer support the following procedure:
> If the manifest being requested uses the new format, and the appropriate media type is not present in an Accept header, the registry will assume that the client cannot handle the manifest as-is, and rewrite it on the fly into the old format.

Which is totally acceptable.

Issue #1493, opened 2024-01-31
Pull-through caching may fail on a non-localfilesystem storage backend


lubosmj:
Steps to reproduce:
1. Install Pulp with the S3 storage
2. Run `podman pull` to trigger the pull-through caching
3. Repeat 1-2 steps until you get the error

CI logs (pulp/S3):
```
Trying to pull [pulp.example.com:443/db63fe02-27e1-422e-95b4-6a6be3568c60/pulp/test-fixture-1:manifest_a.](http://pulp.example.com:443/db63fe02-27e1-422e-95b4-6a6be3568c60/pulp/test-fixture-1:manifest_a.)..
Getting image source signatures
Copying blob sha256:463b9bfafbaf88a2dee0aa77fd293a7cb91558af9664bc23d4ed00830463615d
Copying blob sha256:d7ab49ce0b755a907e914d148cc8c1f437f7688048aea18a15d965f60c5d0a43
time="2024-01-30T10:38:51Z" level=warning msg="Failed, retrying in 1s ... (1/3). Error: reading blob sha256:d7ab49ce0b755a907e914d148cc8c1f437f7688048aea18a15d965f60c5d0a43: fetching blob: blob unknown: Blob not found."
Getting image source signatures
Copying blob sha256:463b9bfafbaf88a2dee0aa77fd293a7cb91558af9664bc23d4ed00830463615d
Copying blob sha256:d7ab49ce0b755a907e914d148cc8c1f437f7688048aea18a15d965f60c5d0a43
Error: reading blob sha256:463b9bfafbaf88a2dee0aa77fd293a7cb91558af9664bc23d4ed00830463615d: fetching blob: received unexpected HTTP status: 500 Internal Server Error
```

Traceback:
```
Error: 1-30 10:38:53 +0000] [503] [ERROR] Error handling request
Traceback (most recent call last):
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
    resp = await request_handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_app.py", line 543, in _handle
    resp = await handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_middlewares.py", line 114, in impl
    return await handler(request)
  File "/usr/local/lib/python3.8/site-packages/pulpcore/content/authentication.py", line 48, in authenticate
    return await handler(request)
[...]

PR #1529, opened 2024-02-26
Remove the manifest schema conversion machinery

git-hyagi:
closes: #1509


unittest_requirements.txt
```diff
@@ -2,3 +2,4 @@ mock
 pytest-django
 pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pytest<8
+pyjwkest>=1.4,<=1.4.2
```
lubosmj, line 5: Do we still need this dependency?

pulp_container/tests/functional/api/test_content_cache.py
```diff
@@ -186,6 +192,11 @@ def fetch_response_metadata(self, response):
         if DEFAULT_FILE_STORAGE in STANDARD_FILE_STORAGE_FRAMEWORKS:
             return response.status_code, response.headers.get("X-PULP-CACHE")
         else:
+            """Corner case for deprecated v2 schema1"""
```
lubosmj, line 195: Which corner cases are we catching here? I thought that we were dealing only with manifest v2 schema 2 content inside this case.

pulp_container/app/registry.py
```diff
@@ -231,6 +229,11 @@ async def dispatch_tag(self, request, tag, response_headers):
                 streamed back to the client.
 
         """
+        # return 404 in case the client is requesting docker manifest v2 schema 1
```
lubosmj, line 232: I think this part is not required inside `dispatch_tag`. How about we remove the code and add the highlighted comment just above the `raise PathNotResolved` call placed at the end of the `get_tag` method? So, we will not forget about the rationale behind raising the exception.

unittest_requirements.txt
git-hyagi: Hum... I probably had issues before removing the test_convert.py file:
https://github.com/pulp/pulp_container/pull/1529/files#diff-2f5092e427a2a7997e2ce75c8c252cbd9d3d30b33f2ab32dbe1b241194de98bdL6
I just tried again without it and had no dependency errors. Removing it.

pulp_container/tests/functional/api/test_content_cache.py
git-hyagi: I don't remember now, but it was probably an error in test_05_different_headers.
I'm rerunning the tests without this block to confirm that the "corner case" was actually because of wrong headers in my tests.

pulp_container/app/registry.py
git-hyagi: > I think this part is not required inside dispatch_tag.

Agreed. I added this part when I was having issues with the test_05_different_headers and setting "_wrong headers_". I'm rerunning the tests without this modification in dispatch_tag and will check if the tests will pass.

> How about we remove the code and add the highlighted comment just above the raise PathNotResolved call placed at the end of the get_tag method? So, we will not forget about the rationale behind raising the exception.

Good idea!

ipanova:
you forgot about redirects.py file, we should not 302-->404 but raise 404 right away

pulp_container/app/registry_api.py
```diff
@@ -383,6 +384,13 @@ def create_dr(self, path, request):
 
         return distribution, repository
 
+    def handle_old_manifest_version(self, media_type, pk):
+        """
+        For non supported manifest versions, instead of redirecting the request, return earlier.
+        """
+        if media_type == MEDIA_TYPE.MANIFEST_V1:
+            raise ManifestNotFound(reference=pk)
+
```
lubosmj, line 393: If a manifest's media type is `MEDIA_TYPE.MANIFEST_V1`, we should not raise `ManifestNotFound`. Instead, let's return it because we have got the same media type as requested by a user. Also, the `media_type` variable can contain comma separated values.

pulp_container/app/registry_api.py
```diff
@@ -962,6 +970,9 @@ def get(self, request, path, pk):
     @RegistryApiCache(base_key=lambda req, cac: find_base_path_cached(req, cac))
     def handle_safe_method(self, request, path, pk):
         """Handles safe requests for Blobs."""
+        # Fail early in case of unsupported manifest schemas
+        self.handle_old_manifest_version(request.accepted_media_type, pk)
```
lubosmj, line 974: We do not care about media types when it comes to blobs. You can safely remove this check here.

pulp_container/app/redirects.py
```diff
@@ -38,6 +38,13 @@ def issue_tag_redirect(self, tag):
         """
         Issue a redirect for the passed tag.
         """
+        manifest_media_type = tag.tagged_manifest.media_type
+        if (
+            manifest_media_type not in get_accepted_media_types(self.request.headers)
+            and manifest_media_type != MEDIA_TYPE.MANIFEST_V1
+        ):
+            raise Http404("Unsupported manifest version: %s" % manifest_media_type)
```
lubosmj, line 46: Have you tried raising `ManifestNotFound` instead?

CHANGES/1509.removal
```diff
@@ -0,0 +1,5 @@
+Removed the manifest schema conversion machinery. If the manifest is stored locally in the newer
+format and old clients request v2 schema1 manifest they will receive 404. v2 schema1 manifest is
+still going to be mirrored from remote source during sync if available and passed to the old clients
+on the request.
+This change also removed the ``pyjwkest`` dependency, since it is not used anymore.
```
lubosmj, line 5: I am not sure if this dependency removal is necessary to announce. @ipanova, what do you think?

pulp_container/app/redirects.py
git-hyagi: hum....not yet after this change.
I'll try it!

CHANGES/1509.removal
ipanova: I agree.

pulp_container/app/redirects.py
```diff
@@ -34,10 +35,17 @@ class FileStorageRedirects(CommonRedirects):
...
+    def issue_tag_redirect(self, tag, pk):
         """
         Issue a redirect for the passed tag.
         """
+        manifest_media_type = tag.tagged_manifest.media_type
+        if (
+            manifest_media_type not in get_accepted_media_types(self.request.headers)
+            and manifest_media_type != MEDIA_TYPE.MANIFEST_V1
+        ):
+            raise ManifestNotFound(reference=pk)
```
lubosmj, line 47: Is it possible to use `tag.name` instead? If so, there is no need to have a an additional argument.

lubosmj (commented):
Would you update the staging variant of the documentation, please?

staging_docs/user/tutorials/01-sync-and-host.md

pulp_container/app/redirects.py
```diff
@@ -60,7 +68,8 @@ class S3StorageRedirects(CommonRedirects):
 
     def issue_tag_redirect(self, tag):
         """
-        Issue a redirect or perform a schema conversion if an accepted media type requires it.
+        Issue a redirect if an accepted media type requires it or return not found if manifest
+        version is not supported.
```
ipanova, line 72: your changes are not really related but i noticed and think that we should swap the if/elif branches

ipanova (approved):
looks good,thanks!

ipanova:
@lubosmj  seems like pull-through tests is failing on azure, a fluke?

lubosmj:
> @lubosmj seems like pull-through tests is failing on azure, a fluke? https://github.com/pulp/pulp_container/actions/runs/8111223816/job/22170105281?pr=1529#step:14:2184

I have noticed this. It started failing a couple weeks ago. It might be related to https://github.com/pulp/pulp_container/issues/1493 or artifacts/content being already present in the storage and there is a race condition. I will open an issue for that next week if it occurs again.

ipanova:
> or artifacts/content being already present in the storage and there is a race condition. 

This is test of pull operation and not push, so no.

lubosmj:
Yes, the error is the same as noted in the aforementioned issue:
```
    raise NotImplementedError("Redirecting to this storage is not implemented.")

```

PR #1532, opened 2024-03-03
Distinguish between the nature of images

### LINKED ISSUE #1437: As a user I want to differentiate between different nature of images
As a user when I inspect a manifest through pulp API I can tell whether the image is :
- bootable
- application based
- flatpak
- etc
org.flatpak.ref
**Describe the solution you'd like**
Expose Image Annotations/Labels based on which a user would be able to say/search  what kid of image it is.
For example flatpak image contain `org.flatpak.ref` label. This label identifies the image as Flatpak.

**Additional context**
We might need to inspect both Label and Annotations depending on the Docker/Oci image format.
In OCI spec labels are superseded by annotations but docker format still uses labels. 

Note: Even if the image is derived, the labels/annotations are passed down to the child image.

lubosmj:
closes #1437


lubosmj:
This is the output after syncing data https://github.com/pulp/pulp_container/pkgs/container/bootc-labeled.
```
http :5001/pulp/api/v3/content/container/manifests/?is_bootable=True
```
```
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "artifact": "/pulp/api/v3/artifacts/018e05e1-d54a-743b-87c3-e641a2305f64/",
            "blobs": [],
            "config_blob": null,
            "digest": "sha256:6a4a6aded92e720d91ec9bfdb0bf80a37846ac837be652f4d87140a46dc4ba71",
            "is_bootable": true,
            "is_flatpak": false,
            "listed_manifests": [
                "/pulp/api/v3/content/container/manifests/018e05e1-d6e9-7212-b323-24c60c40579e/"
            ],
            "media_type": "application/vnd.oci.image.index.v1+json",
            "pulp_created": "2024-03-03T19:54:39.966583Z",
            "pulp_href": "/pulp/api/v3/content/container/manifests/018e05e1-d563-77dc-af34-65cedfd92aa5/",
            "pulp_last_updated": "2024-03-03T19:54:39.966598Z",
            "schema_version": 2
        },
        {
            "artifact": "/pulp/api/v3/artifacts/018e05e1-d6d0-7260-9ff2-a483de045536/",
            "blobs": [
                "/pulp/api/v3/content/container/blobs/018e05e1-d6eb-72ed-b420-a71f6c0fa12c/"
            ],
            "config_blob": "/pulp/api/v3/content/container/blobs/018e05e1-d6ed-7f53-86c8-6fd9b87173a1/",
            "digest": "sha256:23973d32066ae1cb15ce3cef7fe66a8aa87038936b6c828cc5c129d46a3a9233",
            "is_bootable": true,
            "is_flatpak": false,
            "listed_manifests": [],
            "media_type": "application/vnd.oci.image.manifest.v1+json",
            "pulp_created": "2024-03-03T19:54:39.948038Z",
            "pulp_href": "/pulp/api/v3/content/container/manifests/018e05e1-d6e9-7212-b323-24c60c40579e/",
            "pulp_last_updated": "2024-03-03T19:54:39.948052Z",
            "schema_version": 2
        }
    ]
}
```


```
http :5001/pulp/api/v3/content/container/manifests/018e05e1-d6e9-7212-b323-24c60c40579e/
```
```
{
    "annotations": {
        "org.opencontainers.image.base.digest": "sha256:695ae78b4957fef4e53adc51febd07f5401eb36fcd80fff3e5107a2b4aa42ace",
        "org.opencontainers.image.base.name": "docker.io/library/alpine:3.18"
    },
    "artifact": "/pulp/api/v3/artifacts/018e05e1-d6d0-7260-9ff2-a483de045536/",
    "blobs": [
        "/pulp/api/v3/content/container/blobs/018e05e1-d6eb-72ed-b420-a71f6c0fa12c/"
    ],
    "config_blob": "/pulp/api/v3/content/container/blobs/018e05e1-d6ed-7f53-86c8-6fd9b87173a1/",
    "digest": "sha256:23973d32066ae1cb15ce3cef7fe66a8aa87038936b6c828cc5c129d46a3a9233",
    "is_bootable": true,
    "is_flatpak": false,
    "labels": {
        "containers.bootc": "1",
        "io.buildah.version": "1.33.5"
    },
    "listed_manifests": [],
    "media_type": "application/vnd.oci.image.manifest.v1+json",
    "pulp_created": "2024-03-03T19:54:39.948038Z",
    "pulp_href": "/pulp/api/v3/content/container/manifests/018e05e1-d6e9-7212-b323-24c60c40579e/",
    "pulp_last_updated": "2024-03-03T19:54:39.948052Z",
    "schema_version": 2
}
```
Detail exposes additional fields for `annotations` and `labels`. Manifest lists inherit the nature of listed containers. However, `annotations` and `labels` are not a subject of the inheritance.

pulp_container/app/registry_api.py
```diff
@@ -1181,6 +1181,11 @@ def put(self, request, path, pk=None):
             )
             manifest = manifest_list
 
+            # once relations for listed manifests are established,
+            # it is possible to initialize the metadata, like labels
+            manifest.init_metadata(artifact)
```
ipanova, line 1186: same here and further down

pulp_container/app/tasks/download_image_data.py
```diff
@@ -64,7 +64,7 @@ async def run(self):
                 self.manifest_list_dcs.append(list_dc)
         else:
             # Simple tagged manifest
-            man_dc = self.create_tagged_manifest(
+            man_dc = await self.create_tagged_manifest(
```
ipanova, line 67: why this is needed, it's not async;
what about create_manifest_list?

pulp_container/app/tasks/sync_stages.py
```diff
@@ -386,7 +381,9 @@ def create_tagged_manifest_list(self, tag_name, saved_artifact, manifest_list_da
         manifest_list_dc.extra_data["listed_manifests"] = []
         return manifest_list_dc
 
-    def create_tagged_manifest(self, tag_name, saved_artifact, manifest_data, raw_data, media_type):
+    async def create_tagged_manifest(
```
ipanova, line 384: i see

pulp_container/tests/functional/constants.py
```diff
@@ -35,6 +35,9 @@
 # a repository containing 4 manifest lists and 5 manifests
 PULP_FIXTURE_1 = "pulp/test-fixture-1"
 
+# a dummy repository containing two manifests with an arbitrary bootc label
```
ipanova, line 38: two manifests(index and image)

pulp_container/app/viewsets.py
```diff
@@ -249,14 +251,34 @@ class TagViewSet(ContainerContentQuerySetMixin, ReadOnlyContentViewSet):
     }
 
 
-class ManifestViewSet(ContainerContentQuerySetMixin, ReadOnlyContentViewSet):
+class ManifestViewSetDetailMixin:
+    """
+    A mixin that allows differentiating between listing manifests and retrieving a single manifest.
```
ipanova, line 256: in the nutshell this is nice idea, in practice i think users will rely to see same fields for manifests whether on list of get. I.e. they will need to write 2 different parsers(a bit of inconvenience)  that will be able to cope with no labels/annotations fields during listing and take them into account when getting a specific instance.

pulp_container/app/migrations/0038_add_manifest_metadata_fields.py
```diff
@@ -0,0 +1,148 @@
...
+            fields_to_update = ['annotations', 'labels', 'is_bootable', 'is_flatpak']
+            manifests_qs.model.objects.bulk_update(
+                manifests_to_update,
+                fields_to_update,
+            )
+
+
+def update_manifest(manifest):
+    with suppress(ObjectDoesNotExist):
+        artifact = manifest._artifacts.get()
```
ipanova, line 50: i don't know if we should keep this as a migration or rather a management command. this is expensive

ipanova (commented):
Thanks for the PR it does exactly what it meant to, however, I would definitely rework how you are initializing the metadata because you are reading from file multiple times data you have access to and can just pass it. This is a concern especially with objects storage because you pay for every get operation on the file.
Also this PR code path is tightly related to the artifactless manifest, so Michal will need to rework after you, how init_metadata is done.  If you don't want to address my previous point I would suggest then wait until artifactless manifests/config is done.

pulp_container/app/viewsets.py
lubosmj: You do recommend returning `annotations` and `labels` for the list view too, right? Or, have you considered not exposing such fields at all?

pulp_container/app/migrations/0038_add_manifest_metadata_fields.py
lubosmj: Makes sense. Let me write a new `populate-annotations-labels` command.

pulp_container/app/viewsets.py
ipanova: no no, i mean exposing them for list and get.
@mdellweg any opinion on the approach?

pulp_container/app/migrations/0038_add_manifest_metadata_fields.py
ipanova: see 0033_raise_warning_for_repair.py

pulp_container/app/registry.py
```diff
@@ -460,6 +460,9 @@ async def init_pending_content(self, digest, manifest_data, media_type, artifact
             media_type=media_type,
             config_blob=config_blob,
         )
+
+        await sync_to_async(manifest.init_metadata)(artifact)
```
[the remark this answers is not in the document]
lubosmj: Sounds good. I like the idea.

pulp_container/app/viewsets.py
mdellweg: This is a hard question that should ideally be answered on the application level. Do we promise our users in general that the objects from list and view calls are equivalent.

I fear that ship has sailed already. We do have similar things for tasks and rpm packages.

lubosmj: I decided to use one and only `ManifestSerializer` to not do any other gimmicks. Thanks for your inputs!

pulp_container/app/migrations/0038_add_manifest_metadata_fields.py
```diff
@@ -0,0 +1,47 @@
...
+import warnings
+
+from django.db import migrations, models
+
+
+def print_warning_for_initializing_image_nature(apps, schema_editor):
+    warnings.warn(
+        "Manifests containing additional metadata, like image manifest annotations and "
+        "image configuration labels, need to be updated. Please, run 'pulpcore-manager "
+        "init-image-nature' to extract and initiliaze the metadata from related artifacts."
```
lubosmj, line 11: Update this to a more friendlier manner: If you are interested in initializing annotations and labels, run this django command...

pulp_container/app/models.py
```diff
@@ -107,6 +118,73 @@ class Manifest(Content):
         through_fields=("image_manifest", "manifest_list"),
     )
 
+    def init_metadata(self, manifest_data=None, manifest_artifact=None, config_artifact=None):
```
lubosmj, line 121:
```suggestion
    def init_metadata(self, manifest_data=None):
```

lubosmj, line 121:
```suggestion
    def init_metadata(self, manifest_data=None):
```
These arguments are not used anywhere right now.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -674,3 +675,8 @@ def _post_save(self, batch):
             BlobManifest.objects.bulk_create(blob_manifests, ignore_conflicts=True)
         if manifest_list_manifests:
             ManifestListManifest.objects.bulk_create(manifest_list_manifests, ignore_conflicts=True)
+
+        # the nature of manifest lists can be determined after creating the listed manifest relation
+        for ml in manifest_lists:
```
ipanova, line 680: is operations is not really related to relation creation, can you rename the variable from manifest_lits to something more explicit, like manifests_lists_nature_to_init

ipanova (commented):
looks good!

pulp_container/app/tasks/sync_stages.py
lubosmj: I was trying to say in the comment that we can run this initialization only after creating the relations between listed manifests and manifest lists. It appears that the combination of the comment and variable name causes confusion. 😞 Let me remove the comment and rename the variable.

ipanova: i think the comment is helpful. you can leave it + rename the var ;)

lubosmj: Okay, I decided to rewrite the comment instead:
```
        # after creating the relation between listed manifests and manifest lists,
        # it is possible to initialize the nature of the corresponding manifest lists
```
How do you like it now? The suggested explicit name is hard to read and understand. We iterate over manifest lists and thus it does make sense to call the list "manifest_lists".

pulp_container/app/migrations/0038_add_manifest_metadata_fields.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 4.2.10 on 2024-02-29 16:04
+import warnings
+
+from django.db import migrations, models
+
+
+def print_warning_for_initializing_image_nature(apps, schema_editor):
+    warnings.warn(
+        "Run 'pulpcore-manager init-image-nature' to initialize metadata (i.e., annotations "
+        "and labels) for all manifests. The metadata are stored in associated artifacts and "
```
ipanova, line 10: metadata is singular, hence is stored

ipanova: I would remove last sentence, what kind of important info it brings? "their contents should be accessible through the Manifest model now" what contents? all contents? or just selected ones.
Users are not interested in tech.details of how a manifests is composed, imo

ipanova: a shorter concise version: "Run 'pulpcore-manager init-image-nature' to initialize and expose metadata (i.e., annotations and labels) for the manifests"

ipanova (approved):
Left some non-blocking comments.

pulp_container/app/migrations/0038_add_manifest_metadata_fields.py
lubosmj: I wanted to be explicit when it comes to reading data from the storage back-end. If administrators run this migration, they might be a bit surprised by the the rise in expenses.

lubosmj:
Merging because the failures are unrelated to the PR.

pulp_container/app/migrations/0038_add_manifest_metadata_fields.py
ipanova: the migration itself does not read from disk anything, it is the django management command that it does

lubosmj: I meant the django-management command migration. We are getting lost in this terminology because we always talk about migrating data, xixi.

Issue #1494, opened 2024-02-01
Our OCI support is out of sync with the latest updates to the spec


ipanova:
**Version**
1. We do not support - manifest.json with artifactType, subject https://github.com/opencontainers/image-spec/blob/main/manifest.md
2.  Descriptor properties need to be updated - urls, data, artifactType https://github.com/opencontainers/image-spec/blob/main/config.md
3. We need to handle empty descriptor mediatype https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidance-for-an-empty-descriptor
4. We might want to reconsider out media_type validation for config, layers, and instead ignore if unknown https://github.com/opencontainers/image-spec/blob/main/artifacts-guidance.md https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Additional context**
Add any other context about the problem here.  Please provide links to any previous discussions via Discourse or Bugzilla.

RCMariko:
The links are great, however for future reference or just quick info - is there a specific version of the spec we are trying to align with in this issue?

ipanova:
it's latest available here https://github.com/opencontainers/image-spec, currently v1.1

lubosmj:
This issue incorporates disabling the media_type validation to some extent.

PR #1533, opened 2024-03-04
Update image config schema/mediatype validation

git-hyagi:
closes: https://github.com/pulp/pulp_container/issues/1494

### DIFF
--- a/CHANGES/1494.feature
+++ b/CHANGES/1494.feature
@@ -1,0 +1,3 @@
+Updated the OCI manifest schema validation to comply with the changes from the OCI Image Manifest
+Specification.
+Deprecated ``ADDITIONAL_OCI_ARTIFACT_TYPES`` setting in favour of the relaxed validation.
--- a/pulp_container/app/json_schemas.py
+++ b/pulp_container/app/json_schemas.py
@@ -1,6 +1,4 @@
 from pulp_container.constants import (
-    ALLOWED_ARTIFACT_TYPES,
-    ALLOWED_BLOB_CONTENT_TYPES,
     BLOB_CONTENT_TYPE,
     MEDIA_TYPE,
     SIGNATURE_TYPE,
@@ -8,16 +6,27 @@ from pulp_container.constants import (
 
 
 def get_descriptor_schema(
-    allowed_media_types, additional_properties=None, additional_required=None
+    allowed_media_types=None, additional_properties=None, additional_required=None
 ):
     """Return a concrete descriptor schema for manifests."""
 
+    media_type_config = {"type": "string"}
+    if allowed_media_types is not None:
+        media_type_config = {"type": "string", "enum": allowed_media_types}
+
     properties = {
-        "mediaType": {"type": "string", "enum": allowed_media_types},
+        "mediaType": media_type_config,
         "size": {"type": "number"},
         "digest": {"type": "string"},
         "annotations": {"type": "object", "additionalProperties": True},
-        "urls": {"type": "array", "items": {"type": "string"}},
+        "urls": {
+            "type": "array",
+            "items": {"type": "string"},
+            "format": "uri",
+            "pattern": "^https?://",
+        },
+        "data": {"type": "string", "contentEncoding": "base64"},
+        "artifactType": {"type": "string"},
     }
 
     if additional_properties:
@@ -78,13 +87,20 @@ OCI_MANIFEST_SCHEMA = {
             "type": "string",
             "enum": [MEDIA_TYPE.MANIFEST_OCI],
         },
-        "config": get_descriptor_schema(ALLOWED_ARTIFACT_TYPES),
+        "artifactType": {"type": "string"},
+        "config": get_descriptor_schema(),
         "layers": {
             "type": "array",
-            "items": get_descriptor_schema(ALLOWED_BLOB_CONTENT_TYPES),
+            "items": get_descriptor_schema(),
         },
     },
     "required": ["schemaVersion", "config", "layers"],
+    "if": {
+        "properties": {
+            "config": {"properties": {"mediaType": {"const": MEDIA_TYPE.OCI_EMPTY_JSON}}}
+        }
+    },
+    "then": {"dependentRequired": {"config.mediaType": ["artifactType"]}},
 }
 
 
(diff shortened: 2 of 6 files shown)


pulp_container/app/json_schemas.py
```diff
@@ -1,19 +1,22 @@
...
-    ALLOWED_BLOB_CONTENT_TYPES,
     BLOB_CONTENT_TYPE,
     MEDIA_TYPE,
     SIGNATURE_TYPE,
 )
 
 
 def get_descriptor_schema(
-    allowed_media_types, additional_properties=None, additional_required=None
+    allowed_media_types=[], additional_properties=None, additional_required=None
```
lubosmj, line 10: Defining an empty list as a default argument in a function may pose side effects.
```
>>> def foo(param=[]):
...     param.append(1)
...     return param
... 
>>> 
>>> foo()
[1]
>>> foo()
[1, 1]
```

pulp_container/app/json_schemas.py
```diff
@@ -1,19 +1,22 @@
...
 ):
     """Return a concrete descriptor schema for manifests."""
 
+    MEDIA_TYPE_CONFIG = {"type": "string"}
```
lubosmj, line 14: Please, try using variables with lowercase inside the function's scope. This is not a real constant, is it?

Besides that, can you point me to the specification that allows an *empty* media type? The `get_descriptor_schema` function is used in various places, not just for configs.

pulp_container/app/json_schemas.py
```diff
@@ -1,19 +1,22 @@
...
+    if len(allowed_media_types) > 0:
+        MEDIA_TYPE_CONFIG = {"type": "string", "enum": allowed_media_types}
+
     properties = {
-        "mediaType": {"type": "string", "enum": allowed_media_types},
+        "mediaType": MEDIA_TYPE_CONFIG,
```
lubosmj, line 19: I am missing a new property here: `data`.

See https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties.

lubosmj (changes requested):
Thanks for opening this PR!

I think the idea was also to partially disable the validation because of newly emerging media types and configurations for blobs.

Due to current circumstances and the number of possible media types, it is worth reconsidering the need for `ADDITIONAL_OCI_ARTIFACT_TYPES`. This goes hand in hand with the 4th point on the attached issue:
> We might want to reconsider out media_type validation for config, layers, and instead ignore if unknown https://github.com/opencontainers/image-spec/blob/main/artifacts-guidance.md https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage

Do you mind updating the validation schema accordingly?

EDIT: the partial disablement i was referencing to is related to removing "enum" from the `media_typey` property (which is what @git-hyagi did). Therefore, it somehow matches the expectations. I thought that we were about to remove most of the `get_descriptor_schema` code.

CHANGES/1494.feature
```diff
@@ -0,0 +1 @@
+Updated our OCI MANIFEST SCHEMA to comply with the changes from OCI Image Manifest Specification.
```
lubosmj, line 1:
```suggestion
Updated  the OCI manifest schema validation to comply with the OCI Image Manifest Specification.
```

Also, can we add a disclaimer about deprecating `ADDITIONAL_OCI_ARTIFACT_TYPES` here? It should be visible because we are going to kill it on the next release.

pulp_container/app/json_schemas.py
```diff
@@ -1,23 +1,32 @@
...
         "annotations": {"type": "object", "additionalProperties": True},
-        "urls": {"type": "array", "items": {"type": "string"}},
+        "urls": {
+            "type": "array",
+            "items": {"type": "string"},
+            "format": "uri",
+            "pattern": "^https?://",
+        },
+        "data": {"type": "string", "contentEncoding": "base64"},
+        "artifactType": {"type": "string"},
```
ipanova, line 29: you forgot subject :grimacing:

pulp_container/constants.py
```diff
@@ -29,6 +30,7 @@
             MEDIA_TYPE.MANIFEST_LIST,
             MEDIA_TYPE.INDEX_OCI,
             MEDIA_TYPE.MANIFEST_OCI,
+            MEDIA_TYPE.OCI_EMPTY_JSON,
```
ipanova, line 33: this does not go into the accept headers. only manifests mediatype types go. this is  blob instead.

pulp_container/constants.py
```diff
@@ -39,6 +41,7 @@
         MEDIA_TYPE.MANIFEST_V1_SIGNED,
         MEDIA_TYPE.MANIFEST_V2,
         MEDIA_TYPE.MANIFEST_OCI,
+        MEDIA_TYPE.OCI_EMPTY_JSON,
```
ipanova, line 44: same here, this is not a new type of manifest, just a blob type

pulp_container/app/json_schemas.py
```diff
@@ -78,13 +87,20 @@ def get_descriptor_schema(
...
+            "items": get_descriptor_schema(),
         },
     },
     "required": ["schemaVersion", "config", "layers"],
+    "if": {
+        "properties": {
+            "config": {"properties": {"mediaType": {"const": MEDIA_TYPE.OCI_EMPTY_JSON}}}
+        }
+    },
+    "then": {"dependentRequired": {"config.mediaType": ["artifactType"]}},
```
ipanova, line 103: nice!

pulp_container/app/json_schemas.py
git-hyagi: hum... not sure if I'm looking at the wrong schema, but I could not find the definition for `subject` in the descriptor:
https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties

pulp_container/app/json_schemas.py
```diff
@@ -78,13 +87,20 @@ def get_descriptor_schema(
             "type": "string",
             "enum": [MEDIA_TYPE.MANIFEST_OCI],
         },
-        "config": get_descriptor_schema(ALLOWED_ARTIFACT_TYPES),
+        "artifactType": {"type": "string"},
```
ipanova, line 90: sorry i was supposed to comment here: you forgot the subject! :D

pulp_container/app/json_schemas.py
ipanova: https://github.com/pulp/pulp_container/pull/1533/files#r1516647629

pulp_container/app/json_schemas.py
ipanova: it is part of index and image manifest

git-hyagi: ahnn!! nice catch!

PR #1580, opened 2024-04-05
Add a new artifactless Blob Config model

git-hyagi:
closes: https://github.com/pulp/pulp_container/issues/1296


pulp_container/app/models.py
```diff
@@ -38,6 +38,54 @@
...
+
+    TYPE = "config-blob"
+
+    data = models.TextField()
+    created = models.DateTimeField()
+    author = models.TextField(default="", blank=True)
+    os = models.TextField()
+    architecture = models.TextField()
+    digest = models.TextField(db_index=True)
+    config = models.JSONField(default=dict)
```
lubosmj, line 69:
```suggestion
    json_data = models.JSONField(default=dict)
```

pulp_container/app/registry_api.py
```diff
@@ -607,18 +606,16 @@ def get(self, request):
                     tag.name, tag.tagged_manifest, req_oss, req_architectures, manifests
                 )
             for manifest, tagged in manifests.items():
-                with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
-                    raw_data = file.read()
-                config_data = json.loads(raw_data)
-                labels = config_data.get("config", {}).get("Labels")
+                config_data = manifest.config
```
lubosmj, line 609:
```suggestion
                config = manifest.config
```
Please, use correct naming conventions.

Note that this will not work in case the user has not run the django-management command to migrate artifacts.

We need to support two code paths -- for users who have run the django-management command and those who have not run it yet.

pulp_container/app/registry_api.py
```diff
@@ -919,9 +916,9 @@ def put(self, request, path, pk=None):
                 artifact.touch()
 
         blob = self.create_blob(artifact, digest)
+        repository.pending_blobs.add(blob)
```
lubosmj, line 919: Is this change necessary?

pulp_container/app/registry_api.py
```diff
@@ -979,12 +976,20 @@ def handle_safe_method(self, request, path, pk):
...
+                repository = repository.cast()
+                try:
+                    blob = repository.pending_blobs.get(digest=pk)
+                    blob.touch()
+                except models.Blob.DoesNotExist:
+                    try:
+                        blob = repository.pending_config_blobs.get(digest=pk)
+                        blob.touch()
+                    except models.ConfigBlob.DoesNotExist:
+                        raise BlobNotFound(digest=pk)
```
lubosmj, line 992: If we could have the relation between the Blob and ConfigBlob model, we could reduce the complexity a bit.

pulp_container/app/registry_api.py
```diff
@@ -1087,14 +1092,16 @@ def get_content_units_to_add(self, manifest, tag):
             models.MEDIA_TYPE.MANIFEST_LIST,
             models.MEDIA_TYPE.INDEX_OCI,
         ):
-            for listed_manifest in manifest.listed_manifests:
+            for listed_manifest in manifest.listed_manifests.all():
                 add_content_units.append(listed_manifest.pk)
+                add_content_units.append(listed_manifest.config_id)
                 add_content_units.append(listed_manifest.config_blob_id)
```
lubosmj, line 1098: Now, we should not add `config_blob_id` to the list of added content. While appending the content, we should follow the new workflow only. Please, remove the same code a couple of lines below.

pulp_container/app/registry_api.py
```diff
@@ -1214,11 +1224,16 @@ def put(self, request, path, pk=None):
...
 
             config_digest = config_layer.get("digest")
-            found_config_blobs = models.Blob.objects.filter(
-                digest=config_digest, pk__in=content_pks
-            )
+            found_config_blobs = models.ConfigBlob.objects.filter(digest=config_digest)
+            # create the artifactless ConfigBlob from Blob if it is not stored yet
             if not found_config_blobs.exists():
-                raise BlobInvalid(digest=config_digest)
+                blob = models.Blob.objects.filter(digest=config_digest)
```
lubosmj, line 1230: What if there is not such a blob? I suppose we can raise `BlobInvalid` in this case. I would put this code in a `transaction` so that we will succeed only if we could initialize a config blob and remove the "placeholder" blob.

pulp_container/app/serializers.py
```diff
@@ -132,6 +140,37 @@ class Meta:
...
+class ConfigBlobSerializer(NoArtifactContentSerializer):
+    """
+    Serializer for Blob Config Manifest
+    """
+
+    architecture = serializers.CharField(
+        help_text="The CPU architecture which the binaries in this image are built to run on"
+    )
+    os = serializers.CharField(
+        help_text="The name of the operating system which the image is built to run on"
```
lubosmj, line 152: Let's wrap the help messages around with the `gettext` call. This enables others to localize Pulp.

```
help_text=_("blabla
```

pulp_container/app/tasks/builder.py
```diff
@@ -34,15 +35,25 @@ def get_or_create_blob(layer_json, manifest, path):
...
-        BlobManifest(manifest=manifest, manifest_blob=blob).save()
+        if layer_json["mediaType"] in [MEDIA_TYPE.CONFIG_BLOB_OCI, MEDIA_TYPE.CONFIG_BLOB]:
+            with open(layer_file_name, "r") as content_file:
+                raw_data = content_file.read()
+            content_data = json.loads(raw_data)
+            config_blob = ConfigBlob.build(
+                raw_data=raw_data, digest=layer_json["digest"], content_data=content_data
+            )
+            config_blob.save()
+            return config_blob
```
lubosmj, line 48: What happens to the old blob we have just saved? It seems to be redundant because we may create and return the config blob instead. Can we add the if/else statement before the `blob.save()` call, please?

pulp_container/app/tasks/recursive_add.py
```diff
@@ -57,12 +58,13 @@ def recursive_add_content(repository_pk, content_units):
...
-        Blob.objects.filter(pk__in=content_units)
-        | Blob.objects.filter(pk__in=manifests_to_add.values_list("blobs", flat=True))
-        | Blob.objects.filter(pk__in=manifests_to_add.values_list("config_blob", flat=True))
+    blobs_to_add = Blob.objects.filter(pk__in=content_units) | Blob.objects.filter(
+        pk__in=manifests_to_add.values_list("blobs", flat=True)
     )
 
+    config_blobs_to_add = ConfigBlob.objects.filter(
+        pk__in=manifests_to_add.values_list("config", flat=True)
+    )
```
lubosmj, line 67: Will the copy/remove workflow work for config blobs that have not been migrated yet? I think not, but this might be a corner case we may want to support for a couple of releases. We will want to have `Blob.objects.filter(pk__in=manifests_to_add.values_list("config_blob", flat=True))` included as well. Do you agree?

pulp_container/app/tasks/sync_stages.py
```diff
@@ -355,9 +356,9 @@ async def handle_blobs(self, manifest_dc, content_data):
             await self.put(blob_dc)
         layer = content_data.get("config", None)
         if layer:
-            blob_dc = self.create_blob(layer, deferred_download=False)
-            manifest_dc.extra_data["config_blob_dc"] = blob_dc
-            await self.put(blob_dc)
+            config_blob_dc = await self.create_config_blob(layer)
+            manifest_dc.extra_data["config_blob_dc"] = config_blob_dc
+            await self.put(config_blob_dc)
```
lubosmj, line 361: If you do not mind, please, rename `layer` to something self-explanatory, like `config`.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -515,6 +516,25 @@ async def create_listed_manifest(self, manifest_data):
         )
         return {"manifest_dc": man_dc, "platform": platform, "content_data": content_data}
 
+    async def create_config_blob(self, blob_data):
+        digest = blob_data.get("digest") or blob_data.get("blobSum")
```
lubosmj, line 520: I could not find the `blobSum` field in https://github.com/opencontainers/image-spec/blob/main/config.md. Can you verify this exists?

lubosmj (commented):
Thank you for opening this PR! 😍 

Do you mind reviewing the comments?

**It appears that we do not support a scenario for users who did not run the django-management command, meaning that they still have config blobs referencing artifacts living inside Pulp.** Is that correct? If so, we should be still able to serve those artifacts.

Have you tried running a Pulp instance with a 3rd-party storage backend (like [S3](https://github.com/pulp/oci_env/tree/main/profiles/pminio)) to see if we can still serve artifacts after running just a database migration while omitting running the django-management command? This is to verify the behaviour after the partial upgrade.

Also, I would like to introduce a couple of TODO comments in registry_api.py suggesting that in the near future we will no longer use redirects to serve config blobs' data. For me, it does not make sense to issue a redirect if we could return config_blob.data directly from registry_api.py. I suggested the same approach for the artifactless manifests PR.

pulp_container/app/registry_api.py
lubosmj: Maybe it would make sense to revisit a concept of having a single blob model which has a ".data" field, containing the extracted file data in case of a config blob.

pulp_container/app/registry.py
```diff
@@ -292,15 +299,19 @@ async def get_by_digest(self, request):
...
                     return web.Response(text=raw_manifest, headers=headers)
-                elif content_type == "blobs":
+                elif content_type == "blobs" or content_type == "config-blobs":
                     # there might be a case where the client has all the manifest data in place
                     # and tries to download only missing blobs; because of that, only the reference
                     # to a remote blob is returned (i.e., RemoteArtifact)
                     blob = await pull_downloader.init_remote_blob()
                     ca = await blob.contentartifact_set.afirst()
                     return await self._stream_content_artifact(request, web.StreamResponse(), ca)
+                # elif content_type=="config-blobs":
```
dkliban, line 309: did you mean to leave this comment here?

git-hyagi:
After some discussions, we decided to keep with https://github.com/pulp/pulp_container/pull/1603
We will continue the investigations and analysis of the benefits of an artifactless Blob on #1603.

PR #1589, opened 2024-04-13
Check if the Authorization header for Basic Authentication is valid

### LINKED ISSUE #1577: RegistryAuthentication fails to fall back to PulpRemoteUserAuthentication if no authorization header exists
**Version**
I believe any version of pulp-container that supports container push (for me, specifically 2.16.3)

**Describe the bug**
The following auth settings are set:

```
REMOTE_USER_ENVIRON_NAME = 'HTTP_REMOTE_USER'
AUTHENTICATION_BACKENDS = ['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']

REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
    'rest_framework.authentication.SessionAuthentication',
    'pulpcore.app.authentication.PulpRemoteUserAuthentication'
)
TOKEN_AUTH_DISABLED=True
```

When trying to push container content (like blobs) to Pulp, authentication always fails even if REMOTE_USER is correctly set to 'admin'.

It's caused by the following: https://github.com/pulp/pulp_container/blob/ef0a1943458f4a071dfed919b450df791e0a40a5/pulp_container/app/token_verification.py#L82-L88

If there is no auth header, or if it doesn't have "Bearer ..." in it, `super().authenticate(request)` returns `None` instead of throwing `AuthenticationFailed`. This in turn causes the AnonymousUser to be used instead of admin.

The rules for returning `None` vs throwing `AuthenticationFailed` are here: https://github.com/encode/django-rest-framework/blob/085b7e166ba80aa973645e5249b441f2dbdc0c96/rest_framework/authentication.py#L66-L67

**To Reproduce**
Try to push container content, like blobs with the settings above. HTTP_REMOTE_USER should be set to 'admin'.

**Expected behavior**
Auth is successful

**Additional context**
This was tested on a Katello development install with in-development container push code. Ping me if anything should be tested on my environment.

lubosmj:
If the header is not valid, DRF returns None when calling the
authenticate() method. This can cause troubles when users are
leveraging the remote authentication because Pulp thinks they
are anonymous users. In the end, authorized users cannot
push or pull content from Pulp. This affects only admin users
in scenarios where the token authentication is disabled.

closes #1577

### DIFF
--- a/pulp_container/app/token_verification.py
+++ b/pulp_container/app/token_verification.py
@@ -64,28 +64,28 @@ class RegistryAuthentication(BasicAuthentication):
     A basic authentication class that accepts empty username and password as anonymous.
     """
 
-    PULP_AUTHENTICATION_CLASS = "pulpcore.app.authentication.PulpRemoteUserAuthentication"
+    PULP_REMOTE_AUTHENTICATION_CLASS = "pulpcore.app.authentication.PulpRemoteUserAuthentication"
     AUTH_CLASSES = settings.REST_FRAMEWORK["DEFAULT_AUTHENTICATION_CLASSES"]
+    ALLOWS_REMOTE_AUTHENTICATION = PULP_REMOTE_AUTHENTICATION_CLASS in AUTH_CLASSES
 
     def authenticate(self, request):
         """
         Perform basic authentication with the exception to accept empty credentials.
 
-        For anonymous user, Podman sends 'Authorization': 'Basic Og=='.
-        This represents ":" in base64.
-
         If basic authentication could not success, remote webserver authentication is considered.
         """
-        if request.headers.get("Authorization") == "Basic Og==":
-            return (AnonymousUser, None)
-
         try:
-            return super().authenticate(request)
+            user = super().authenticate(request)
         except AuthenticationFailed:
-            if self.PULP_AUTHENTICATION_CLASS in self.AUTH_CLASSES:
+            if self.ALLOWS_REMOTE_AUTHENTICATION:
                 return RemoteUserRegistryAuthentication().authenticate(request)
             else:
                 raise
+
+        if user is None and self.ALLOWS_REMOTE_AUTHENTICATION:
+            return RemoteUserRegistryAuthentication().authenticate(request)
+        else:
+            return user
 
 
 class RemoteUserRegistryAuthentication(RemoteUserAuthentication):
--- a/CHANGES/1577.bugfix
+++ b/CHANGES/1577.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug that disallowed users from leveraging the remote authentication.


lubosmj:
I successfully tested image pushing with the following settings (with the enabled basic authentication and remote authentication at the same time):

#### dynaconf list
```
AUTHENTICATION_BACKENDS<list> ['django.contrib.auth.backends.ModelBackend', 
 'pulpcore.backends.ObjectRolePermissionBackend',
 'pulpcore.app.authentication.PulpNoCreateRemoteUserBackend'] 

REST_FRAMEWORK<dict> {'DEFAULT_AUTHENTICATION_CLASSES': ['rest_framework.authentication.BasicAuthentication',
                                    'rest_framework.authentication.SessionAuthentication',
                                    'rest_framework.authentication.SessionAuthentication',
                                    'pulpcore.app.authentication.PulpRemoteUserAuthentication'],

REMOTE_USER_ENVIRON_NAME<str> 'HTTP_REMOTE_USER'
```

Note that when **omitting** `['django.contrib.auth.backends.ModelBackend', 'pulpcore.backends.ObjectRolePermissionBackend']` in `AUTHENTICATION_BACKENDS`, Pulp receives [AuthenticationFailed](https://github.com/encode/django-rest-framework/blob/9864c47018238dd9f8477df0e4ab2ef62a24fd2b/rest_framework/authentication.py#L101) on `super().authenticate(request)` if a user is logged in as `admin`. **Could this be the reason why it worked before? Katello could have used Basic Authentication together with Remote Authentication (https://github.com/pulp/pulp_container/issues/558).**

#### etc/nginx/pulp/pulp_container.conf
```
location /v2/ {
    ...
    proxy_set_header Remote_User admin;
}

location /extensions/v2/ {
    ...
    proxy_set_header Remote_User admin;
}

location /pulp/container/ {
    ...
    proxy_set_header Remote_User admin;
}

location /token/ {
    ...
    proxy_set_header Remote_User admin;
}
```

1. Test **with** proxy headers being set (using both the basic authentication and remote authentication):
```
podman logout localhost:5001
podman push localhost:5001/myorg/mygroup/ubuntu:latest --tls-verify=false  # OK, using implicit 'Remote-User': 'admin' header being favoured
podman login localhost:5001 -u admin -p password --tls-verify=false
podman push localhost:5001/myorg/mygroup/ubuntu:latest --tls-verify=false  # OK, using implicit 'Remote-User': 'admin' header being favoured
```

2. Test **without** proxy headers being set (using the basic authentication only):
```
podman logout localhost:5001
podman push localhost:5001/myorg/mygroup/ubuntu:latest --tls-verify=false  # Error: writing manifest: uploading manifest latest to localhost:5001/myorg/mygroup/ubuntu: denied: Access to the requested resource is not authorized.
podman login localhost:5001 -u admin -p password --tls-verify=false
podman push localhost:5001/myorg/mygroup/ubuntu:latest --tls-verify=false  # OK, using basic auth header, 'Authorization': 'Basic YWRtaW46cGFzc3dvcmQ='
```

pulp_container/app/token_verification.py
```diff
@@ -80,13 +81,18 @@ def authenticate(self, request):
             return (AnonymousUser, None)
 
         try:
-            return super().authenticate(request)
+            result = super().authenticate(request)
```
lubosmj, line 84: `None` is returned only when https://github.com/encode/django-rest-framework/blob/9864c47018238dd9f8477df0e4ab2ef62a24fd2b/rest_framework/authentication.py#L66 this is evaluated to `True`. We exploit this check.

ianballou (approved):
This works well testing container push with Katello.

pulp_container/app/token_verification.py
mdellweg: That sounds brittle. Even with this explanation it's hard to understand. You should have a comment at the very least. Is the "empty username and password" working at all with remote auth?

lubosmj: What does the empty username and password mean in this context? If there is no `Remote_User` passed with the request, and no Authorization header is present, Pulp consider this request being anonymous, allowing standard pull operations for public repositories.

mdellweg: What i mean is: When remote auth is configured, and the reverse proxy terminates the authentication process, does it also accept empty for anonymous?

lubosmj: Yes, `RemoteUserRegistryAuthentication().authenticate(request)` returns `None` if there is no `Remote_User` header set.

lubosmj: I renamed `result` into a more explicit value: `authenticated_user`.

lubosmj:
After discussing this with Ina, I will try to test this once again with different settings for the remote authentication and return either `(AnonymousUser, None)` or `None` from the `authenticate()` method to get rid of discrepancies.

ipanova:
While you listed the combinations you tested of auth backends, did you test auth classes with remote auth only and remote auth + basic auth?

ipanova (approved):

ianballou (approved):
Verified again on Katello latest, still works :+1:

ianballou:
@lubosmj I'd like to request a backport for this on the 2.16 branch if possible. It applied cleanly on my system.

lubosmj:
@ianballou, I have just released https://github.com/pulp/pulp_container/releases/tag/2.16.8!

ianballou:
Thanks @lubosmj !

Issue #1537, opened 2024-03-08
Deprecate `ADDITIONAL_OCI_ARTIFACT_TYPES` in favour of more relaxed schema validation


lubosmj:
In the second upcoming release (2.20), we should remove the `ADDITIONAL_OCI_ARTIFACT_TYPES` setting. We decided to deprecate this setting.

References:
https://github.com/pulp/pulp_container/commit/7ace741f6620b09b6d6dae65eaa3793d61af68d4
https://github.com/pulp/pulp_container/commit/d6201aad9a9befdcdbaf8d4e578da380e118e8d2

PR #1590, opened 2024-04-15
Remove deprecated ADDITIONAL_OCI_ARTIFACT_TYPES

git-hyagi:
closes: #1537

### DIFF
--- a/docs/workflows/oci-artifacts.rst
+++ b/docs/workflows/oci-artifacts.rst
@@ -3,58 +3,14 @@ Managing additional OCI media types
 
 .. _default-oci-types:
 
-By default the following list of media types is enabled in the Container Registry::
+Pulp is not only a container registry, it also supports OCI artifacts by leveraging the config property on the image manifest.
+Here are some examples of compliant OCI artifacts supported by `pulp_container` plugin::
 
- * OCI images
- * Helm
- * Cosign, SBOMs, attestations
+ * [OCI images](./workflows/build-containerfile.rst)
+ * [Helm](./workflows/helm-support.rst)
+ * [Flatpak images](./workflows/flatpak-support.rst)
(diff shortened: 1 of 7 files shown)


lubosmj (dismissed):

docs/workflows/oci-artifacts.rst
```diff
@@ -1,60 +0,0 @@
-Managing additional OCI media types
-===================================
-
-.. _default-oci-types:
-
-By default the following list of media types is enabled in the Container Registry::
```
ipanova, line 6: I would not remove entirely this file and leave a section in documentation stating that we are an OCI artifact registry. I would reword some of these docs with few examples what kind of additional to regular mediatypes we support ( out of the box) so it is clear for the end user that we are not just container but also artifact registry.

ipanova (changes requested):

docs/workflows/oci-artifacts.rst
lubosmj: How about we mention this information on our landing page (https://docs.pulpproject.org/pulp_container/)? I do not think it is necessary to have a dedicated page for it.

lubosmj: @git-hyagi, do you think we should keep this doc section? If so, we may want to revisit the paragraphs and state clearly that we support `OCI Artifacts`, e.g., by referencing [helm charts](https://github.com/pulp/pulp_container/blob/55a5af36b4fa180a0df19d98479a010da903d7c3/docs/workflows/helm-support.rst) and providing more insights into what a generic OCI Artifact might look like. Of course, omitting the removed setting.

git-hyagi: I am in favor of keeping this doc section and rewriting the paragraphs to make it clear that this not just a container registry.

PR #1603, opened 2024-04-24
Add a new `data` field to Blob model

git-hyagi:
This is an alternative solution for issue #1296.
Instead of creating a new model, this commit adds a new field to the existing Blob model and, to avoid a lot of issues to identify the file type (ASCII for container configs and binary for layers), will keep the blobs (configs and layers) in the database if they have a size < 22Kb.


pulp_container/app/management/commands/container-handle-image-data.py
```diff
@@ -6,10 +6,11 @@
 
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.management import BaseCommand
+from django.core.files.storage import default_storage as storage
 
 from pulp_container.app.models import Manifest
 
-from pulp_container.constants import MEDIA_TYPE
+from pulp_container.constants import ARTIFACTLESS_BLOB_SIZE, MEDIA_TYPE
```
ipanova, line 13: how about DB_BLOB_SIZE?

pulp_container/app/models.py
```diff
@@ -46,6 +46,7 @@ class Blob(Content):
 
     Fields:
         digest (models.TextField): The blob digest.
+        data (models.TextField): The container config content.
```
ipanova, line 49: this is binary

pulp_container/app/models.py
```diff
@@ -138,10 +140,17 @@ def init_annotations(self, manifest_data=None):
...
-            config_data, _ = get_content_data(config_artifact)
-            self.labels = config_data.get("config", {}).get("Labels") or {}
+            try:
+                # if config-blob is stored in the database
+                config = json.loads(self.config_blob.config_blob.instance.data)["config"]
+                if "Labels" in config.keys():
+                    self.labels = config["Labels"] or {}
+            except json.decoder.JSONDecodeError:
+                # if not, we'll retrieve the config from file
+                config_artifact = self.config_blob._artifacts.get()
```
ipanova, line 150: I just realized. I know the chances are small, but this will fail if the image is after reclaim_space. What should we do?
a) nothing, we will not initialize labels because file is unavailable
b) fetch the file

I tend to go with a) because if space was intentionally reclaimed then it would be a surprise for the user to again see that data occupying the space.

pulp_container/app/models.py
```diff
@@ -56,6 +57,7 @@ class Blob(Content):
     TYPE = "blob"
 
     digest = models.TextField(db_index=True)
+    data = models.BinaryField(blank=True)
```
ipanova, line 60: data can be null if the blob is stored on disk

pulp_container/app/models.py
```diff
@@ -138,10 +140,17 @@ def init_annotations(self, manifest_data=None):
 
     def init_labels(self):
         if self.config_blob:
-            config_artifact = self.config_blob._artifacts.get()
-
-            config_data, _ = get_content_data(config_artifact)
-            self.labels = config_data.get("config", {}).get("Labels") or {}
+            try:
+                # if config-blob is stored in the database
+                config = json.loads(self.config_blob.config_blob.instance.data)["config"]
```
ipanova, line 145: should this be self.config_blob.data?

pulp_container/app/models.py
```diff
@@ -138,10 +140,17 @@ def init_annotations(self, manifest_data=None):
...
+                if "Labels" in config.keys():
+                    self.labels = config["Labels"] or {}
```
ipanova, line 147: by the specs config can be optional and this will fail with key error

ipanova: sorry line 145

pulp_container/app/models.py
```diff
@@ -138,10 +140,17 @@ def init_annotations(self, manifest_data=None):
 
     def init_labels(self):
         if self.config_blob:
-            config_artifact = self.config_blob._artifacts.get()
-
-            config_data, _ = get_content_data(config_artifact)
-            self.labels = config_data.get("config", {}).get("Labels") or {}
+            try:
```
ipanova, line 143: you could get rid of try/except block if you just simply check for `if self.config_blob.data`

pulp_container/app/models.py
ipanova: same observation stands for the manifest._artifacts.get() in the init_annotations...unrelated to this PR but while we are here...

pulp_container/app/registry_api.py
```diff
@@ -608,8 +610,11 @@ def get(self, request):
                     tag.name, tag.tagged_manifest, req_oss, req_architectures, manifests
                 )
             for manifest, tagged in manifests.items():
-                with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
-                    raw_data = file.read()
+                if manifest.config_blob.data != "":
+                    raw_data = manifest.config_blob.data
+                else:
+                    with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
```
ipanova, line 616: one more occurrence where this will fail if artifact was reclaimed

pulp_container/app/registry_api.py
```diff
@@ -785,19 +790,33 @@ def create_blob(self, artifact, digest):
...
-                ca = ContentArtifact.objects.get(content=blob, relative_path=digest)
-                if not ca.artifact:
-                    ca.artifact = artifact
-                    ca.save(update_fields=["artifact"])
+
+            blob_size = artifact.file.size
+            if blob_size > ARTIFACTLESS_BLOB_SIZE:
+                self._create_blob_content_artifact(artifact, blob, digest)
+            else:
+                with storage.open(artifact.file.name) as artifact_file:
```
ipanova, line 798: I would remove the dangling artifact so it does not occupy storage

pulp_container/app/registry_api.py
```diff
@@ -785,19 +790,33 @@ def create_blob(self, artifact, digest):
...
-                )
-                blob_artifact.save()
-            except IntegrityError:
-                # re-upload artifact in case it was previously removed.
-                ca = ContentArtifact.objects.get(content=blob, relative_path=digest)
-                if not ca.artifact:
-                    ca.artifact = artifact
-                    ca.save(update_fields=["artifact"])
+
+            blob_size = artifact.file.size
```
ipanova, line 794: This is comples part of code, however, I was wondering if it worth investing time so we do not create unnecessary Artifact if the size is small enough. We first initialize PulpTemporaryUploadedFile which has sizr info we could use and then depending on that create or not the Artifact.
(might be too complicated but anyway have thrown this idea on the table)

pulp_container/app/registry_api.py
```diff
@@ -987,6 +1006,15 @@ def handle_safe_method(self, request, path, pk):
             except models.Blob.DoesNotExist:
                 raise BlobNotFound(digest=pk)
 
+        if blob.data:
+            media_type = MEDIA_TYPE.CONFIG_BLOB_OCI
```
ipanova, line 1010: the media type is application/octet-stream so BLOB_CONTENT_TYPE

pulp_container/app/serializers.py
```diff
@@ -126,6 +127,9 @@ class BlobSerializer(SingleArtifactContentSerializer):
     """
 
     digest = serializers.CharField(help_text="sha256 of the Blob file")
+    artifact = SingleContentArtifactField(
```
ipanova, line 130: why this is needed?

pulp_container/app/tasks/builder.py
```diff
@@ -34,15 +34,21 @@ def get_or_create_blob(layer_json, manifest, path):
...
-        layer_artifact = Artifact.init_and_validate(layer_file_name)
-        layer_artifact.save()
         blob = Blob(digest=layer_json["digest"])
-        blob.save()
-        ContentArtifact(
-            artifact=layer_artifact, content=blob, relative_path=layer_json["digest"]
-        ).save()
-    if layer_json["mediaType"] != MEDIA_TYPE.CONFIG_BLOB_OCI:
-        BlobManifest(manifest=manifest, manifest_blob=blob).save()
+        if layer_json["mediaType"] in [MEDIA_TYPE.CONFIG_BLOB_OCI, MEDIA_TYPE.CONFIG_BLOB]:
```
ipanova, line 38: I don't think this code is correct. Are not you suppose to check on the layer size and depending on that you whether save it as artifact or not?

pulp_container/app/tasks/sync_stages.py
```diff
@@ -353,9 +353,9 @@ async def handle_blobs(self, manifest_dc, content_data):
             blob_dc = self.create_blob(layer)
             manifest_dc.extra_data["blob_dcs"].append(blob_dc)
             await self.put(blob_dc)
-        layer = content_data.get("config", None)
-        if layer:
-            blob_dc = self.create_blob(layer, deferred_download=False)
+        config = content_data.get("config", None)
+        if config:
+            blob_dc = await self.create_config_blob(config)
```
ipanova, line 358: I don't think it matters if it is a config blob or binary blob. You need to adjust `create_blob` that depending on size, the blob will be downloaded and save into DB or a Declarative Artifact will be created and passed further down to the pipeline.
Manifest.json file has `size` for each layer descriptor. It is a mandatory field so you can use it.

pulp_container/app/serializers.py
git-hyagi: This was a trick part. `BlobSerializer` inherits from `SingleArtifactContentSerializer` in which an artifact is a required field. With this PR, an artifact for `Blob.data` is now optional (because depending on the size of the blob we would store it in the database or keep it as a file).

Removing the `required=False` constraint would work, but our bindings would fail when we store the config or layer in DB (no artifact would be created) with:
```
>           raise ValueError("Invalid value for `artifact`, must not be `None`")  # noqa: E501
E           ValueError: Invalid value for `artifact`, must not be `None`
```

git-hyagi: hum.... I think serializing artifacts with "null value" was not an issue before because we were allowing it already:
https://github.com/pulp/pulpcore/blob/6fc92a61785dae6522702f2844e00ced901f86e3/pulpcore/app/serializers/fields.py#L31

with the `required=False` we removed the constraint of artifacts being None (not defined!?).

ipanova: my concern is more about that we will be serializing artifact=null and end user won't know whether it is broken/missing or in DB

ipanova: could we instead look into how we can change binding to cope with optional artifact?

pulp_container/app/models.py
git-hyagi: > I tend to go with a) because if space was intentionally reclaimed then it would be a surprise for the user to again see that data occupying the space.

I agree with this.

pulp_container/app/tasks/sync_stages.py
ipanova: I forgot about docker schema1, its' json schema does not have the size field. For this corner case ( and i consider it as a corner because nowadays schema1 is quite rarely used, it is a legacy format) and you can just decide to store all the layers on disk in case of size field absence. This schema1 anyway  does not contain config layer

ipanova:
@lubosmj @git-hyagi  I gathered some more information and insights about the config blob. Firstly containers/image limits the config to 4MB but that number is really unhelpful.
These specs https://github.com/opencontainers/image-spec/commits/main/config.md are being implemented by multiple runtimes, that makes it hard to introduce changes. If having to guess there might be few changes only per a decade.
It boils down to how we want to approach risks and complexity - do we want to have a more complex solution now and deal with simpler/cheaper migrations later (which may never happen), or do we want to have a simpler solution now and pay the price later?
Nowadays complexity consists in the fact that we cannot say in a deterministic way the nature of the blob on upload. As a result we will not have a deterministic knowledge of where certain blob will be stored. 
Heuristics trying to detect on blob upload, based on the first few bytes of the file, JSON (perhaps config) vs. TAR / gzip / zstd/ … (not a config) should be trivially cheap and decisive for *most* blobs but not all ( OCI artifacts could also be JSON)  does not give a lot of confidence. 
Moving around the blob artifact from storage to DB on manifest upload would be precise but with I/O overhead and dealing with the objects relations also does not sound right.

Quay were facing same challenges and decided to store in DB only manifests, blobs outside of DB, while having only labels information in DB.
If we are ok to store few more KB in DB and want to safeguard, we could partially parse config blob on manifest upload and store its' config directive in DB as cached metadata. Or just limit the scope to labels only because we know they are being used now.
We must not forget that we cannot blindly parse any config in the manifest because there are additional config types that probably have different specs and structure https://github.com/pulp/pulp_container/blob/2.19/pulp_container/app/settings.py#L24-L49

Given this information, what do you think we should do best?

git-hyagi:
after checking all the pros and cons, it seems like the current running implementation (not this PR) would be better:
- we learned that there will be no real costs benefit in storing the data into the database instead of object storage
- storing in db based on size has some concerns:
  - adding load to the database (can be an issue for postgres installations shared with other apps, or dbs that are already with heavy load)
  - adding load to the content pods (instead of some blobs being served directly by the object storage, they will be handled by content pods now)
  - defining a "good" DB_BLOB_SIZE to avoid keeping "everything" in the database or keeping "everything" in the storage

benefits on keeping the current implementation:
- maintainability
  - no new logic to check what happened with a blob (if it is a blob - layer or config - it will be stored in the storage)
  - no edge cases for storing the blob in storage (it will always be in the storage, apart from its size or specs and structure)
- no surprises with possible non-mapped new bugs
- no need for new migrations
- no rollback risks

lubosmj:
Great overviews!

I am in favour of storing blobs in the object storage only. As mentioned, unfortunately, the specs do not differ between layers and config blobs when it comes to uploading and storing data. Also, In the Pulp world, we should never change the same content unit between executions as we handle the upload workflow. The content data should be immutable once we put it into the database.

Having the same content type stored "randomly" (based on the threshold set empirically) either in the database or object storage might not seem right. Considering the maintainability aspect, we do not know how config blobs will look in the future and setting the threshold to the wrong value might have devastating effects for us. The most intimidating thing is that it will be really hard to revert the change once we release it.

We already store the relevant labels field in the manifest model. So, this information is already being extracted during the initialization of content units without affecting their internals. Even though, at the cost of reading data from the object storage.

I am now more leaning towards closing this PR and its related issue. 😢

ipanova:
With the additional information provided from quay and container devs  on the config blob,  I am in favor of avoiding the complexity that would be coming with this PR and in favor of keeping the current implementation as is and deal with hypothetical future migrations when they arise.
@git-hyagi Thanks for all the time invested into this and trying to find best solution.

PR #1627, opened 2024-05-17
Add missing dashes to a CLI argument in a doc example

lubosmj:
[noissue]


ipanova (approved):
Looks good to me, if you do not mind please phrase the commit message in a way that it is clear where the changes are made so there is no false impression that there was a change in the code itself. So in this case it will be docs.

PR #1642, opened 2024-05-21
Catch connection errors when fetching data during pull-through caching

### LINKED ISSUE #1499: An error is raised when using pull-through caching in a disconnected environment
This happens when using the pull-through caching workflow with no internet connection.

Traceback:
```
pulp [None]: backoff:ERROR: Giving up download_wrapper(...) after 4 tries (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
[2024-02-03 17:20:47 +0000] [6911] [ERROR] Error handling request
Traceback (most recent call last):
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/connector.py", line 1173, in _create_direct_connection
    hosts = await asyncio.shield(host_resolved)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/connector.py", line 884, in _resolve_host
    addrs = await self._resolver.resolve(host, port, family=self._family)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/resolver.py", line 33, in resolve
    infos = await self._loop.getaddrinfo(
  File "/usr/lib64/python3.8/asyncio/base_events.py", line 825, in getaddrinfo
    return await self.run_in_executor(
  File "/usr/lib64/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib64/python3.8/socket.py", line 918, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
    resp = await request_handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_app.py", line 543, in _handle
    resp = await handler(request)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/web_middlewares.py", line 114, in impl
    return await handler(request)
  File "/src/pulpcore/pulpcore/content/authentication.py", line 48, in authenticate
    return await handler(request)
  File "/src/pulpcore/pulpcore/content/instrumentation.py", line 230, in middleware
    resp = await handler(request)
  File "/src/pulp_container/pulp_container/app/registry.py", line 158, in get_tag
    response = await downloader.run(
  File "/src/pulpcore/pulpcore/download/http.py", line 269, in run
    return await download_wrapper()
  File "/usr/local/lib/python3.8/site-packages/backoff/_async.py", line 151, in retry
    ret = await target(*args, **kwargs)
  File "/src/pulpcore/pulpcore/download/http.py", line 254, in download_wrapper
    return await self._run(extra_data=extra_data)
  File "/src/pulp_container/pulp_container/app/downloaders.py", line 79, in _run
    async with session_http_method(
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/client.py", line 1187, in __aenter__
    self._resp = await self._coro
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/client.py", line 574, in _request
    conn = await self._connector.connect(
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/connector.py", line 544, in connect
    proto = await self._create_connection(req, traces, timeout)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/connector.py", line 911, in _create_connection
    _, proto = await self._create_direct_connection(req, traces, timeout)
  File "/usr/local/lib64/python3.8/site-packages/aiohttp/connector.py", line 1187, in _create_direct_connection
    raise ClientConnectorError(req.connection_key, exc) from exc
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known]
```

lubosmj:
closes #1499

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -12,7 +12,7 @@ import logging
 import hashlib
 import re
 
-from aiohttp.client_exceptions import ClientResponseError
+from aiohttp.client_exceptions import ClientResponseError, ClientConnectionError
 from itertools import chain
 from urllib.parse import urljoin, urlparse, urlunparse, parse_qs, urlencode
 from tempfile import NamedTemporaryFile
@@ -29,6 +29,8 @@ from pulpcore.plugin.models import Artifact, ContentArtifact, UploadChunk
 from pulpcore.plugin.files import PulpTemporaryUploadedFile
 from pulpcore.plugin.tasking import add_and_remove, dispatch
 from pulpcore.plugin.util import get_objects_for_user, get_url
+from pulpcore.plugin.exceptions import TimeoutException
+
 from rest_framework.exceptions import (
     AuthenticationFailed,
     NotAuthenticated,
@@ -55,6 +57,8 @@ from pulp_container.app.cache import (
     RegistryApiCache,
 )
 from pulp_container.app.exceptions import (
+    BadGateway,
+    GatewayTimeout,
     InvalidRequest,
     RepositoryNotFound,
     RepositoryInvalid,
@@ -1117,9 +1121,13 @@ class Manifests(RedirectsMixin, ContainerRegistryApiMixin, ViewSet):
                 # the client could request the manifest outside the docker hub pull limit;
                 # it is necessary to pass this information back to the client
                 raise Throttled()
-            else:
-                # TODO: do not mask out relevant errors, like HTTP 502
+            elif response_error.status == 404:
                 raise ManifestNotFound(reference=pk)
+            else:
+                raise BadGateway(detail=response_error.message)
+        except (ClientConnectionError, TimeoutException):
+            # The remote server is not available at the moment
+            raise GatewayTimeout()
         else:
             digest = response.headers.get("docker-content-digest")
             return models.Manifest.objects.filter(digest=digest).first()
--- a/CHANGES/1499.bugfix
+++ b/CHANGES/1499.bugfix
@@ -1,0 +1,1 @@
+Made the pull-through caching machinery resilient to connection errors.
--- a/pulp_container/app/exceptions.py
+++ b/pulp_container/app/exceptions.py
@@ -1,4 +1,16 @@
-from rest_framework.exceptions import NotFound, ParseError
+from rest_framework.exceptions import APIException, NotFound, ParseError
+
+
+class BadGateway(APIException):
+    status_code = 502
+    default_detail = "Invalid response received from the upstream."
+    default_code = "BAD_GATEWAY"
+
+
+class GatewayTimeout(APIException):
+    status_code = 504
+    default_detail = "Response from the upstream timed out."
+    default_code = "GATEWAY_TIMEOUT"
 
 
 class RepositoryNotFound(NotFound):
--- a/pulp_container/app/registry.py
+++ b/pulp_container/app/registry.py
@@ -8,7 +8,7 @@ from contextlib import suppress
 from urllib.parse import urljoin
 
 from aiohttp import web
-from aiohttp.client_exceptions import ClientResponseError
+from aiohttp.client_exceptions import ClientResponseError, ClientConnectionError
 from aiohttp.web_exceptions import HTTPTooManyRequests
 from django_guid import set_guid
 from django_guid.utils import generate_guid
@@ -21,6 +21,7 @@ from pulpcore.plugin.content import Handler, PathNotResolved
 from pulpcore.plugin.models import RemoteArtifact, Content, ContentArtifact
 from pulpcore.plugin.content import ArtifactResponse
 from pulpcore.plugin.tasking import dispatch
+from pulpcore.plugin.exceptions import TimeoutException
 
 from pulp_container.app.cache import RegistryContentCache
 from pulp_container.app.models import ContainerDistribution, Tag, Blob, Manifest, BlobManifest
@@ -157,7 +158,7 @@ class Registry(Handler):
                 response = await downloader.run(
                     extra_data={"headers": V2_ACCEPT_HEADERS, "http_method": "head"}
                 )
-            except ClientResponseError:
+            except (ClientResponseError, ClientConnectionError, TimeoutException):
                 # the manifest is not available on the remote anymore
                 # but the old one is still stored in the database
                 pass
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 jsonschema>=4.4,<4.23
-pulpcore>=3.43.0,<3.55
+pulpcore>=3.46.0,<3.55
 pyjwt[crypto]>=2.4,<2.9


pulp_container/app/serializers.py
```diff
@@ -308,6 +308,24 @@ class ContainerPullThroughRemoteSerializer(RemoteSerializer):
...
+            "aiohttp.ClientTimeout.connect (q.v.) for download-connections. The default is null, "
+            "which will cause the default from the aiohttp library to be used."
+        ),
+        min_value=0.0,
+    )
+    max_retries = serializers.IntegerField(
+        default=0,
+        help_text=(
+            "Maximum number of retry attempts after a download failure. If not set then the "
+            "default value (3) will be used."
```
ipanova, line 324: the default you have here is 0

pulp_container/app/serializers.py
```diff
@@ -308,6 +308,24 @@ class ContainerPullThroughRemoteSerializer(RemoteSerializer):
 
     policy = serializers.ChoiceField(choices=[Remote.ON_DEMAND], default=Remote.ON_DEMAND)
 
+    connect_timeout = serializers.FloatField(
+        default=2.0,
+        required=False,
+        help_text=(
+            "aiohttp.ClientTimeout.connect (q.v.) for download-connections. The default is null, "
```
ipanova, line 315: the default you're having here is 2

ipanova: while i understand why to decrease the max-retries, I would keep the default timeout for the aiohttp

pulp_container/app/registry_api.py
```diff
@@ -1118,8 +1121,10 @@ def fetch_manifest(self, remote, pk):
                 # it is necessary to pass this information back to the client
                 raise Throttled()
             else:
-                # TODO: do not mask out relevant errors, like HTTP 502
                 raise ManifestNotFound(reference=pk)
+        except (ClientConnectionError, TimeoutException):
+            # The remote server is not available
+            raise ServiceUnavailable()
```
ipanova, line 1127: mmm timeout and 503 are two different things. can't you just re-raise the error?

lubosmj: I cannot re-raise the errors because we are working with two different stacks here: rest_framework and aiohttp.

Let me share my train of thought: the most suitable exception here is `ServiceUnavailable` because when we exceed the timeout or receive the connection error, it means that the pull-through caching service cannot operate properly due to the remote registry unavailability.

pulp_container/app/serializers.py
lubosmj: This would mean that a user will need to wait 15 seconds or more (based on the pulpcore's backoff default) to get an error like: "Connection not available". I thought that it makes sense to cap the max value.

ipanova: I am not sure 2 seconds is enough for the overloaded server to create a new connection. Where did you take the 15s and why you think pulp will backoff? It will backoff only if you have set max_retries which you're setting to 0 here.

pulp_container/app/registry_api.py
lubosmj: I can forward the error back to the user, but I do not know which HTTP code to use apart from 503 (Service unavailable).

> Common causes are a server that is down for maintenance or that is overloaded. This response should be used for temporary conditions and the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) HTTP header should, if possible, contain the estimated time for the recovery of the service.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503

pulp_container/app/serializers.py
lubosmj: When I turn off the internet connection (I cannot reproduce the timeout thingy), I am seeing this in the logs:
```
pulp_1   | pulp [78399887f2ef439eb8a89b4f18b921c3]: django.request:WARNING: Unauthorized: /v2/
pulp_1   | ('pulp [78399887f2ef439eb8a89b4f18b921c3]: ::ffff:127.0.0.1 - - [29/May/2024:12:01:59 +0000] "GET /v2/ HTTP/1.0" 401 106 "-" "containers/5.29.2 (github.com/containers/image)"',)
pulp_1   | Backing off download_wrapper(...) for 0.2s (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | pulp [0ad9a5e4b10d4b08a26bc7700984da47]: backoff:INFO: Backing off download_wrapper(...) for 0.2s (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | Backing off download_wrapper(...) for 0.3s (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | pulp [0ad9a5e4b10d4b08a26bc7700984da47]: backoff:INFO: Backing off download_wrapper(...) for 0.3s (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | Backing off download_wrapper(...) for 2.6s (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | pulp [0ad9a5e4b10d4b08a26bc7700984da47]: backoff:INFO: Backing off download_wrapper(...) for 2.6s (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | Giving up download_wrapper(...) after 4 tries (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | pulp [0ad9a5e4b10d4b08a26bc7700984da47]: backoff:ERROR: Giving up download_wrapper(...) after 4 tries (aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host registry-1.docker.io:443 ssl:default [Name or service not known])
pulp_1   | pulp [0ad9a5e4b10d4b08a26bc7700984da47]: django.request:ERROR: Service Unavailable: /v2/docker-cache/library/busybox/manifests/latest
pulp_1   | ('pulp [0ad9a5e4b10d4b08a26bc7700984da47]: ::ffff:127.0.0.1 - admin [29/May/2024:12:03:23 +0000] "GET /v2/docker-cache/library/busybox/manifests/latest HTTP/1.0" 503 92 "-" "containers/5.29.2 (github.com/containers/image)"',)
```

In total, it takes this much time for podman to figure out the remote server is not available:
```
time podman pull localhost:5001/docker-cache/library/busybox --tls-verify=false

Trying to pull localhost:5001/docker-cache/library/busybox:latest...
Error: initializing source docker://localhost:5001/docker-cache/library/busybox:latest: reading manifest latest in localhost:5001/docker-cache/library/busybox: received unexpected HTTP status: 504 Gateway Time-out

real	1m0.116s
```

I believe the default timeout is set to 300 seconds, see the `ClientTimeout` setting argument: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession. Reducing this value as well as `max_retry` might be useful for user.

lubosmj: With `connect_timeout` set to 2.0, I get this:
```
pulp_1   | ('pulp [ff2655d508f94c90af9128257e68a8a0]: ::ffff:127.0.0.1 - - [29/May/2024:12:14:29 +0000] "GET /v2/ HTTP/1.0" 401 106 "-" "containers/5.29.2 (github.com/containers/image)"',)
pulp_1   | Backing off download_wrapper(...) for 0.7s (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | pulp [a40f99af754d45f1b96ef47240b36684]: backoff:INFO: Backing off download_wrapper(...) for 0.7s (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | Backing off download_wrapper(...) for 1.3s (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | pulp [a40f99af754d45f1b96ef47240b36684]: backoff:INFO: Backing off download_wrapper(...) for 1.3s (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | Backing off download_wrapper(...) for 1.3s (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | pulp [a40f99af754d45f1b96ef47240b36684]: backoff:INFO: Backing off download_wrapper(...) for 1.3s (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | Giving up download_wrapper(...) after 4 tries (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | pulp [a40f99af754d45f1b96ef47240b36684]: backoff:ERROR: Giving up download_wrapper(...) after 4 tries (pulpcore.exceptions.base.TimeoutException: Request timed out for https://registry-1.docker.io/v2/library/busybox/manifests/latest. Increasing the total_timeout value on the remote might help.)
pulp_1   | pulp [a40f99af754d45f1b96ef47240b36684]: django.request:ERROR: Service Unavailable: /v2/docker-cache/library/busybox/manifests/latest
pulp_1   | ('pulp [a40f99af754d45f1b96ef47240b36684]: ::ffff:127.0.0.1 - admin [29/May/2024:12:14:40 +0000] "GET /v2/docker-cache/library/busybox/manifests/latest HTTP/1.0" 503 92 "-" "containers/5.29.2 (github.com/containers/image)"',)
```

```
time podman pull localhost:5001/docker-cache/library/busybox --tls-verify=false

Trying to pull localhost:5001/docker-cache/library/busybox:latest...
Error: initializing source docker://localhost:5001/docker-cache/library/busybox:latest: reading manifest latest in localhost:5001/docker-cache/library/busybox: received unexpected HTTP status: 503 Service Unavailable

real	0m11.751s
```

This is what the user sees with the proposed changes. I think it is much more better.

lubosmj: It appears like lowering the value of `connect_timeout` has a side effect of giving up the connection sooner, leading to faster error times.

lubosmj: Even though `max_retries` is set to 0, I am receiving more than 1 backoff messages. Which is questionable why it is not working in the first place...

ipanova: Because you decreased the value you even get message that says `Increasing the total_timeout value on the remote might help` meaning that it was not enough time to establish the connection( which in this particular case would not work anyway), but you risk with this break even 'normal' cases with internet connections to fail.
I really think you should not  change the timeout connect value or at least leave it 10s ( that seems reasonable enough value but again are you really willing to take the responsibility to set a value which you have not tested how it manifests in connected cases?)

pulp_container/app/registry_api.py
ipanova: is this what you see in logs and podman during pull without your changes? https://github.com/pulp/pulp_container/pull/1642#discussion_r1618773307

lubosmj: Yes, or with connect_timeout set to its default value. But it takes 1 minute to complete. That is why I do not like it. Because it means that Pulp is not responding, correct?

ipanova: The only change that I agree with is this one https://github.com/pulp/pulp_container/pull/1642/files#diff-f8d25b3069358a2d8f46973c532f4818940ad468516ceaea0bbe1ba974c55bceL160

lubosmj: My ultimate goal is to provide a resilient facility for pull-through caching. I can leave all the default values untouched for now. But, we definitely need to catch possible aiohttp exceptions inside api-app to prevent internal server errors. What do you think?

ipanova: can you just raise the APIException with the original error message?

lubosmj: APIException is an HTTP 500 internal server error: https://github.com/encode/django-rest-framework/blob/36d5c0e74f562cbe3055f0d20818bd48d3c32359/rest_framework/exceptions.py#L104.

ipanova: ok, i think as long as you will propagate the original error message it should be ok to use 503. how about you do not change the remote config and let on the user whenever he will be creating/adjusting the remote object.

lubosmj: After thorough consideration, I am leaning towards returning "504 Gateway Timeout" for (ClientConnectionError, TimeoutException). For other exceptions, I am returning 502.

> HTTP 502 Bad Gateway: Used when a server acting as a gateway or proxy receives an invalid response from an upstream server (e.g., invalid response, misconfigured backend, DNS issues).

> HTTP 504 Gateway Timeout: Used when a server acting as a gateway or proxy does not receive a timely response from the upstream server (e.g., slow upstream server, network latency, database query timeout).

pulp_container/app/registry_api.py
```diff
@@ -1117,9 +1121,13 @@ def fetch_manifest(self, remote, pk):
...
                 raise Throttled()
-            else:
-                # TODO: do not mask out relevant errors, like HTTP 502
+            elif response_error.status == 404:
                 raise ManifestNotFound(reference=pk)
+            else:
+                raise BadGateway(detail=response_error.message)
+        except (ClientConnectionError, TimeoutException):
+            # The remote server is not available at the moment
+            raise GatewayTimeout()
```
ipanova, line 1130: can you also pass the error message here in the detail?

ipanova (approved):
one last comment https://github.com/pulp/pulp_container/pull/1642/files#r1622599267

pulp_container/app/registry_api.py
lubosmj: `TimeoutException` does not have an attribute for the error message: https://github.com/pulp/pulpcore/blob/778108af02e75a44d53a66749529e1b0acd751c7/pulpcore/exceptions/base.py#L71.

The same goes for `ClientConnectionError`: https://docs.aiohttp.org/en/stable/_modules/aiohttp/client_exceptions.html#ClientConnectionError.

I would leave this part untouched. Can we agree on that?

PR #1644, opened 2024-05-23
Remove backward compatibility for manifest with artifact

### LINKED ISSUE #1621: Stop supporting manifests with artifacts and enforce the data migration
In one of the upcoming releases (2.22+), we would like to remove the backward compatible code which can handle both artifactless manifests and manifests with artifacts. We want to deal only with artifactless manifests and enforce the migration.

Furthermore, it does make sense to revisit the redirect to content-app/object-storage from api-app because we can return `manifest.data` immediately from api-app.

The relevant code paths are annotated with:
> TODO: BACKWARD COMPATIBILITY - remove after fully migrating to artifactless manifest"

MichalPysik:
This commit removes support for manifests storing their data inside an artifact instead of using the recently introduced  Manifest.data text field.

closes #1621

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -1084,7 +1084,7 @@ class Manifests(RedirectsMixin, ContainerRegistryApiMixin, ViewSet):
                 else:
                     raise ManifestNotFound(reference=pk)
 
-            return redirects.issue_manifest_redirect(manifest)
+            return Response(manifest.data)
 
     def get_content_units_to_add(self, manifest, tag):
         add_content_units = [str(tag.pk), str(manifest.pk)]
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -34,7 +34,6 @@ from pulp_container.app.utils import (
     determine_media_type,
     validate_manifest,
     calculate_digest,
-    get_content_data,
 )
 
 log = logging.getLogger(__name__)
@@ -77,25 +76,9 @@ class ContainerFirstStage(Stage):
 
         digest = response.headers.get("docker-content-digest")
 
-        if (
-            manifest := await Manifest.objects.prefetch_related("contentartifact_set")
-            .filter(digest=digest)
-            .afirst()
-        ):
-            if raw_text_data := manifest.data:
-                content_data = json.loads(raw_text_data)
-
-            # TODO: BACKWARD COMPATIBILITY - remove after fully migrating to artifactless manifest
-            elif saved_artifact := await manifest._artifacts.afirst():
-                content_data, raw_bytes_data = await sync_to_async(get_content_data)(saved_artifact)
-                raw_text_data = raw_bytes_data.decode("utf-8")
-            # if artifact is not available (due to reclaim space) we will download it again
-            else:
-                content_data, raw_text_data, response = await self._download_manifest_data(
-                    response.url
-                )
-            # END OF BACKWARD COMPATIBILITY
-
+        if manifest := await Manifest.objects.filter(digest=digest).afirst():
+            raw_text_data = manifest.data
+            content_data = json.loads(raw_text_data)
         else:
             content_data, raw_text_data, response = await self._download_manifest_data(response.url)
 
@@ -333,9 +316,7 @@ class ContainerFirstStage(Stage):
         while True:
             link = urljoin(self.remote.url, rel_link)
             list_downloader = self.remote.get_downloader(url=link)
-            # FIXME this can be rolledback after https://github.com/pulp/pulp_container/issues/1288
-            # tags/list endpoint does not like any unnecessary headers to be sent
-            await list_downloader.run(extra_data={"repo_name": repo_name, "headers": {}})
+            await list_downloader.run(extra_data={"repo_name": repo_name})
             with open(list_downloader.path) as tags_raw:
                 tags_dict = json.loads(tags_raw.read())
                 tag_list.extend(tags_dict["tags"])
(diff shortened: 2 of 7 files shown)


pulp_container/app/tasks/sync_stages.py
```diff
@@ -477,15 +468,8 @@ async def create_listed_manifest(self, manifest_data):
...
-            # TODO: BACKWARD COMPATIBILITY - remove after fully migrating to artifactless manifest
-            elif saved_artifact := await manifest._artifacts.afirst():
-                content_data, _ = await sync_to_async(get_content_data)(saved_artifact)
-            # if artifact is not available (due to reclaim space) we will download it again
             else:
-                content_data, manifest = await self._download_and_instantiate_manifest(
-                    manifest_url, digest
-                )
-            # END OF BACKWARD COMPATIBILITY
+                pass  # TODO: possible exception?
```
lubosmj, line 472: We can go directly for `manifest.data` without any checks. What do you think?

lubosmj:
Have you considered returning `manifest.data` directly from api-app? We may want to omit redirects to content-app when the data is available.

lubosmj:
I think tests are failing because of the caching machinery:
```
2024-06-03T10:51:37.9321965Z   File "/usr/local/lib/python3.9/site-packages/pulp_container/app/registry_api.py", line 1016, in get
2024-06-03T10:51:37.9322259Z     return self.handle_safe_method(request, path, pk)
2024-06-03T10:51:37.9323128Z   File "/usr/local/lib/python3.9/site-packages/pulpcore/cache/cache.py", line 171, in cached_function
2024-06-03T10:51:37.9323626Z     response = self.make_entry(key, bk, func, args, kwargs, self.default_expires_ttl)
2024-06-03T10:51:37.9324448Z   File "/usr/local/lib/python3.9/site-packages/pulpcore/cache/cache.py", line 220, in make_entry
2024-06-03T10:51:37.9324840Z     entry["content"] = response.content.decode("utf-8")
2024-06-03T10:51:37.9325657Z   File "/usr/local/lib/python3.9/site-packages/django/template/response.py", line 135, in content
2024-06-03T10:51:37.9325900Z     raise ContentNotRenderedError(
2024-06-03T10:51:37.9326730Z django.template.response.ContentNotRenderedError: The response content must be rendered before it can be accessed.
```

We started returning `manifest.data` from api-app which results in the said error. We may want to update the code on the pulpcore's side or overwrite the existing behaviour, where `decode` is always called on the TextField data, from the plugin itself.

```
# cache.py

     def make_entry():
          ...
          elif isinstance(response, HttpResponse):
            entry["type"] = "Response"
            if isinstance(response.data, str):
               # TODO: store response.content data
            else:
                entry["content"] = response.content.decode("utf-8")
        else:
            # We don't cache StreamResponses or errors
            return response
```

MichalPysik:
> I think tests are failing because of the caching machinery:
> 
> ```
> 2024-06-03T10:51:37.9321965Z   File "/usr/local/lib/python3.9/site-packages/pulp_container/app/registry_api.py", line 1016, in get
> 2024-06-03T10:51:37.9322259Z     return self.handle_safe_method(request, path, pk)
> 2024-06-03T10:51:37.9323128Z   File "/usr/local/lib/python3.9/site-packages/pulpcore/cache/cache.py", line 171, in cached_function
> 2024-06-03T10:51:37.9323626Z     response = self.make_entry(key, bk, func, args, kwargs, self.default_expires_ttl)
> 2024-06-03T10:51:37.9324448Z   File "/usr/local/lib/python3.9/site-packages/pulpcore/cache/cache.py", line 220, in make_entry
> 2024-06-03T10:51:37.9324840Z     entry["content"] = response.content.decode("utf-8")
> 2024-06-03T10:51:37.9325657Z   File "/usr/local/lib/python3.9/site-packages/django/template/response.py", line 135, in content
> 2024-06-03T10:51:37.9325900Z     raise ContentNotRenderedError(
> 2024-06-03T10:51:37.9326730Z django.template.response.ContentNotRenderedError: The response content must be rendered before it can be accessed.
> ```
> 
> We started returning `manifest.data` from api-app which results in the said error. We may want to update the code on the pulpcore's side or overwrite the existing behaviour, where `decode` is always called on the TextField data, from the plugin itself.
> 
> ```
> # cache.py
> 
>      def make_entry():
>           ...
>           elif isinstance(response, HttpResponse):
>             entry["type"] = "Response"
>             if isinstance(response.data, str):
>                # TODO: store response.content data
>             else:
>                 entry["content"] = response.content.decode("utf-8")
>         else:
>             # We don't cache StreamResponses or errors
>             return response
> ```

@lubosmj If I understand correctly, before we started returning ```manifest.data``` (which is text, not bytes), we used to return it in bytes (abstracting away the fact that it went through multiple additional functions to get to that point)? And if that is the case, would simply returning ```manifest.data.encode()``` fix the issue?

lubosmj:
Oh, actually, the issue is that we cannot access `response.content`. Thus, the execution never reaches the `decode` call. Can you verify if calling "response.render()" inside the pulpcore's code will resolve the issue? Other than that, I think we should update the caching code as a whole to deal with `HttpResponse` properly. It is plausible that [this code path](https://github.com/pulp/pulpcore/blob/02bb455315a17a39cb0d936a12a13fbbdb887bb3/pulpcore/cache/cache.py#L220) was never executed. Or, we may think about returning something else from api-app...

pulp_container/app/registry_api.py
```diff
@@ -1084,7 +1084,7 @@ def handle_safe_method(self, request, path, pk):
                 else:
                     raise ManifestNotFound(reference=pk)
 
-            return redirects.issue_manifest_redirect(manifest)
+            return Response(manifest.data)
```
lubosmj, line 1087: Bear in mind that we would like to return headers as well, similarly to `return web.Response(text=tag.tagged_manifest.data, headers=response_headers)`.

MichalPysik: @lubosmj What headers would you like to return in this case?

lubosmj:
The work on this can be resumed now.

Issue #1213, opened 2023-02-16
Container Distribution API should report if the registry is using HTTP or HTTPS


dkliban:
Currently the Container Distribution provides a `registry_path` and a `base_path`. It would be nice if it provided an additional field boolean field called `secure`. Another option would a `scheme` field that would be either `https` or `http`. 
```
  {
    "content_guard": "/pulp/api/v3/contentguards/core/content_redirect/a9e3803d-f5c1-4528-824d-7341f8b85861/",
    "pulp_href": "/pulp/api/v3/distributions/container/container/9e240245-df6c-4067-8d00-04eb0b10c5b9/",
    "pulp_labels": {},
    "base_path": "foo/bar",
    "pulp_created": "2023-02-16T16:45:49.954541Z",
    "repository": "/pulp/api/v3/repositories/container/container/4292bfe3-1404-4283-b094-7cc37a2601de/",
    "name": "foo",
    "repository_version": null,
    "registry_path": "localhost:5001/foo/bar",
    "namespace": "/pulp/api/v3/pulp_container/namespaces/7f673a7f-dfe4-4547-9c14-8f483a67bfd5/",
    "private": false,
    "description": null
  }
```

lubosmj:
Can this information be extracted from `pulp status` instead?

```
  "content_settings": {
    "content_origin": "http://localhost:5001",
    "content_path_prefix": "/pulp/content/"
  },
```

Administrators should know whether we serve content over HTTP or HTTPS. Do users read distributions from the API to identify the need for `--tls-verfiy=true` in podman calls?

ipanova:
If i recall correctly, that was a request incoming for the replication workflow, where the registry_path is going to be used in the remote url that requires a scheme to be provided.

lubosmj:
But, generally, you do not specify the scheme when running `podman pull` or `podman push`. The `registry_path` field is the entry that tells users where to pull from or push to. Additional details can be extracted from the context.

ipanova:
I did not suggest to add the scheme to the registry_path. I just provided the context for this request, which needs to know the scheme  or to know where to retrieve the scheme in order to create a valid remote.url

PR #1645, opened 2024-05-27
Distribution API reports whether it uses HTTP/S

MichalPysik:
Distribution now reports whether the registry is using HTTP (unsecure) or HTTPS (secure) via the new 'secure' field inside ContainerDistributionSerializer.

closes: #1213

### DIFF
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -351,6 +351,13 @@ class ContainerDistributionSerializer(DistributionSerializer, GetOrCreateSeriali
         view_name_pattern=r"remotes(-.*/.*)?-detail",
         read_only=True,
     )
+    secure = serializers.SerializerMethodField(
+        help_text="Whether the distribution is served over HTTP (insecure) or HTTPS (secure)."
+    )
+
+    def get_secure(self, obj):
+        request = self.context["request"]
+        return request.is_secure()
 
     def validate(self, data):
         """
@@ -406,6 +413,7 @@ class ContainerDistributionSerializer(DistributionSerializer, GetOrCreateSeriali
             "namespace",
             "private",
             "description",
+            "secure",
         )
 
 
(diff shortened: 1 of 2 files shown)


pulp_container/app/serializers.py
```diff
@@ -351,6 +351,13 @@ class ContainerDistributionSerializer(DistributionSerializer, GetOrCreateSeriali
         view_name_pattern=r"remotes(-.*/.*)?-detail",
         read_only=True,
     )
+    secure = serializers.SerializerMethodField(
+        help_text="Whether the distribution is served over HTTP (unsecure) or HTTPS (secure)."
```
ipanova, line 355: s/unsecure/insecure

ipanova:
@MichalPysik @lubosmj  While this fix is trivial, I understand the questioning of its value. The original request came from the need of replication feature. In order to replicate distributions which in turn become remotes one needs to know  the scheme. My suggestion is to try to first add replication feature to container plugin to see whether we can extract this information ( which will be all same across pulp) from elsewhere or we indeed need an additional field on distribution.
Having that said, @MichalPysik would you be willing to look into replication feature?

lubosmj:
@dkliban, is this ^ understanding correct? Why would you isolate the request for adding replicas capabilities and displaying the schema value?

pulp_container/app/serializers.py
MichalPysik: @ipanova They're synonyms and unsecure is more commonly used for non-living objects while insecure is more common in emotional context, but I'll change it to insecure if you prefer it.

ipanova: Not going to argue on this because I am not a native speaker but it really feels weird when reading it.
Results from google say it all:
`unsecure network` ---> About 852,000 results (0.22 seconds) 
`insecure network` --> About 32,900,000 results (0.25 seconds)

MichalPysik:
Closing this PR as its content was moved to https://github.com/pulp/pulp_container/issues/1648.

Issue #1602, opened 2024-04-24
Tasks created after uploading manifests should not be deleted


lubosmj:
In some cases, we do remove `add_and_remove` tasks invoked by the push workflow. We would want to keep **all** tasks related to the uploading. Admins may want to run the task purge workflow frequently.

### Additional references
https://github.com/pulp/pulp_container/blob/39252e4e56835dadc0d4d41c37cd83682181f758/pulp_container/app/registry_api.py#L1284
https://github.com/pulp/pulp_container/blob/39252e4e56835dadc0d4d41c37cd83682181f758/pulp_container/app/utils.py#L139

PR #1647, opened 2024-05-27
Stop deleting tasks after uploading manifests

MichalPysik:
Tasks created after uploading manifests are no longer automatically
deleted in any case.

closes: #1602

### DIFF
--- a/CHANGES/1602.bugfix
+++ b/CHANGES/1602.bugfix
@@ -1,0 +1,1 @@
+Tasks created after uploading manifests will now remain available for further inspection and will not be deleted.
--- a/pulp_container/app/utils.py
+++ b/pulp_container/app/utils.py
@@ -136,14 +136,11 @@ def has_task_completed(dispatched_task, wait_in_seconds=3):
         time.sleep(1)
         task = Task.objects.get(pk=dispatched_task.pk)
         if task.state == "completed":
-            task.delete()
             return True
         elif task.state in ["waiting", "running"]:
             continue
         else:
-            error = task.error
-            task.delete()
-            raise Exception(str(error))
+            raise Exception(str(task.error))
     raise Throttled()
 
 


CHANGES/1602.bugfix
```diff
@@ -0,0 +1 @@
+Tasks created after uploading manifests are now automatically deleted once they're not needed.
```
lubosmj, line 1:
```suggestion
Tasks created after uploading manifests are now automatically deleted once they are not needed.
```

ipanova:
This requests was valid when we were creating mutliple tasks for one image push to not clutter things, however now, if my memory does not lie to me, we create just one final task that would contain blob upload, manifest/tag upload. When operation fails it is good to have access to failed tasks to debug them. Having that said, should we actually *not* delete the task?:)

lubosmj:
I am in favour of deleting them because this task is not transparent to an end user. We can store the tasks that failed/errored (where "raise Exception" is used) so admins can inspect them later.

ipanova:
what do you mean not transparent to end user? it is same task as sync task.

lubosmj:
They are created in the background and they cannot be easily tracked.

ipanova:
It is true that the href of task is not returned to the user in the same fashion as it would be done with e.g. sync task, but  the task still can be found in the task list api. Given that we no longer produce X tasks per blob upload and have just one, I really think it would not hurt to keep that single task around. If things do not work out during push operation there is not really any forensic evidence left on pulp side that can help with debugging.

lubosmj:
😞 @MichalPysik, can you ensure not a single task is deleted from the database? So, instead of deleting them all, we want to stop deleting them in the existing places.

ipanova:
@lubosmj  don't get me wrong, I am not demanding this, just sharing my thinking. We could still delete them if this is what the team think it's best ;)

lubosmj:
I mean, we can leave the tasks untouched (not delete them at all). Admins should run the task purge more frequently to not bloat their system.

ipanova:
great, I agree :)

MichalPysik:
@lubosmj @ipanova I'm a bit confused, what's the conclusion? Do we want to stop deleting the tasks that used to get deleted (before this PR), so that this PR basically does the opposite of what it was supposed to do initially?

lubosmj:
Yes.

lubosmj (approved):

ipanova (approved):

PR #1649, opened 2024-05-28
Disable token authentication for the S3 runner

### LINKED ISSUE #1607: Introduce a new test runner that has the token authentication disabled
Some of our users do not use RBAC and are purely relying on the admin user or anonymous tokens to pull images from Pulp or push images to Pulp.

We should have a test scenario that runs all the tests with the `TOKEN_AUTH_DISABLED=True` setting.

lubosmj:
This change might help us identify errors in scenarios where the token authentication is disabled (e.g., Katello installations).

closes #1607

### DIFF
--- a/pulp_container/tests/functional/api/test_push_content.py
+++ b/pulp_container/tests/functional/api/test_push_content.py
@@ -5,6 +5,8 @@ import pytest
 import requests
 import unittest
 
+from django.conf import settings
+
 from subprocess import CalledProcessError
 from urllib.parse import urlparse, urljoin
 
@@ -71,7 +73,6 @@ def test_push_without_login(
 def test_push_with_dist_perms(
     add_to_cleanup,
     gen_user,
-    anonymous_user,
     registry_client,
     local_registry,
     container_push_repository_api,
@@ -83,6 +84,9 @@ def test_push_with_dist_perms(
 
     It also checks read abilities for users with different set of permissions.
     """
+    if settings.TOKEN_AUTH_DISABLED:
+        pytest.skip("RBAC cannot be tested when token authentication is disabled")
+
     user_creator = gen_user(model_roles=["container.containernamespace_creator"])
     user_dist_collaborator = gen_user()
     user_dist_consumer = gen_user()
@@ -164,12 +168,12 @@ def test_push_with_no_perms(
     gen_user,
     registry_client,
     local_registry,
+    container_distribution_api,
     container_namespace_api,
 ):
     """
     Test that user with no permissions can't perform push.
     """
-    user_creator = gen_user(model_roles=["container.containernamespace_creator"])
     user_helpless = gen_user()
     repo_name = "unsuccessful/perms"
     local_url = f"{repo_name}:2.0"
@@ -178,16 +182,35 @@ def test_push_with_no_perms(
     with user_helpless, pytest.raises(CalledProcessError):
         local_registry.tag_and_push(image_path, local_url)
 
-    # test a user can still pull
-    with user_creator:
+    # test if the helpless user can still pull
+    if settings.TOKEN_AUTH_DISABLED:
+        # push by using the admin user
         local_registry.tag_and_push(image_path, local_url)
         namespace = container_namespace_api.list(name="unsuccessful").results[0]
         add_to_cleanup(container_namespace_api, namespace.pulp_href)
 
-    with user_helpless:
-        with pytest.raises(CalledProcessError):
+        with user_helpless:
+            with pytest.raises(CalledProcessError):
+                local_registry.tag_and_push(image_path, local_url)
+            local_registry.pull(local_url)
+
+        # flagging the repository as "private" does not have an effect on pulling
+        distribution = container_distribution_api.list(name=repo_name).results[0]
+        container_distribution_api.partial_update(distribution.pulp_href, {"private": True})
+        with user_helpless:
+            local_registry.pull(local_url)
+    else:
+        # push by using the creator user
+        user_creator = gen_user(model_roles=["container.containernamespace_creator"])
+        with user_creator:
             local_registry.tag_and_push(image_path, local_url)
-        local_registry.pull(local_url)
+            namespace = container_namespace_api.list(name="unsuccessful").results[0]
+            add_to_cleanup(container_namespace_api, namespace.pulp_href)
+
+        with user_helpless:
+            with pytest.raises(CalledProcessError):
+                local_registry.tag_and_push(image_path, local_url)
+            local_registry.pull(local_url)
 
 
 def test_push_to_existing_namespace(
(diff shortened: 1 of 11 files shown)


pulp_container/tests/functional/api/test_repositories_list.py
```diff
@@ -120,6 +128,7 @@ def test_listing_repositories(self):
         self.assertEqual(repositories.json(), {"repositories": repositories_names})
 
 
+@unittest.skipIf(TOKEN_AUTH_DISABLED, "Token authentication is not enabled")
```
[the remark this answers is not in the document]
ipanova: hmm..to summarize our logic.. So when we have token_auth enabled only public repos are visible to users with 0 perms or anon user. But when we have token_auth disabled we expose everything because there is no concept of public or private repo.

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -169,6 +173,9 @@ def test_push_with_no_perms(
     """
     Test that user with no permissions can't perform push.
```
ipanova, line 174: maybe we can still run it even with token_auth disabled, with user_helpless

ipanova: i.e check with user helpless that he cannot tag_and push, check with user helpless that he can still pull repo whether it's private or public

ipanova (changes requested):
left some comments, thanks!

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -273,6 +292,9 @@ def test_push_private_repository(
     Test that the same user can pull, but another cannot.
     Test that the other user can pull after marking it non-private.
     """
+    if settings.TOKEN_AUTH_DISABLED:
```
ipanova, line 295: what about testing helpless_user pull of private/public repo?

pulp_container/tests/functional/api/test_repositories_list.py
```diff
@@ -186,6 +194,7 @@ def test_none_user(self):
         repositories = self.get_listed_repositories(auth)
```
ipanova, line 194: I am a bit confused, should not user_none with toke_auth disabled see all 3 repos? (2 private and 1 public)

lubosmj: Not because of this: https://github.com/pulp/pulp_container/blob/6d4b4f4912ec8542e90d139c83be551b74215c65/pulp_container/app/registry_api.py#L518

pulp_container/tests/functional/api/test_push_content.py
lubosmj: I extended the `test_push_with_no_perms` test case to include pulling of a private repository.

pulp_container/tests/functional/api/test_repositories_list.py
ipanova: aha, but then don't we state actually in our docs that the concept of private and public repo does not work if you disabled token auth?

lubosmj: yes, but is not the interpretation of it like: admin can do everything, others can do nothing apart from pulling?

pulp_container/tests/functional/api/test_push_content.py
```diff
@@ -178,15 +182,34 @@ def test_push_with_no_perms(
...
+
+        with user_helpless:
+            with pytest.raises(CalledProcessError):
+                local_registry.tag_and_push(image_path, local_url)
+            local_registry.pull(local_url)
+
+    # flagging the repository as a private causes the helpless user to fail the pull now
+    distribution = container_distribution_api.list(name=repo_name).results[0]
+    container_distribution_api.partial_update(distribution.pulp_href, {"private": True})
+    with user_helpless, pytest.raises(CalledProcessError):
```
lubosmj, line 212: @ipanova ^

pulp_container/tests/functional/api/test_repositories_list.py
ipanova: seems like we have a discrepancy 
```
$ podman logout puffy.example.com
Error: not logged into puffy.example.com
$ http https://puffy.example.com/v2/_catalog  --ignore-netrc
HTTP/1.1 200 OK
Allow: GET, HEAD, OPTIONS
Connection: keep-alive
Content-Length: 49
Correlation-ID: 08fb832d01cd4df5945e3ad88ba740c8
Cross-Origin-Opener-Policy: same-origin
Date: Wed, 29 May 2024 15:20:44 GMT
Docker-Distribution-Api-Version: registry/2.0
Referrer-Policy: same-origin
Server: nginx
Strict-Transport-Security: max-age=15768000
Vary: Accept
X-Frame-Options: DENY
X-Registry-Supports-Signatures: 1

{
    "repositories": [
        "docker-cache/library/busybox"
    ]
}


$ podman pull puffy.example.com/rhosp-rhel8/openstack-cron:16.2
Trying to pull puffy.example.com/rhosp-rhel8/openstack-cron:16.2...
Copying blob 2b814fe20317 skipped: already exists  
Copying blob b243f36441d6 skipped: already exists  
Copying blob 6f474dfa2585 skipped: already exists  
Copying config 6cb8a7d9e4 done  
Storing signatures
6cb8a7d9e4333a95f5f6cd8a34d40b10ada4d4362096573f6e5e751b8ed9baba
$ pulp container distribution update --name 'rhosp-rhel8/openstack-cron' --public
Started background task /pulp/api/v3/tasks/018fc4f1-00ff-7f0a-8bce-0c7b5ab20409/
Done.
$ http https://puffy.example.com/v2/_catalog  --ignore-netrc
HTTP/1.1 200 OK
Allow: GET, HEAD, OPTIONS
Connection: keep-alive
Content-Length: 78
Correlation-ID: 3c6d43c6fa7c4f8d82c46be927682e30
Cross-Origin-Opener-Policy: same-origin
Date: Wed, 29 May 2024 15:21:24 GMT
Docker-Distribution-Api-Version: registry/2.0
Referrer-Policy: same-origin
Server: nginx
Strict-Transport-Security: max-age=15768000
Vary: Accept
X-Frame-Options: DENY
X-Registry-Supports-Signatures: 1

{
    "repositories": [
        "docker-cache/library/busybox",
        "rhosp-rhel8/openstack-cron"
    ]
}




```

pulp_container/tests/functional/api/test_push_content.py
ipanova: where?

lubosmj: I removed it after you opened the new issue: https://github.com/pulp/pulp_container/pull/1649#discussion_r1619075596.

ipanova: the pull of private repo will pass, so you can add it back. it's the catalog list of private repos that will fail. or you decided not to add the test unless we figure out whether it is ok for anon user to be able to pull private repos with token auth disabled?

lubosmj: I was still thinking whether it is right to completely ignore the private flag even though the token authentication is disabled. I have just added back the change. I am not catching an exception anymore, just verifying if using the private flag does not have any impact on pulling.

ipanova (approved):

PR #1660, opened 2024-06-12
Sign ignores .sig, .att, .sbom

### LINKED ISSUE #1347: Signing task should ignore cosign produced signatures, attestations and sboms
**Version**
all

**Describe the bug**
Atomic signature signing task signs images. Cosign signatures, attestations and sboms are presented as an image too. We should ignore them and not sign. 

**To Reproduce**
Trigger atomic signature pulp signing task ( skopeo signing)

**Expected behavior**
it does not make sense to sign cosign produced signatures, attestations and sboms

**Additional context**
We can filter out tags that end in  `.sig, .att, .sbom`  https://github.com/pulp/pulp_container/blob/main/pulp_container/app/tasks/sign.py#L50

An sbom *can* be signed, but I don't know how much it makes sense to sign it with skopeo?

MichalPysik:
The signing tasks no longer signs cosign signatures, attestations and sboms (images that end with .sigg, .att, or .sbom) and ignores them instead.

closes #1347

### DIFF
--- a/CHANGES/1347.bugfix
+++ b/CHANGES/1347.bugfix
@@ -1,0 +1,2 @@
+The pulp signing task that produces atomic type signature no longer signs cosign signatures,
+attestations and sboms (images that end with .sig, .att, or .sbom), and ignores them instead.
--- a/pulp_container/app/tasks/sign.py
+++ b/pulp_container/app/tasks/sign.py
@@ -5,6 +5,7 @@ import hashlib
 from aiofiles import tempfile
 from asgiref.sync import sync_to_async
 from django.conf import settings
+from django.db.models import Q
 
 from pulpcore.plugin.models import Repository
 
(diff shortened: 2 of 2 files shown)


pulp_container/app/tasks/sign.py
```diff
@@ -46,10 +47,21 @@ def sign(repository_pk, signing_service_pk, reference, tags_list=None):
     latest_version = repository.latest_version()
     if tags_list:
         latest_repo_content_tags = latest_version.content.filter(
-            pulp_type=Tag.get_pulp_type(), pk__in=tags_list
+            pulp_type=Tag.get_pulp_type(),
+            pk__in=tags_list,
+        ).exclude(
+            Q(container_tag__endswith=".sig")
+            | Q(container_tag__endswith=".att")
+            | Q(container_tag__endswith=".sbom")
```
lubosmj, line 55: I wonder if there are any more types we would like to ignore. I could not find any while briefly googling the subject.

CHANGES/1347.bugfix
```diff
@@ -0,0 +1,3 @@
+The signing tasks no longer signs cosign signatures, attestations and
```
ipanova, line 1: s/tasks/task

pulp_container/app/tasks/sign.py
ipanova: instead of traversing into another table do it directly from the Tag.name on line 65

ipanova (changes requested):

lubosmj (approved):
The change addresses the problem. I did not find any other suffices that we should consider during the evaluation.

CHANGES/1347.bugfix
```diff
@@ -0,0 +1,3 @@
+The signing task no longer signs cosign signatures, attestations and
+sboms (images that end with .sigg, .att, or .sbom), and ignores them
```
ipanova, line 2: s/.sigg/sig

CHANGES/1347.bugfix
```diff
@@ -0,0 +1,3 @@
+The signing task no longer signs cosign signatures, attestations and
```
ipanova, line 1: I just realized maybe providing a bit more info might be useful. How about: The pulp signing task that produces atomic type signature no longer....

ipanova (approved):

PR #1664, opened 2024-06-16
Update CI files for branch 2.14

pulpbot:


pulp_container/tests/functional/api/test_rbac_repo_versions.py
```diff
@@ -286,9 +283,12 @@ def test_cross_repository_blob_mount(
         content_response, auth = mount_blob(blobs[0], source_repo, f"{dest_tester_namespace}/test")
         assert content_response.status_code == 401
 
+    with suppress(Exception):
+        monitor_task(container_namespace_api.delete(distribution.namespace).task)
```
lubosmj, line 287: Inspired by https://github.com/pulp/pulpcore/blob/5721e58d94e15c56be30ab4a5b5db46bb74e746f/pulpcore/pytest_plugin.py#L901.

Previous implementation was not resilient to task failures, see: https://github.com/pulp/pulpcore/blob/3.24/pulpcore/tests/functional/__init__.py#L933. We no longer support 3.24 branch even though it is still being installed by the CI. 😞 Therefore, I decided to include the change from newer pulpcore versions in the plugin itself.

lubosmj (approved):

Issue #1651, opened 2024-05-29
Disrepancy beetween what a user can pull and see in catalog with token_auth off


ipanova:
**Version**
Please provide the versions of the pulpcore and pulp_container packages in use, and how they are installed.  If you are using Pulp via Katello, please provide the Katello version.

**Describe the bug**

**To Reproduce**
Disabled token_auth
Create a mix of public and privates repos. Notice that non-authed user can pull privately marked repo but does not see it in catalog.

**Expected behavior**
With token-auth disabled, non-authed user should see all repos in catalog (private and public) and pull also all of them

**Additional context**
https://github.com/pulp/pulp_container/pull/1649#discussion_r1619084577

ipanova:
We need to add a test when token auth is disabled, non-authed user should be able to see public and private repos in the catalog

PR #1668, opened 2024-06-18
Return all catalog repos when token-auth disabled

git-hyagi:
With token-auth disabled, allow non-authed users to see all repos (public and private) in catalog.

closes: #1651

### DIFF
--- a/CHANGES/1651.bugfix
+++ b/CHANGES/1651.bugfix
@@ -1,0 +1,2 @@
+Modified the `_catalog` endpoint to allow non-authed users to see all repos in catalog
+(private and public) when token-auth is disabled.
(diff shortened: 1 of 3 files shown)


CHANGES/1651.bugfix
```diff
@@ -0,0 +1,2 @@
+Modified the `_catalog` endpoint to allow non-authed users to see all repos in catalog
+(private and public) even when token-auth is disabled.
```
ipanova, line 2: s/even//

pulp_container/app/registry_api.py
```diff
@@ -519,6 +519,9 @@ def get_queryset(self, *args, **kwargs):
         distribution_permission = "container.pull_containerdistribution"
         namespace_permission = "container.namespace_pull_containerdistribution"
 
+        if settings.get("TOKEN_AUTH_DISABLED", True):
```
ipanova, line 522: This is dangerous because if the setting is not defined then it will se set to True

Do instead `if settings.get("TOKEN_AUTH_DISABLED", False)`

ipanova (changes requested):

ipanova (approved):

PR #1677, opened 2024-06-25
Add includes/excludes fields to filter remotes

git-hyagi:
closes: #459


pulp_container/app/tasks/sync_stages.py
```diff
@@ -115,10 +117,13 @@ async def run(self):
         async with ProgressReport(
             message="Downloading tag list", code="sync.downloading.tag_list", total=1
         ) as pb:
-            repo_name = self.remote.namespaced_upstream_name
+            try:
+                repo_name = filter_repo(self.remote)
+            except InvalidRequest:
```
lubosmj, line 122: A side note: this is a `rest_framework` exception that is supposed to be used within the api-app stack. It helps us to better render error messages. It is not very helpful to "raise" an HTTP 400 error from a task running inside a worker because there is no HTTP interface involved and a user cannot benefit from it.

git-hyagi:
> Can you describe a workflow for this use case? Do you think a user would want to execute syncing of a remote repository while at the same time excluding the repository she wants to sync from?

My bad! I thought that we could sync/mirror an entire namespace from a remote upstream, but testing it again, I realized that it is not possible, which invalidates the use of `includes`/`excludes` in syncs.

lubosmj (changes requested):

pulp_container/app/models.py
```diff
@@ -355,13 +355,17 @@ class ContainerRemote(Remote, AutoAddObjPermsMixin):
...
+        excludes (models.JSONField): Dictionary of repositories to exclude. [default=null]
     """
 
     upstream_name = models.TextField(db_index=True)
     include_foreign_layers = models.BooleanField(default=False)
     include_tags = fields.ArrayField(models.TextField(null=True), null=True)
     exclude_tags = fields.ArrayField(models.TextField(null=True), null=True)
     sigstore = models.TextField(null=True)
+    includes = models.JSONField(null=True)
+    excludes = models.JSONField(null=True)
```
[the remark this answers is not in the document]
git-hyagi: hum... I had left it there because the `get_or_create` from `get_pull_through_drv` was expecting it (if not provided, it was raising a `FieldError` exception):
https://github.com/pulp/pulp_container/blob/b126f4ca621e0559a00bf23e2d5e85c52ed65e44/pulp_container/app/registry_api.py#L318-L324

but, now that you mentioned it, I guess we can remove it and just delete the `includes`/`excludes` keys from `remote_data` dictionary before calling the `get_or_create`.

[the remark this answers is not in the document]
lubosmj: Interesting. Are not these two models completely isolated?

pulp_container/app/registry_api.py
```diff
@@ -1110,9 +1111,11 @@ def get_content_units_to_add(self, manifest, tag):
         return add_content_units
 
     def fetch_manifest(self, remote, pk):
-        relative_url = "/v2/{name}/manifests/{pk}".format(
-            name=remote.namespaced_upstream_name, pk=pk
-        )
+        try:
+            repo_name = filter_repo(remote)
+        except RepositoryNotFound:
+            raise
```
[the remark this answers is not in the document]
git-hyagi: > Do you think it makes more sense to move the check to a different place?

Oh!! Nice!
At first, I thought that putting it there would be good because we would be *failing early* (before downloading the manifest), but I didn't think about the previous resources that had been created.

pulp_container/app/models.py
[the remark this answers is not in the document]
git-hyagi: hum... kind of!?

from my understanding, when we create the `ContainerRemote` https://github.com/pulp/pulp_container/blob/b126f4ca621e0559a00bf23e2d5e85c52ed65e44/pulp_container/app/registry_api.py#L320-L325 we are passing the remote values (`remote_data`) extracted from the `ContainerPullThroughRemote` https://github.com/pulp/pulp_container/blob/ba2e297b8414e5f52474c35dcc8a5cada24a204e/pulp_container/app/registry_api.py#L319
https://github.com/pulp/pulp_container/blob/ba2e297b8414e5f52474c35dcc8a5cada24a204e/pulp_container/app/registry_api.py#L393-L399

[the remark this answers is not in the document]
lubosmj: Aha, I see. Then, we may want to ignore these fields. Thus, we can alter `IGNORED_PULL_THROUGH_REMOTE_ATTRIBUTES` to suit our needs.

pulp_container/app/utils.py
```diff
@@ -309,3 +314,39 @@ def get_content_data(saved_artifact):
...
+    return not include(x, patterns)
+
+
+def filter_resource(element, include_patterns, exclude_patterns):
+    """
+    Verify if the repository should have access to the remote upstream based on the include_patterns
+    and exclude_patterns filters.
+    """
+    if not (include_patterns or []) and not (exclude_patterns or []):
+        return True
```
lubosmj, line 339:
```suggestion
    if not (include_patterns or exclude_patterns):
        return True
```

pulp_container/app/registry_api.py
```diff
@@ -309,6 +312,12 @@ def get_pull_through_drv(self, path):
         if not pull_through_cache_distribution:
             raise RepositoryNotFound(name=path)
 
+        pull_through_remote = models.ContainerPullThroughRemote.objects.get(
+            pk=pull_through_cache_distribution.remote
+        )
+        if not filter_resource(path, pull_through_remote.includes, pull_through_remote.excludes):
+            raise RepositoryNotFound(name=path)
```
lubosmj, line 319: I think we can access the pull-through remote "directly". Then, we can also pass `pull_through_cache_distribution.remote` to `_get_pull_through_remote_data`.

git-hyagi: `pull_through_cache_distribution.remote` is of type `Remote` (not `ContainerPullThroughRemote`) ... so there is no includes/excludes definition in it :cry:

lubosmj: You can call cast() on it to get the correct type.

pulp_container/app/registry_api.py
```diff
@@ -309,13 +312,21 @@ def get_pull_through_drv(self, path):
         if not pull_through_cache_distribution:
             raise RepositoryNotFound(name=path)
 
+        pull_through_remote = models.ContainerPullThroughRemote.objects.filter(
+            pk=pull_through_cache_distribution.remote_id
+        ).values()[0]
+        if not filter_resource(
+            path, pull_through_remote.get("includes"), pull_through_remote.get("excludes")
```
lubosmj, line 319: We may want to run the validation against `upstream_name` (i.e., repository name), not `path`. `path` includes the namespace of the caching distribution that is not transparent to the remote registry.

pulp_container/app/serializers.py
```diff
@@ -307,9 +307,11 @@ class ContainerPullThroughRemoteSerializer(RemoteSerializer):
     """
 
     policy = serializers.ChoiceField(choices=[Remote.ON_DEMAND], default=Remote.ON_DEMAND)
+    includes = serializers.JSONField(required=False, allow_null=True)
+    excludes = serializers.JSONField(required=False, allow_null=True)
```
lubosmj, line 311: Maybe, we should go with `serializers.ListField` and `fields.ArrayField` like we did for `include_tags`/`exclude_tags`. This will force users to use a valid list as an input.

pulp_container/app/registry_api.py
```diff
@@ -309,13 +312,21 @@ def get_pull_through_drv(self, path):
         if not pull_through_cache_distribution:
             raise RepositoryNotFound(name=path)
 
+        pull_through_remote = models.ContainerPullThroughRemote.objects.filter(
+            pk=pull_through_cache_distribution.remote_id
+        ).values()[0]
```
lubosmj, line 317: I think we can access `pull_through_remote` directly from `pull_through_cache_distribution.remote`. To save one DB query, we can run this a couple of lines above:

        pull_through_cache_distribution = (
            models.ContainerPullThroughDistribution.objects.annotate(path=Value(path))
            .filter(path__startswith=F("base_path"))
            .order_by("-base_path")
            .select_related("remote")
            .first()
        )

pulp_container/tests/functional/api/test_remote_filter_pull_through.py
```diff
@@ -0,0 +1,74 @@
...
+@pytest.fixture
+def pull_and_verify(
+    capfd,
+    delete_orphans_pre,
+    local_registry,
+    registry_client,
+):
+    def _pull_and_verify(images, pull_through_distribution, includes, excludes):
+        distr = pull_through_distribution(includes, excludes)
+        for _, image_path in enumerate(images, start=1):
```
lubosmj, line 21:
```suggestion
        for image_path in images:
```

pulp_container/tests/functional/api/test_remote_filter_pull_through.py
```diff
@@ -0,0 +1,74 @@
...
+                with pytest.raises(subprocess.CalledProcessError):
+                    local_registry.pull(local_image_path)
+                assert (
+                    re.search(
+                        ".*Repository not found.*",
+                        capfd.readouterr().err,
+                    )
+                    is not None
+                )
+                continue
```
lubosmj, line 35: It is very hard to read and understand this part. We can pass expected values into this function and do the following:
```
if image_path not in XXX:
    with pytest.raises(subprocess.CalledProcessError) as exc_info:
        local_registry.pull(local_image_path)
    assert "Repository not found" in exc_info.value  # if possible, https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions
else:
    local_registry.pull(local_image_path)
```

pulp_container/app/registry_api.py
```diff
@@ -389,15 +388,37 @@ def create_dr(self, path, request):
...
+        raise RepositoryNotFound(name=path)
+
+    upstream_name = path.split(pull_through_cache_distribution.base_path, maxsplit=1)[1].strip("/")
+    pull_through_remote = models.ContainerPullThroughRemote.objects.filter(
+        pk=pull_through_cache_distribution.remote_id
+    ).values()[0]
+    if not filter_resource(
+        upstream_name, pull_through_remote.get("includes"), pull_through_remote.get("excludes")
+    ):
+        raise RepositoryNotFound(name=path)
```
lubosmj, line 417: I wonder if it would not be better to return an HTTP 403 error instead (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403). What do you think?

pulp_container/app/registry_api.py
```diff
@@ -389,15 +388,37 @@ def create_dr(self, path, request):
...
+        .order_by("-base_path")
+        .first()
+    )
+    if not pull_through_cache_distribution:
+        raise RepositoryNotFound(name=path)
+
+    upstream_name = path.split(pull_through_cache_distribution.base_path, maxsplit=1)[1].strip("/")
+    pull_through_remote = models.ContainerPullThroughRemote.objects.filter(
+        pk=pull_through_cache_distribution.remote_id
+    ).values()[0]
```
lubosmj, line 413: According to chatgpt, there are two ways we can follow to refactor this part:
```
from django.db import models
from django.forms.models import model_to_dict

class YourModel(models.Model):
    field1 = models.CharField(max_length=100)
    field2 = models.IntegerField()
    field3 = models.DateField()

    def to_dict(self):
        return {
            'field1': self.field1,
            'field2': self.field2,
            'field3': self.field3,
        }

# 1. Using model_to_dict
model_instance = YourModel.objects.get(pk=1)
model_dict = model_to_dict(model_instance)

# 2. Using the custom to_dict method
model_dict = model_instance.to_dict()
```

Do you find any of these methods more elegant? Maybe, we will not even need `_get_pull_through_remote_data`.

pulp_container/app/models.py
```diff
@@ -503,6 +503,10 @@ class ContainerPullThroughRemote(Remote, AutoAddObjPermsMixin):
     from within a single instance of this remote.
     """
 
+    TYPE = "pull-through"
+    includes = fields.ArrayField(models.TextField(null=True), null=True)
```
lubosmj, line 507:
```suggestion
    TYPE = "pull-through"

    includes = fields.ArrayField(models.TextField(null=True), null=True)
```

It looks neat if we semantically separate model fields and constants.

pulp_container/tests/functional/api/test_remote_filter_pull_through.py
```diff
@@ -0,0 +1,74 @@
...
+            assert local_image[0]["Id"] == remote_image[0]["Id"]
+
+    return _pull_and_verify
+
+
+def test_no_filter(pull_through_distribution, pull_and_verify):
+    images = [f"{PULP_FIXTURE_1}:manifest_a", f"{PULP_FIXTURE_1}:manifest_b"]
+    includes = None
+    excludes = []
+    pull_and_verify(images, pull_through_distribution, includes, excludes)
```
lubosmj, line 49: If you run out of function names, it is still feasible to run the same test with different parameters, like https://github.com/pulp/pulp_container/blob/e52fa5dd29af91ef0857035d7f1476f49bb74880/pulp_container/tests/functional/api/test_sync.py#L107.

lubosmj (changes requested):
Thanks! We are almost there! 🎸

staging_docs/admin/guides/pull-through-caching.md
```diff
@@ -43,3 +43,39 @@ ensures a more reliable container deployment system in production environments.
...
+podman pull localhost:24817/docker-cache/pulp/molecule_debian11
+Error response from daemon: unknown: No repository found for the defined remote filters
+```
+
+and since we are *including* only repositories with \*pulp\* in its name (`includes=["*pulp*"]`),
+the following image pull should also fail:
+
+```
+podman pull localhost:24817/docker-cache/library/hello-world
+Error response from daemon: unknown: No repository found for the defined remote filters
```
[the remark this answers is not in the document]
git-hyagi: oops... my bad
I changed the code and forgot to update the doc :sweat_smile:

pulp_container/app/registry_api.py
git-hyagi: hum... when I see a 403 I think about roles/rights and, in this new filter feature, we are not denying or blocking users access, it feels more like "not making it visible" or just removing/bypassing a resource from the results as a convenience and not denying it since they could create a new pull-through remote with different includes/excludes ... and maybe a 403 could "misdirect" into a role troubleshoot. Do you agree?

lubosmj: Yes, you are right. Let's leave it unchanged then.

pulp_container/tests/functional/api/test_remote_filter_pull_through.py
git-hyagi: I tried to use the exception output, but the outputs from subprocesses seem to not be captured:
https://docs.pytest.org/en/6.2.x/capture.html

> If you want to capture on filedescriptor level you can use the capfd fixture which offers the exact same interface but allows to also capture output from libraries or subprocesses that directly write to operating system level output streams (FD1 and FD2).

pulp_container/app/models.py
```diff
@@ -512,6 +517,35 @@ class Meta:
             ),
         ]
 
+    def model_to_dict(self):
```
lubosmj, line 520: Did not `from django.forms.models import model_to_dict` suit our needs?

pulp_container/tests/functional/api/test_remote_filter_pull_through.py
```diff
@@ -0,0 +1,63 @@
...
+    local_registry,
+    registry_client,
+):
+    def _pull_and_verify(images, pull_through_distribution, includes, excludes):
+        distr = pull_through_distribution(includes, excludes)
+        for image_path in images:
+            remote_image_path = f"{REGISTRY_V2}/{image_path}"
+            local_image_path = f"{distr.base_path}/{image_path}"
+
+            if excludes and "hello-world" not in image_path:
```
lubosmj, line 24: What is `hello-world`? It came out of nowhere. Is not it better to use the following structure?
```
if image_path not in expected_paths:
   blabla
```

pulp_container/app/models.py
git-hyagi: I think it does, but when I tried to use it I had 3 concerns:
* the "old version" was returning a field called `pulp_domain` whereas with django's model_to_dict it returns `pulp_domain_id`, so we would need to somehow handle this corner case
* with django's model_to_dict we would still need _get_pull_through_remote_data (not the method itself, but the `IGNORED_PULL_THROUGH_REMOTE_ATTRIBUTES` that is used by it)
* had to remove the url from `IGNORED_PULL_THROUGH_REMOTE_ATTRIBUTES` (it was causing an exception without it)

Here is what I tried:
```
    remote_data = model_to_dict(pull_through_remote, exclude=IGNORED_PULL_THROUGH_REMOTE_ATTRIBUTES)
    remote_data['pulp_domain_id'] = remote_data['pulp_domain']
    remote_data.pop('pulp_domain', None)
```

Do you think we should use it instead?

lubosmj: Did the execution fail when you passed "pulp_domain_id" to "Remote.create"? Either way, we can try doing what you have suggested:
```
IGNORED_PULL_THROUGH_REMOTE_ATTRIBUTES = ("includes", "excludes", "pulp_domain")
remote_data = model_to_dict(pull_through_remote, exclude=IGNORED_PULL_THROUGH_REMOTE_ATTRIBUTES)
```

pulp_container/tests/functional/api/test_remote_filter_pull_through.py
```diff
@@ -0,0 +1,76 @@
...
+                with pytest.raises(subprocess.CalledProcessError):
+                    local_registry.pull(local_image_path)
+                assert "Repository not found" in capfd.readouterr().err
+                continue
+
+            local_registry.pull(local_image_path)
+            local_image = local_registry.inspect(local_image_path)
+            registry_client.pull(remote_image_path)
+            remote_image = registry_client.inspect(remote_image_path)
+            assert local_image[0]["Id"] == remote_image[0]["Id"]
```
lubosmj, line 34:
```suggestion
        else:
            local_registry.pull(local_image_path)
            local_image = local_registry.inspect(local_image_path)
            registry_client.pull(remote_image_path)
            remote_image = registry_client.inspect(remote_image_path)
            assert local_image[0]["Id"] == remote_image[0]["Id"]
```

pulp_container/app/utils.py
```diff
@@ -309,3 +314,39 @@ def get_content_data(saved_artifact):
         raw_data = file.read()
     content_data = json.loads(raw_data)
     return content_data, raw_data
+
+
+def include(x, patterns):
+    """
+    Checks if any item from `patterns` matches x, meaning it should be included as a remote repo.
```
lubosmj, line 321: We should make these docstrings more generic. We do not know how the functions are going to be used. The caller is the one that decides the purpose of the functions. For these one-line functions, it might be worth omitting the docstring too. Also, we do not care only about remote repositories in this context.

lubosmj (approved):
Thanks! A little more nitpicking, but approving.

PR #1685, opened 2024-07-02
Fix container remote upstream name option

stejoo:
The example code to create a container remote contains an argument that is incorrect:

> Error: No such option: --upstream_name (Possible options: --upstream-name, --username)

This commit modifies the underscore in the argument to a hyphen. That resolves the "No such option" error.

### DIFF
--- a/staging_docs/user/tutorials/01-sync-and-host.md
+++ b/staging_docs/user/tutorials/01-sync-and-host.md
@@ -23,7 +23,7 @@ including `quay`, `google container registry`, or even another instance of Pulp.
     Container plugin supports both Docker and OCI media types.
 
 ```bash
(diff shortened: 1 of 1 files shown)


lubosmj:
Thanks for opening this PR! Would you mind also adding `[noissue]` to the commit description to make the CI green, please?

stejoo:
I modified the commit message. Is this what you had in mind? :smile:

lubosmj:
Kind of. But, we can accept it as it is. 🪂

A description is not a commit message. For example, see this commit: https://github.com/pulp/pulp_container/commit/e52fa5dd29af91ef0857035d7f1476f49bb74880. It was created like:
```
git commit -m "Update CI" -m "[noissue]"
```

stejoo:
Ah. You prefer it in the commit message body instead of title. Got it. Hold on. :wink:

lubosmj (approved):

Issue #479, opened 2021-12-15
I can specify a repository_version as the build context for container images


pulpbot:
Author: @mdellweg (mdellweg)

Redmine Issue: 7899, https://pulp.plan.io/issues/7899

---

Instead specifying a list of artifacts with relative paths, one should be able to specify a repository_version.
The contained artifacts would then describe the layout of the build context for container images.

lubosmj:
This should go hand in hand with serving artifacts from pulp-file, which is now a part of pulpcore.

lubosmj:
It was suggested that it would be a better experience if the viewset performed all the permission checks.

PR #1687, opened 2024-07-02
Add repository_version param as a building context

git-hyagi:
closes: #479

### DIFF
--- a/CHANGES/479.feature
+++ b/CHANGES/479.feature
@@ -1,0 +1,2 @@
+The `build_image` endpoint has been refactored to accept `build_context`
+(i.e., a file repository version) instead of raw artifacts. The same applies to Containerfile."
--- a/docs/admin/guides/build-image.md
+++ b/docs/admin/guides/build-image.md
@@ -6,24 +6,33 @@
     guaranteed.
 
 Users can add new images to a container repository by uploading a Containerfile. The syntax for
-Containerfile is the same as for a Dockerfile. The same REST API endpoint also accepts a JSON
-string that maps artifacts in Pulp to a filename. Any artifacts passed in are available inside the
-build container at `/pulp_working_directory`.
+Containerfile is the same as for a Dockerfile.
 
-## Create a Repository
-
-```bash
-REPO_HREF=$(pulp container repository create --name building | jq -r '.pulp_href')
+To pass arbitrary files or artifacts to the image building context, the `build_context` property (a reference to a file repository) can be provided in the request payload.
+These files can be referenced in Containerfile by passing their `relative-path`:
+```
+ADD/COPY <file relative-path> <location in container>
 ```
 
-## Create an Artifact
+It is possible to define the Containerfile in two ways:
+* from a [local file](site:pulp_container/docs/admin/guides/build-image#build-from-a-containerfile-uploaded-during-build-request) and pass it during build request
+* from an [existing file](site:pulp_container/docs/admin/guides/build-image#upload-the-containerfile-as-a-file-content) in the `build_context`
+
+## Create a Container Repository
 
 ```bash
+CONTAINER_REPO=$(pulp container repository create --name building | jq -r '.pulp_href')
+```
+
+## Create a File Repository and populate it
+
+```bash
+FILE_REPO=$(pulp file repository create --name bar --autopublish | jq -r '.pulp_href')
+
 echo 'Hello world!' > example.txt
 
-ARTIFACT_HREF=$(http --form POST http://localhost/pulp/api/v3/artifacts/ \
-    file@./example.txt \
-    | jq -r '.pulp_href')
+pulp file content upload --relative-path foo/bar/example.txt \
+--file ./example.txt --repository bar
 ```
 
 ## Create a Containerfile
@@ -31,22 +40,38 @@ ARTIFACT_HREF=$(http --form POST http://localhost/pulp/api/v3/artifacts/ \
 ```bash
 echo 'FROM centos:7
 
-# Copy a file using COPY statement. Use the relative path specified in the 'artifacts' parameter.
+# Copy a file using COPY statement. Use the path specified in the '--relative-path' parameter.
 COPY foo/bar/example.txt /inside-image.txt
 
 # Print the content of the file when the container starts
 CMD ["cat", "/inside-image.txt"]' >> Containerfile
 ```
 
-## Build an OCI image
+
+## Build from a Containerfile uploaded during build request
 
 ```bash
-TASK_HREF=$(http --form POST :$REPO_HREF'build_image/' containerfile@./Containerfile \
-artifacts="{\"$ARTIFACT_HREF\": \"foo/bar/example.txt\"}"  | jq -r '.task')
+TASK_HREF=$(http --form POST ${BASE_ADDR}${CONTAINER_REPO}'build_image/' "containerfile@./Containerfile" \
+build_context=${FILE_REPO}versions/1/ | jq -r '.task')
 ```
+
+## Upload the Containerfile to a File Repository and use it to build
+
+### Upload the Containerfile as a File Content
+
+```bash
+pulp file content upload --relative-path MyContainerfile --file ./Containerfile --repository bar
+```
+
+### Build an OCI image from a Containerfile present in build_context
+
+```bash
+TASK_HREF=$(http --form POST ${BASE_ADDR}${CONTAINER_REPO}'build_image/' containerfile_name=MyContainerfile \
+build_context=${FILE_REPO}versions/2/ | jq -r '.task')
+```
+
 
 !!! warning
 
-    Non-staff users, lacking read access to the `artifacts` endpoint, may encounter restricted
-    functionality as they are prohibited from listing artifacts uploaded to Pulp and utilizing
-    them within the build process.
+    File repositories synced with the on-demand policy will not automatically download the missing artifacts.
+    Trying to build an image using a file that has not yet been downloaded will fail.
--- a/pulp_container/app/serializers.py
+++ b/pulp_container/app/serializers.py
@@ -1,12 +1,10 @@
 from gettext import gettext as _
-import os
 import re
 
 from django.core.validators import URLValidator
 from rest_framework import serializers
 
 from pulpcore.plugin.models import (
-    Artifact,
     ContentRedirectContentGuard,
     Remote,
     Repository,
@@ -30,8 +28,10 @@ from pulpcore.plugin.serializers import (
     ValidateFieldsMixin,
 )
 
+from pulp_file.app.models import FileContent
 from pulp_container.app import models
 from pulp_container.constants import SIGNATURE_TYPE
+
 
 VALID_SIGNATURE_NAME_REGEX = r"^sha256:[0-9a-f]{64}@[0-9a-f]{32}$"
 VALID_TAG_REGEX = r"^[A-Za-z0-9][A-Za-z0-9._-]*$"
@@ -758,13 +758,12 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
     A repository must be specified, to which the container image content will be added.
     """
 
-    containerfile_artifact = RelatedField(
-        many=False,
-        lookup_field="pk",
-        view_name="artifacts-detail",
-        queryset=Artifact.objects.all(),
+    containerfile_name = serializers.CharField(
+        required=False,
+        allow_blank=True,
         help_text=_(
-            "Artifact representing the Containerfile that should be used to run podman-build."
+            "Name of the Containerfile, from build_context, that should be used to run "
+            "podman-build."
         ),
     )
     containerfile = serializers.FileField(
@@ -774,65 +773,77 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
     tag = serializers.CharField(
         required=False, default="latest", help_text="A tag name for the new image being built."
     )
-    artifacts = serializers.JSONField(
+    build_context = RepositoryVersionRelatedField(
         required=False,
-        help_text="A JSON string where each key is an artifact href and the value is it's "
-        "relative path (name) inside the /pulp_working_directory of the build container "
-        "executing the Containerfile.",
+        help_text=_("RepositoryVersion to be used as the build context for container images."),
+        allow_null=True,
+        queryset=RepositoryVersion.objects.filter(repository__pulp_type="file.file"),
     )
-
-    def __init__(self, *args, **kwargs):
-        """Initializer for OCIBuildImageSerializer."""
-        super().__init__(*args, **kwargs)
-        self.fields["containerfile_artifact"].required = False
 
     def validate(self, data):
         """Validates that all the fields make sense."""
         data = super().validate(data)
 
-        if "containerfile" in data:
-            if "containerfile_artifact" in data:
-                raise serializers.ValidationError(
-                    _("Only one of 'containerfile' and 'containerfile_artifact' may be specified.")
-                )
-            data["containerfile_artifact"] = Artifact.init_and_validate(data.pop("containerfile"))
-        elif "containerfile_artifact" in data:
-            data["containerfile_artifact"].touch()
-        else:
+        if bool(data.get("containerfile", None)) == bool(data.get("containerfile_name", None)):
             raise serializers.ValidationError(
-                _("'containerfile' or 'containerfile_artifact' must " "be specified.")
+                _("Exactly one of 'containerfile' or 'containerfile_name' must be specified.")
             )
-        artifacts = {}
-        if "artifacts" in data:
-            for url, relative_path in data["artifacts"].items():
-                if os.path.isabs(relative_path):
+
+        if "containerfile_name" in data and "build_context" not in data:
+            raise serializers.ValidationError(
+                _("A 'build_context' must be specified when 'containerfile_name' is present.")
+            )
+
+        # TODO: this can be removed after https://github.com/pulp/pulpcore/issues/5786
+        if data.get("build_context", None):
+            data["repository_version"] = data["build_context"]
+
+        return data
+
+    def deferred_files_validation(self, data):
+        """
+        Defer the validation of on_demand_artifacts and the `Containerfile` to avoid rerunning
+        unnecessary database queries when checking permissions (DRF Access Policy).
+        """
+        if build_context := data.get("build_context", None):
+
+            # check if the on_demand_artifacts exist
+            for on_demand_artifact in build_context.on_demand_artifacts.iterator():
+                if not on_demand_artifact.content_artifact.artifact:
                     raise serializers.ValidationError(
-                        _("Relative path cannot start with '/'. " "{0}").format(relative_path)
+                        _(
+                            "It is not possible to use File content synced with on-demand "
+                            "policy without pulling the data first."
+                        )
                     )
-                artifactfield = RelatedField(
-                    view_name="artifacts-detail",
-                    queryset=Artifact.objects.all(),
-                    source="*",
-                    initial=url,
+
+            # check if the containerfile_name exists in the build_context (File Repository)
+            if (
+                data.get("containerfile_name", None)
+                and not FileContent.objects.filter(
+                    repositories__in=[build_context.repository.pk],
+                    relative_path=data["containerfile_name"],
+                ).exists()
+            ):
+                raise serializers.ValidationError(
+                    _(
+                        'Could not find the Containerfile "'
+                        + data["containerfile_name"]
+                        + '" in the build_context provided'
+                    )
                 )
-                try:
-                    artifact = artifactfield.run_validation(data=url)
-                    artifact.touch()
-                    artifacts[str(artifact.pk)] = relative_path
-                except serializers.ValidationError as e:
-                    # Append the URL of missing Artifact to the error message
-                    e.detail[0] = "%s %s" % (e.detail[0], url)
-                    raise e
-        data["artifacts"] = artifacts
+
+            data["build_context_pk"] = build_context.repository.pk
+
         return data
 
     class Meta:
         fields = (
-            "containerfile_artifact",
+            "containerfile_name",
             "containerfile",
             "repository",
             "tag",
-            "artifacts",
+            "build_context",
         )
 
 
--- a/pulp_container/app/tasks/builder.py
+++ b/pulp_container/app/tasks/builder.py
@@ -14,7 +14,12 @@ from pulp_container.app.models import (
 )
 from pulp_container.constants import MEDIA_TYPE
 from pulp_container.app.utils import calculate_digest
-from pulpcore.plugin.models import Artifact, ContentArtifact, Content
+from pulpcore.plugin.models import (
+    Artifact,
+    ContentArtifact,
+    Content,
+    PulpTemporaryFile,
+)
 
 
 def get_or_create_blob(layer_json, manifest, path):
@@ -43,7 +48,7 @@ def get_or_create_blob(layer_json, manifest, path):
             artifact=layer_artifact, content=blob, relative_path=layer_json["digest"]
         ).save()
     if layer_json["mediaType"] != MEDIA_TYPE.CONFIG_BLOB_OCI:
-        BlobManifest(manifest=manifest, manifest_blob=blob).save()
+        BlobManifest.objects.get_or_create(manifest=manifest, manifest_blob=blob)
     return blob
 
 
@@ -68,15 +73,13 @@ def add_image_from_directory_to_repository(path, repository, tag):
     manifest_digest = calculate_digest(bytes_data)
     manifest_text_data = bytes_data.decode("utf-8")
 
-    manifest = Manifest(
+    manifest, _ = Manifest.objects.get_or_create(
         digest=manifest_digest,
         schema_version=2,
         media_type=MEDIA_TYPE.MANIFEST_OCI,
         data=manifest_text_data,
     )
-    manifest.save()
-    tag = Tag(name=tag, tagged_manifest=manifest)
-    tag.save()
+    tag, _ = Tag.objects.get_or_create(name=tag, tagged_manifest=manifest)
 
     with repository.new_version() as new_repo_version:
         manifest_json = json.loads(manifest_text_data)
@@ -95,10 +98,100 @@ def add_image_from_directory_to_repository(path, repository, tag):
     return new_repo_version
 
 
+def build_image(
+    containerfile_name=None,
+    containerfile_tempfile_pk=None,
+    build_context_pk=None,
+    repository_pk=None,
+    tag=None,
+):
+    """
+    Builds an OCI container image from a Containerfile.
+
+    The artifacts are made available inside the build container at the paths specified by their
+    values. The Containerfile can make use of these files during build process.
+
+    Args:
+        containerfile_name (str): The Containerfile relative_path from the build_context repository
+        containerfile_tempfile_pk (str): The pk of a PulpTemporaryFile that contains
+                                         the Containerfile
+        build_context_pk (str): The pk of the RepositoryVersion used as the build context
+        repository_pk (str): The pk of a Repository to add the OCI container image
+        tag (str): Tag name for the new image in the repository
+
+    Returns:
+        A class:`pulpcore.plugin.models.RepositoryVersion` that contains the new OCI container
+        image and tag.
+
+    """
+    if not containerfile_tempfile_pk and not containerfile_name:
+        raise RuntimeError("Neither a name nor temporary file for the Containerfile was specified.")
+
+    if containerfile_tempfile_pk:
+        containerfile_artifact = PulpTemporaryFile.objects.get(pk=containerfile_tempfile_pk)
+
+    repository = ContainerRepository.objects.get(pk=repository_pk)
+    name = str(uuid4())
+    with tempfile.TemporaryDirectory(dir=".") as working_directory:
+        working_directory = os.path.abspath(working_directory)
+        context_path = os.path.join(working_directory, "context")
+        os.makedirs(context_path, exist_ok=True)
+
+        if build_context_pk:
+            content_artifacts = ContentArtifact.objects.filter(
+                content__pulp_type="file.file", content__repositories__in=[build_context_pk]
+            ).order_by("-content__pulp_created")
+            for content_artifact in content_artifacts.select_related("artifact").iterator():
+                if content_artifact.relative_path == containerfile_name:
+                    containerfile_artifact = content_artifact.artifact
+                    continue
+                _copy_file_from_artifact(
+                    context_path, content_artifact.relative_path, content_artifact.artifact.file
+                )
+
+        containerfile_name = containerfile_name or "Containerfile"
+        _copy_file_from_artifact(working_directory, containerfile_name, containerfile_artifact)
+        containerfile_path = os.path.join(working_directory, containerfile_name)
+
+        bud_cp = subprocess.run(
+            [
+                "podman",
+                "build",
+                "-f",
+                containerfile_path,
+                "-t",
+                name,
+                context_path,
+                "--isolation",
+                "rootless",
+            ],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        if bud_cp.returncode != 0:
+            raise Exception(bud_cp.stderr)
+        image_dir = os.path.join(working_directory, "image")
+        os.makedirs(image_dir, exist_ok=True)
+        push_cp = subprocess.run(
+            ["podman", "push", "-f", "oci", name, "dir:{}".format(image_dir)],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        if push_cp.returncode != 0:
+            raise Exception(push_cp.stderr)
+        repository_version = add_image_from_directory_to_repository(image_dir, repository, tag)
+        if isinstance(containerfile_artifact, PulpTemporaryFile):
+            containerfile_artifact.delete()
+
+    return repository_version
+
+
 def build_image_from_containerfile(
     containerfile_pk=None, artifacts=None, repository_pk=None, tag=None
 ):
     """
+    DEPRECATED: this function is deprecated and will be removed in a future release.
+                Keeping it for now for backward compatibility.
     Builds an OCI container image from a Containerfile.
 
     The artifacts are made available inside the build container at the paths specified by their
@@ -166,3 +259,12 @@ def build_image_from_containerfile(
         repository_version = add_image_from_directory_to_repository(image_dir, repository, tag)
 
     return repository_version
+
+
+def _copy_file_from_artifact(context_path, relative_path, artifact):
+    dest_path = os.path.join(context_path, relative_path)
+    dirs = os.path.dirname(dest_path)
+    if dirs:
+        os.makedirs(dirs, exist_ok=True)
+    with open(dest_path, "wb") as dest:
+        shutil.copyfileobj(artifact.file, dest)
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -7,7 +7,6 @@ Check `Plugin Writer's Guide`_ for more details.
 
 import logging
 
-from django.db import IntegrityError
 from django.db.models import Q
 
 from django_filters import CharFilter, MultipleChoiceFilter
@@ -16,9 +15,9 @@ from drf_spectacular.utils import extend_schema
 from rest_framework import mixins
 from rest_framework.decorators import action
 
-from pulpcore.plugin.models import RepositoryVersion
+from pulpcore.plugin.models import RepositoryVersion, PulpTemporaryFile
 from pulpcore.plugin.serializers import AsyncOperationResponseSerializer
-from pulpcore.plugin.models import Artifact, Content
+from pulpcore.plugin.models import Content
 from pulpcore.plugin.tasking import dispatch, general_multi_delete
 from pulpcore.plugin.util import (
     extract_pk,
@@ -696,6 +695,7 @@ class ContainerRepositoryViewSet(
                 "condition": [
                     "has_model_or_obj_perms:container.build_image_containerrepository",
                     "has_model_or_obj_perms:container.view_containerrepository",
+                    "has_repo_or_repo_ver_param_model_or_obj_perms:file.view_filerepository",
                 ],
             },
             {
@@ -937,26 +937,26 @@ class ContainerRepositoryViewSet(
         )
 
         serializer.is_valid(raise_exception=True)
+        serialized_data = serializer.deferred_files_validation(serializer.validated_data)
 
-        containerfile = serializer.validated_data["containerfile_artifact"]
-        try:
-            containerfile.save()
-        except IntegrityError:
-            containerfile = Artifact.objects.get(sha256=containerfile.sha256)
-            containerfile.touch()
-        tag = serializer.validated_data["tag"]
+        containerfile_tempfile_pk = None
+        if containerfile := serialized_data.get("containerfile", None):
+            temp_file = PulpTemporaryFile(file=containerfile)
+            temp_file.save()
+            containerfile_tempfile_pk = str(temp_file.pk)
 
-        artifacts = serializer.validated_data["artifacts"]
-        Artifact.objects.filter(pk__in=artifacts.keys()).touch()
+        containerfile_name = serialized_data.get("containerfile_name", None)
+        tag = serialized_data["tag"]
 
         result = dispatch(
-            tasks.build_image_from_containerfile,
+            tasks.build_image,
             exclusive_resources=[repository],
             kwargs={
-                "containerfile_pk": str(containerfile.pk),
+                "containerfile_name": containerfile_name,
+                "containerfile_tempfile_pk": containerfile_tempfile_pk,
                 "tag": tag,
                 "repository_pk": str(repository.pk),
-                "artifacts": artifacts,
+                "build_context_pk": serialized_data.get("build_context_pk", None),
             },
         )
         return OperationPostponedResponse(result, request)
--- a/pulp_container/tests/functional/api/test_build_images.py
+++ b/pulp_container/tests/functional/api/test_build_images.py
@@ -2,16 +2,11 @@ import pytest
 
 from tempfile import NamedTemporaryFile
 
-from pulp_smash.pulp3.utils import (
-    gen_distribution,
-    gen_repo,
-)
+from pulpcore.tests.functional.utils import PulpTaskError
+from pulp_smash.pulp3.utils import gen_distribution
 from pulp_smash.pulp3.bindings import monitor_task
 
-from pulpcore.client.pulp_container import (
-    ContainerContainerDistribution,
-    ContainerContainerRepository,
-)
+from pulpcore.client.pulp_container import ApiException, ContainerContainerDistribution
 
 
 @pytest.fixture
@@ -19,7 +14,7 @@ def containerfile_name():
     """A fixture for a basic container file used for building images."""
     with NamedTemporaryFile() as containerfile:
         containerfile.write(
-            b"""FROM busybox:latest
+            b"""FROM quay.io/quay/busybox:latest
 # Copy a file using COPY statement. Use the relative path specified in the 'artifacts' parameter.
 COPY foo/bar/example.txt /tmp/inside-image.txt
 # Print the content of the file when the container starts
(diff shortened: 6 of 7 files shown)


staging_docs/admin/guides/build-image.md
```diff
@@ -50,3 +52,17 @@ artifacts="{\"$ARTIFACT_HREF\": \"foo/bar/example.txt\"}"  | jq -r '.task')
...
+
+```bash
+ARTIFACT_SHA256=$(http :$ARTIFACT_HREF | jq -r '.sha256')
+pulp file repository create --name bar
+
+REPO_VERSION=$(pulp file content create --relative-path foo/bar/example.txt \
+--sha256 $ARTIFACT_SHA256 --repository bar | jq .pulp_href)
+
+TASK_HREF=$(http --form POST  :$REPO_HREF'build_image/' -F "containerfile@./Containerfile" \
+repo_version=$REPO_VERSION | jq -r '.task')
```
lubosmj, line 67:
```suggestion
TASK_HREF=$(http --form POST  :$REPO_HREF'build_image/' "containerfile@./Containerfile" \
repo_version=$REPO_VERSION | jq -r '.task')
```

pulp_container/app/viewsets.py
```diff
@@ -696,6 +696,7 @@ class ContainerRepositoryViewSet(
                 "condition": [
                     "has_model_or_obj_perms:container.build_image_containerrepository",
                     "has_model_or_obj_perms:container.view_containerrepository",
+                    "has_repository_model_or_obj_perms:file.view_filerepository",
```
lubosmj, line 699: Due to this line, I am receiving an HTTP 500 error.
```
[pulp]  | pulp [d8943394f02c4759b3deac8f518d2218]: django.request:ERROR: Internal Server Error: /pulp/api/v3/repositories/container/container/01907884-8984-7905-bc76-58b8c747c347/build_image/
[pulp]  | Traceback (most recent call last):
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 55, in inner
[pulp]  |     response = get_response(request)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 197, in _get_response
[pulp]  |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 56, in wrapper_view
[pulp]  |     return view_func(*args, **kwargs)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/viewsets.py", line 124, in view
[pulp]  |     return self.dispatch(request, *args, **kwargs)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
[pulp]  |     response = self.handle_exception(exc)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
[pulp]  |     self.raise_uncaught_exception(exc)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
[pulp]  |     raise exc
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 497, in dispatch
[pulp]  |     self.initial(request, *args, **kwargs)
[pulp]  |   File "/src/pulpcore/pulpcore/app/viewsets/base.py", line 300, in initial
[pulp]  |     super().initial(request, *args, **kwargs)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 415, in initial
[pulp]  |     self.check_permissions(request)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 332, in check_permissions
[pulp]  |     if not permission.has_permission(request, self):
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_access_policy/access_policy.py", line 69, in has_permission
[pulp]  |     allowed = self._evaluate_statements(statements, request, view, action)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_access_policy/access_policy.py", line 113, in _evaluate_statements
[pulp]  |     matched = self._get_statements_matching_conditions(
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_access_policy/access_policy.py", line 262, in _get_statements_matching_conditions
[pulp]  |     passed = self._check_condition(condition, request, view, action)
[pulp]  |   File "/usr/local/lib/python3.9/site-packages/rest_access_policy/access_policy.py", line 286, in _check_condition
[pulp]  |     result = method(request, view, action, arg)
[pulp]  |   File "/src/pulpcore/pulpcore/app/global_access_conditions.py", line 407, in has_repository_model_or_obj_perms
[pulp]  |     return request.user.has_perm(permission) or has_repository_obj_perms(
[pulp]  |   File "/src/pulpcore/pulpcore/app/global_access_conditions.py", line 367, in has_repository_obj_perms
[pulp]  |     plugin_repository = Repository.objects.get(pk=view.kwargs["repository_pk"]).cast()
[pulp]  | KeyError: 'repository_pk'
```

Reproducer:
```
pulp user create --username test --password krokodyl1 --email test@test.com
pulp user role-assignment add --username test --role "container.containerdistribution_collaborator" --object ""
pulp user role-assignment add --username test --role "container.containerrepository_content_manager" --object ""

FILE_REPO=$(pulp file repository create --name foo --autopublish | jq -r '.pulp_href')
pulp file content upload --repository foo --file ./setup.py --relative-path setup.py

CONTAINER_REPO=$(pulp container repository create --name build | jq -r '.pulp_href') 

http -a test:krokodyl1 --form POST :5001${CONTAINER_REPO}build_image/ "containerfile@./Containerfile" repo_version=${FILE_REPO}versions/1/
```

-------------------------------------

I do not think we can inject the policy from other entities' contexts here (repo_version=FILE_HREF). The policy is available only for the current viewset scope (CONTAINER_REPO). I think we need to run permission checks manually. Similarly to https://github.com/pulp/pulp_container/blob/a832b9fc9a06cf96fda20fc7203f06a2baa6b9c4/pulp_container/app/registry_api.py#L531 or create a customized policy condition, like in https://github.com/pulp/pulp_container/blob/a832b9fc9a06cf96fda20fc7203f06a2baa6b9c4/pulp_container/app/global_access_conditions.py#L55.

@mdellweg, any ideas on how to properly add permission checks for entities created from within a different plugin? In our case, we want to check if a user has permissions to read pulp_file repository and build an image inside pulp_container repository.

pulp_container/tests/functional/api/test_build_images.py
```diff
@@ -61,3 +63,45 @@ def test_build_image(
...
+        "txt": file_content_unit_with_name_factory("foo/bar/example.txt"),
+    }
+
+    units_to_add = list(map(lambda f: f.pulp_href, files.values()))
+    data = RepositoryAddRemoveContent(add_content_units=units_to_add)
+    monitor_task(
+        file_bindings.RepositoriesFileApi.modify(file_repo_with_auto_publish.pulp_href, data).task
+    )
+
+    build_response = container_repository_api.build_image(
```
lubosmj, line 93: Do you mind adding a test case that triggers users' permissions checks before building an image?

```
with user_helpless, pytest.raises(APIException):
    container_repository_api.build_image()
    
with user_manager:
    container_repository_api.build_image()

...
```

lubosmj (changes requested):

pulp_container/app/serializers.py
```diff
@@ -799,6 +811,20 @@ def validate(self, data):
...
+            repo_version_artifacts = RepositoryVersion.objects.get(pk=version_href.pk).artifacts
+            files = FileContent.objects.filter(
+                digest__in=repo_version_artifacts.values("sha256")
+            ).values("_artifacts__pk", "relative_path")
+            if len(files) == 0:
+                raise serializers.ValidationError(
+                    _("No file found for the specified repository version.")
+                )
+            for file in files:
+                artifacts[str(file["_artifacts__pk"])] = file["relative_path"]
```
lubosmj, line 826: Image building is still considered to be [tech-preview](https://pulpproject.org/pulp_container/docs/admin/learn/tech-preview/). Have you considered changing the current API to better fit our needs? Instead of passing a list of artifacts to the task, we can try passing only the reference to the repository version, after verifying user's permissions. The task will then unpack the repository version and run the suggested queries ^.

pulp_container/app/tasks/builder.py
```diff
@@ -124,6 +131,18 @@ def build_image_from_containerfile(
...
+        if repo_version:
+            artifacts = {}
+            repo_version_artifacts = RepositoryVersion.objects.get(pk=repo_version).artifacts
+            files = FileContent.objects.filter(
+                digest__in=repo_version_artifacts.values("sha256")
+            ).values("_artifacts__pk", "relative_path")
+            if len(files) == 0:
+                raise ValidationError(_("No file found for the specified repository version."))
+            for file in files:
+                artifacts[str(file["_artifacts__pk"])] = file["relative_path"]
```
lubosmj, line 144: It seems like we have the list of artifacts already available from `RepositoryVersion.objects.get(pk=repo_version).artifacts`. Have you considered iterating over this queryset instead of devising a new dictionary and then running another DB queries for getting the corresponding artifacts?

pulp_container/app/viewsets.py
```diff
@@ -946,8 +949,13 @@ def build_image(self, request, pk):
...
 
-        artifacts = serializer.validated_data["artifacts"]
-        Artifact.objects.filter(pk__in=artifacts.keys()).touch()
+        artifacts, repo_version = None, None
+        if serializer.validated_data.get("artifacts"):
+            artifacts = serializer.validated_data["artifacts"]
+            Artifact.objects.filter(pk__in=artifacts.keys()).touch()
+        elif serializer.validated_data.get("repo_version"):
+            repo_version = serializer.validated_data["repo_version"]
+            self._check_file_repo_permission(request, repo_version)
```
lubosmj, line 958: Instead of doing this, we may want to declare a new entry inside the access policy, like we did for the repository viewset:
https://github.com/pulp/pulp_container/blob/25abfd6cb0bff1ac8756e4d15c65b83839594f66/pulp_container/app/viewsets.py#L674-L682

pulp_container/tests/functional/api/test_build_images.py
```diff
@@ -29,15 +30,59 @@ def containerfile_name():
...
+    monitor_task(upload_task)
+
+    return container_repo, file_repo
+
+
+@pytest.fixture
+def build_image(container_repository_api):
+    def _build_image(repository, containerfile, artifacts=None, repo_version=None):
+        # workaround for drf-spectacular not allowing artifacts=None and raising an exception
+        # (a bytes-like object is required, not 'NoneType')
```
lubosmj, line 60: Does it mean that when I want to build a new image from a Containerfile that does not require any additional artifacts, I cannot use the python bindings?

CHANGES/479.feature
```diff
@@ -0,0 +1 @@
+Added a feature to allow using file `repository_version` as a building context of a Containerfile.
```
lubosmj, line 1:
```suggestion
Added support for using the file `repository_version` as the build context for a Containerfile.
```

pulp_container/app/tasks/builder.py
```diff
@@ -111,6 +116,8 @@ def build_image_from_containerfile(
                           container executing the Containerfile.
         repository_pk (str): The pk of a Repository to add the OCI container image
         tag (str): Tag name for the new image in the repository
+        repo_version: The pk of a RepositoryVersion with the artifacts used in the build context
```
lubosmj, line 119:
```suggestion
        repo_version: The href of a file RepositoryVersion with the artifacts used in the build context
```

lubosmj (changes requested):
Thank you for working on this! I left a couple of small suggestions. Also, I would still ditch the `artifacts` parameter from the `build` endpoint. What do you think, @ipanova, should we keep it or not?

pulp_container/app/tasks/builder.py
git-hyagi: Yes, but I could not find the `relative_path` (that will be used in the Containerfile) of each file in the *artifacts* queryset. Is there a better way to find it?

pulp_container/tests/functional/api/test_build_images.py
git-hyagi: I think so. Testing (from main branch) with:
```
    build_response = container_repository_api.build_image(
        repository.pulp_href, containerfile=containerfile_name, artifacts=None
    )
```

will fail:
```
E               TypeError: a bytes-like object is required, not 'NoneType'
```

pulp_container/app/tasks/builder.py
lubosmj: Also, you may want to verify whether the artifacts are in place or not. A user could sync a repository with the on-demand policy and then use it as the build context for pulp-container. We should be careful about this.

pulp_container/tests/functional/api/test_build_images.py
lubosmj: This is so unfortunate. Is there any chance that we provide an invalid openapi schema for the generation?

pulp_container/app/global_access_conditions.py
```diff
@@ -118,3 +120,17 @@ def has_distribution_perms(request, view, action, permission):
     return any(
         (request.user.has_perm(permission, distribution.cast()) for distribution in distributions)
     )
+
+
+def has_file_repo_perms(request, view, action, permission):
+    serializer = serializers.OCIBuildImageSerializer(
+        data=request.data, context={"request": request}
+    )
```
lubosmj, line 128: If we cannot make this more generic, we should rename it to something like "has_image_build_file_repo_perms".

pulp_container/app/tasks/builder.py
```diff
@@ -124,16 +123,16 @@ def build_image_from_containerfile(
...
-            artifact = Artifact.objects.get(pk=key)
-            dest_path = os.path.join(context_path, val)
-            dirs = os.path.split(dest_path)[0]
-            if dirs:
-                os.makedirs(dirs, exist_ok=True)
-            with open(dest_path, "wb") as dest:
-                shutil.copyfileobj(artifact.file, dest)
 
-            containerfile_path = os.path.join(working_directory, "Containerfile")
+        if repo_version:
```
lubosmj, line 127: The `repo_version` argument is now required.

CHANGES/479.removal
```diff
@@ -0,0 +1 @@
+Removed support for using raw artifacts as the build context for a Containerfile.
```
lubosmj, line 1: This is not needed.

pulp_container/app/global_access_conditions.py
git-hyagi: My bad!

CHANGES/479.feature
```diff
@@ -0,0 +1,2 @@
+Replaced `artifacts` by `repository_version` as the parameter to provide the build context
+for a Containerfile.
```
lubosmj, line 2: What kind of `repository_version`? pulp-container's one? 😜

staging_docs/admin/guides/build-image.md
```diff
@@ -41,12 +41,18 @@ CMD ["cat", "/inside-image.txt"]' >> Containerfile
 ## Build an OCI image
 
 ```bash
-TASK_HREF=$(http --form POST :$REPO_HREF'build_image/' containerfile@./Containerfile \
-artifacts="{\"$ARTIFACT_HREF\": \"foo/bar/example.txt\"}"  | jq -r '.task')
+ARTIFACT_SHA256=$(http :$ARTIFACT_HREF | jq -r '.sha256')
```
lubosmj, line 44: We no longer care about artifacts, do we?

pulp_container/app/global_access_conditions.py
```diff
@@ -118,3 +119,20 @@ def has_distribution_perms(request, view, action, permission):
...
     )
+
+
+def has_repository_perms(request, view, action, permission):
+    """
+    Check if the user has permissions on the corresponding repository.
+    """
+    obj = view.get_object() if view.detail else None
+    context = {"request": request}
+    serializer = view.serializer_class(instance=obj, data=request.data, context=context)
```
lubosmj, line 130: Help me understand this. This is invoking `ContainerRepositorySerializer` instead of `OCIBuildImageSerializer`? Is the `repo_version` always initialized during the serialization when provided? Or, is this function always returning `True`?

CHANGES/479.feature
git-hyagi: Oops ... good catch!

staging_docs/admin/guides/build-image.md
git-hyagi: Hum ... we don't, I kept this artifact just as a sample/doc to populate the repository:
https://github.com/pulp/pulp_container/pull/1687/files/010d8fb88c2361abbe8f4cefc5a6b62ba42308ee#diff-8d94fc0843f51b6f430b79c609efee42205be12f7fee3369c1dd7824eeabe779R47-R48

I'll update it to create the file content "directly".

pulp_container/app/global_access_conditions.py
git-hyagi: > Help me understand this. This is invoking ContainerRepositorySerializer instead of OCIBuildImageSerializer, correct?

Hum... I'm seeing the _view.serializer_class_ returning **OCIBuildImageSerializer** in the debugger.

> Is the repo_version always initialized during the serialization when provided?

Yes, when we provide a repo_version, the serializer will:
https://github.com/pulp/pulp_container/pull/1687/files/010d8fb88c2361abbe8f4cefc5a6b62ba42308ee#diff-1689ddc49f7d0e21a80ae6006966cfb47766de4a004c7bb1b8e7f442b285384eR752
https://github.com/pulp/pulp_container/pull/1687/files/010d8fb88c2361abbe8f4cefc5a6b62ba42308ee#diff-1689ddc49f7d0e21a80ae6006966cfb47766de4a004c7bb1b8e7f442b285384eR780-R781
```
    repo_version = RepositoryVersionRelatedField(...)
... 
        if "repo_version" in data:
        |   data["repo_version"] = data["repo_version"].pk
```

> Or, is this function always returning True?

If my tests were valid, it is returning _False_ if the user has no permission to the file repo.

Running the following test returned "_You do not have permission to perform this action._":
```
pulp user create --username test --password krokodyl1 --email test@test.com
pulp user role-assignment add --username test --role "container.containerdistribution_collaborator" --object ""
pulp user role-assignment add --username test --role "container.containerrepository_content_manager" --object ""
echo 'Hello world!' > /tmp/example.txt

FILE_REPO=$(pulp file repository create --name foo --autopublish | jq -r '.pulp_href')
pulp file content upload --repository foo --file /tmp/example.txt --relative-path example4.txt

CONTAINER_REPO=$(pulp container repository create --name build | jq -r '.pulp_href')
mkdir -p /tmp/test
echo 'FROM centos:7
COPY example4.txt /inside-image.txt
CMD ["cat", "/inside-image.txt"]' > /tmp/test/Containerfile

http -a test:krokodyl1 --form POST :5001${CONTAINER_REPO}build_image/ "containerfile@/tmp/test/Containerfile" repo_version=${FILE_REPO}versions/1/
```

and then it worked (returned "_202 Accepted_") with:
```
pulp user role-assignment add --username test --role "file.filerepository_viewer" --object ""
http -a test:krokodyl1 --form POST :5001${CONTAINER_REPO}build_image/ "containerfile@/tmp/test/Containerfile" repo_version=${FILE_REPO}versions/1/
```

staging_docs/admin/guides/build-image.md
lubosmj: Please do. We should forget about using the artifacts endpoint.

pulp_container/app/global_access_conditions.py
lubosmj: Okay, thanks for clarifying this. I thought that it would just take the default repository serializer. Since this works out of the box and it is generic enough, do you think we should move this access policy check to pulpcore?

git-hyagi: Hum... considering it could be helpful to other plugins I would vote yes, but considering the need to do a serialization I am not confident that it is a good approach (at least, not yet to be used by other plugins).
Do you think the serialization is fine and we should not be too worried about it?

lubosmj: I mean, something similar is already used in https://github.com/pulp/pulpcore/blob/92785d9e3ed917b8b0bec4c24c3669f594a81018/pulpcore/app/global_access_conditions.py#L146. But, I guess this whole discussion needs someone who has better expertise on this...

lubosmj:
This is blocked by https://github.com/pulp/pulpcore/pull/5618.

pulp_container/app/tasks/builder.py
```diff
@@ -166,3 +165,12 @@ def build_image_from_containerfile(
         repository_version = add_image_from_directory_to_repository(image_dir, repository, tag)
 
     return repository_version
+
+
+def _copy_file_from_artifact(context_path, relative_path, artifact):
+    dest_path = os.path.join(context_path, relative_path)
+    dirs = os.path.split(dest_path)[0]
```
lubosmj, line 172:
```suggestion
    dirs = os.path.dirname(dest_path)
```

lubosmj (dismissed):
Looks fine to me! Let's wait for other members to review this.

pulp_container/app/tasks/builder.py
```diff
@@ -124,16 +123,21 @@ def build_image_from_containerfile(
...
-            dest_path = os.path.join(context_path, val)
-            dirs = os.path.split(dest_path)[0]
-            if dirs:
-                os.makedirs(dirs, exist_ok=True)
-            with open(dest_path, "wb") as dest:
-                shutil.copyfileobj(artifact.file, dest)
 
-            containerfile_path = os.path.join(working_directory, "Containerfile")
+        if repo_version:
+            file_repository = RepositoryVersion.objects.get(pk=repo_version)
```
mdellweg, line 128: The naming here could also be more consistent:

```suggestion
            build_context = RepositoryVersion.objects.get(pk=build_context_pk)
```

pulp_container/app/tasks/builder.py
```diff
@@ -124,16 +123,21 @@ def build_image_from_containerfile(
...
+            content_artifacts = ContentArtifact.objects.filter(
+                content__in=file_repository.content
+            ).order_by("-content__pulp_created")
+            for content in content_artifacts.select_related("artifact").iterator():
```
mdellweg, line 132: Here too: using the name "content" is misleading since it's content artifact.
BTW: I think it would be better to loop over actual FileContent to be future proof.

CHANGES/479.feature
```diff
@@ -0,0 +1,2 @@
+Replaced `artifacts` by `repository_version` (i.e., a file plugin repository version HREF)
+as the parameter to provide the build context for a Containerfile.
```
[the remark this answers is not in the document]
git-hyagi: Yes, image building is still tech-preview. Should we add a removal snippet even this way?

pulp_container/app/tasks/builder.py
git-hyagi: Hum... even if with `FileContent` we would need to do more database hits?

mdellweg: What if we add a `filter(content__pulp_type="file")` to be sure?

CHANGES/479.feature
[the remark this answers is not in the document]
mdellweg: I'd say, yes.

docs/admin/guides/build-image.md
```diff
@@ -7,23 +7,24 @@
 
 Users can add new images to a container repository by uploading a Containerfile. The syntax for
 Containerfile is the same as for a Dockerfile. The same REST API endpoint also accepts a JSON
-string that maps artifacts in Pulp to a filename. Any artifacts passed in are available inside the
+string that maps artifacts in Pulp to a filename. Any file passed in are available inside the
 build container at `/pulp_working_directory`.
```
mdellweg, line 11: Is that still true? Didn't we say we remove the `artifacts` parameter and ask for a `build_context` instead?

pulp_container/app/serializers.py
```diff
@@ -774,11 +773,10 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
     tag = serializers.CharField(
```
mdellweg, line 773: Sorry, i cannot comment on 10 lines above ...
I think, we should not keep the `containerfile_artifact` as is.

pulp_container/app/serializers.py
```diff
@@ -774,11 +773,10 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
         required=False, default="latest", help_text="A tag name for the new image being built."
     )
-    artifacts = serializers.JSONField(
+    build_context = RepositoryVersionRelatedField(
         required=False,
-        help_text="A JSON string where each key is an artifact href and the value is it's "
-        "relative path (name) inside the /pulp_working_directory of the build container "
-        "executing the Containerfile.",
+        help_text=_("RepositoryVersion to be used as the build context for container images."),
+        allow_null=True,
```
mdellweg, line 779: We want this to be a FileRepositoryVersion, right? Can we do something about the queryset here?

pulp_container/app/tasks/builder.py
git-hyagi: Good idea!

pulp_container/app/serializers.py
git-hyagi: Hum... I could think of 2 ways to do that:
- provide a new field to pass the pulp_href of the File Content with the Containerfile (not sure if this could be considered an advantage, but with this approach users could use a different repository to store the Containerfile or pass the same as the build_context)  
- use the build_context field and add a logic to validate if the Containerfile is present in the specified File Repository

What do you think of these ideas?

pulp_container/app/serializers.py
git-hyagi: Hum... I tried to find a `FileRepositoryVersion` model, but I couldn't.
I also tried to pass a `FileRepository` qs, but it didn't work:
```
    build_context = RepositoryVersionRelatedField(
        required=False,
        help_text=_("RepositoryVersion to be used as the build context for container images."),
        allow_null=True,
        queryset = FileRepository.objects.all()
    )


django.core.exceptions.FieldError: Cannot resolve keyword 'number' into field. Choices are: alternatecontentsourcepath, autopublish, content, core_pulp_exporter, description, distributions, group_roles, manifest, name, next_version, pulp_created, pulp_domain, pulp_domain_id, pulp_id, pulp_labels, pulp_last_updated, pulp_type, pulpimporterrepository, remote, remote_id, repository_ptr, repository_ptr_id, repositorycontent, retain_repo_versions, uploads, user_hidden, user_roles, versions
```

I also tried to "adapt" the qs like (to do a reverse lookup of the repository version from repository):
```
        queryset = FileRepository.objects.all().repositoryversion_set,
```
but django complained:
```
AttributeError: 'QuerySet' object has no attribute 'repositoryversion_set'
```

Am I doing something wrong?

pulp_container/app/serializers.py
mdellweg: I think it's valuable to be able to use the same build-context with different Container files.

That being said, we could add a containerfile name field to select one file from the build context. Or, yes, we just allow to specify a file_content item for now. I don't know exactly what workflows we need to cover. We still (already) have the option to upload a file on the go, right? So adding more options feels not too bad.

pulp_container/app/serializers.py
mdellweg: Maybe it's not possible. But maybe again `RepositoryVersion.objects.filter(repository__pulp_type="file")`?

lubosmj (commented):
When I follow the documentation to try out the examples, step by step (the only change I made was to the base container, I used alpine:3.9), I receive the following error. Do you think we can fix that? The issue is that we are trying to save the manifest that already exists in the database.

```
TASK_HREF=$(http --form POST :5001$CONTAINER_REPO'build_image/' containerfile_name=MyContainerfile \
build_context=${FILE_REPO}versions/2/ | jq -r '.task')
```
```
[pulp]           | pulp [b8bd7f171ad44854bcd899be74b43c39]: pulpcore.tasking.tasks:INFO: Task 01915551-60c7-75f4-8e41-7a2bd80b724c failed (duplicate key value violates unique constraint "container_manifest_digest_d362b1f0_uniq"
[pulp]           | DETAIL:  Key (digest)=(sha256:87989b67cf10c8f3471fb7ec3bb959f92c6aa7d6fd8de6d935907e29d125e4d0) already exists.)
[pulp]           | pulp [b8bd7f171ad44854bcd899be74b43c39]: pulpcore.tasking.tasks:INFO:   File "/src/pulpcore/pulpcore/tasking/tasks.py", line 75, in _execute_task
[pulp]           |     result = func(*args, **kwargs)
[pulp]           | 
[pulp]           |   File "/src/pulp_container/pulp_container/app/tasks/builder.py", line 188, in build_image_from_containerfile
[pulp]           |     repository_version = add_image_from_directory_to_repository(image_dir, repository, tag)
[pulp]           | 
[pulp]           |   File "/src/pulp_container/pulp_container/app/tasks/builder.py", line 77, in add_image_from_directory_to_repository
[pulp]           |     manifest.save()
[pulp]           | 
[pulp]           |   File "/usr/lib64/python3.9/contextlib.py", line 79, in inner
[pulp]           |     return func(*args, **kwds)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django_lifecycle/mixins.py", line 139, in save
[pulp]           |     save(*args, **kwargs)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 814, in save
[pulp]           |     self.save_base(
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 877, in save_base
[pulp]           |     updated = self._save_table(
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 1020, in _save_table
[pulp]           |     results = self._do_insert(
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 1061, in _do_insert
[pulp]           |     return manager._insert(
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 87, in manager_method
[pulp]           |     return getattr(self.get_queryset(), name)(*args, **kwargs)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 1805, in _insert
[pulp]           |     return query.get_compiler(using=using).execute_sql(returning_fields)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1822, in execute_sql
[pulp]           |     cursor.execute(sql, params)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 67, in execute
[pulp]           |     return self._execute_with_wrappers(
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
[pulp]           |     return executor(sql, params, many, context)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute
[pulp]           |     return self.cursor.execute(sql, params)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/utils.py", line 91, in __exit__
[pulp]           |     raise dj_exc_value.with_traceback(traceback) from exc_value
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 89, in _execute
[pulp]           |     return self.cursor.execute(sql, params)
[pulp]           | 
[pulp]           |   File "/usr/local/lib/python3.9/site-packages/psycopg/cursor.py", line 97, in execute
[pulp]           |     raise ex.with_traceback(None)
```

image

docs/admin/guides/build-image.md
```diff
@@ -7,46 +7,70 @@
...
-## Build an OCI image
+
+## Build from a Containerfile uploaded during build request
+
+### Build an OCI image with the "local" Containerfile
 
 ```bash
-TASK_HREF=$(http --form POST :$REPO_HREF'build_image/' containerfile@./Containerfile \
-artifacts="{\"$ARTIFACT_HREF\": \"foo/bar/example.txt\"}"  | jq -r '.task')
+TASK_HREF=$(http --form POST :$CONTAINER_REPO'build_image/' "containerfile@./Containerfile" \
```
lubosmj, line 52: Can we somehow use a real port number here? Not sure how we use it across the documentation. Or, at least use an out-of-nowhere environment variable? I got bitten by this.

docs/admin/guides/build-image.md
```diff
@@ -7,46 +7,70 @@
 
 Users can add new images to a container repository by uploading a Containerfile. The syntax for
 Containerfile is the same as for a Dockerfile. The same REST API endpoint also accepts a JSON
-string that maps artifacts in Pulp to a filename. Any artifacts passed in are available inside the
-build container at `/pulp_working_directory`.
+string that maps artifacts in Pulp to a filename. Any files passed in (via `build_context`) are
+available inside the build container at the path defined in File Content `relative-path`.
```
mdellweg, line 11: I don't think this quite captures it. It's more like the files in the build_context repository version are available at there relative path in the build context of the container. Yes this sounds like a tautology (which in turn is a sign that we chose a proper name for it). So we should probably focus more on how the parameter `build_context` defines the actual "build context" the container image is created with.

Note: To make a file actually be present in the running container, you need to first have it in a file repository version, then use the latter as the `build_context` for the image builder feature. Then still, you need to have the `ADD <file> <location in container>` statement.

pulp_container/app/serializers.py
```diff
@@ -774,56 +772,40 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
-                    # Append the URL of missing Artifact to the error message
-                    e.detail[0] = "%s %s" % (e.detail[0], url)
-                    raise e
-        data["artifacts"] = artifacts
+
+        # the "has_repo_or_repo_ver_param_model_or_obj_perms" permission condition function expects
+        # a "repo" or "repository_version" arguments, so we need to pass "build_context" as
+        # "repository_version" to be able to validate the permissions
+        if data.get("build_context", None):
+            data["repository_version"] = data["build_context"]
```
mdellweg, line 807: Ohhh bugger!
I guess we cannot pass extra parameters to those functions (they are used to define the access policies).
One other option would be to roll our own permission condition. I'm not yet convinced we _should_ do that.

pulp_container/app/serializers.py
```diff
@@ -774,56 +772,40 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
-        "executing the Containerfile.",
+        help_text=_("RepositoryVersion to be used as the build context for container images."),
+        allow_null=True,
+        queryset=RepositoryVersion.objects.filter(repository__pulp_type="file.file"),
     )
 
     def __init__(self, *args, **kwargs):
         """Initializer for OCIBuildImageSerializer."""
         super().__init__(*args, **kwargs)
-        self.fields["containerfile_artifact"].required = False
```
mdellweg, line 787: Just make that function disappear completely now.

pulp_container/app/serializers.py
```diff
@@ -774,56 +772,40 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
 
     def validate(self, data):
         """Validates that all the fields make sense."""
         data = super().validate(data)
 
+        if bool(data.get("containerfile", None)) == bool(data.get("containerfile_name", None)):
+            raise serializers.ValidationError(
+                _("'containerfile' or 'containerfile_name' must be specified.")
```
mdellweg, line 792:
```suggestion
                _("Exactly one of 'containerfile' or 'containerfile_name' must be specified.")
```

pulp_container/app/tasks/builder.py
```diff
@@ -96,7 +96,11 @@ def add_image_from_directory_to_repository(path, repository, tag):
 
 
 def build_image_from_containerfile(
```
mdellweg, line 98: We may have an issue here.
You know tasks can be older than your current installation. We never required draining the task queue as a prerequisite for upgrading the codebase. So i wonder if we need to keep the `artifacts` parameter around. That however raises the next question, whether we still need to handle it properly or whether we use it to make the task fail with a good error message.
Since we are changing it almost completely, i guess the feature "image building" is in tech-preview. So the reach of this question is kind of limited.
@lubosmj your call.

pulp_container/app/serializers.py
```diff
@@ -774,56 +772,40 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
+            )
+
         if "containerfile" in data:
-            if "containerfile_artifact" in data:
-                raise serializers.ValidationError(
-                    _("Only one of 'containerfile' and 'containerfile_artifact' may be specified.")
-                )
             data["containerfile_artifact"] = Artifact.init_and_validate(data.pop("containerfile"))
```
mdellweg, line 796: I think we need to persist this artifact still. Also we need to handle the cases it was already in pulp and `touch` it.

pulp_container/app/tasks/builder.py
```diff
@@ -106,34 +110,52 @@ def build_image_from_containerfile(
...
+                _copy_file_from_artifact(
+                    context_path, content_artifact.relative_path, content_artifact.artifact.file
+                )
+
+        try:
+            containerfile
+        except NameError:
+            raise RuntimeError(
+                '"' + containerfile_name + '" containerfile not found in build_context!'
+            )
```
mdellweg, line 156: This is kind of hard to grasp.
I think we can assume that either we have a `containerfile_pk` or we have both the `containerfile_name` and the `buildcontext`.
So can we just add a check that `containerfile_name` is one of the `relative_path`s in the `build_context`?
I think we can even move that check into the serializer helping the user with an early error message.

pulp_container/app/tasks/builder.py
```diff
@@ -106,34 +110,52 @@ def build_image_from_containerfile(
...
+
+        containerfile_path = os.path.join(working_directory, "Containerfile")
 
         with open(containerfile_path, "wb") as dest:
             shutil.copyfileobj(containerfile.file, dest)
```
mdellweg, line 161: There might be a small optimization possible here in case the containerfile is already on disk as part of the build_context. What if we move this copy operation up to line 125 and carry only the `containerfile_path` around from there.
Sounds like it could make use of `copy_file_from_artifact` too.

pulp_container/app/viewsets.py
```diff
@@ -696,6 +696,7 @@ class ContainerRepositoryViewSet(
                 "condition": [
                     "has_model_or_obj_perms:container.build_image_containerrepository",
                     "has_model_or_obj_perms:container.view_containerrepository",
+                    "has_repo_or_repo_ver_param_model_or_obj_perms:file.view_filerepository",
```
mdellweg, line 699: Or maybe these functions _do_ allow to pass parameters.
(It would need a change in pulpcore first, right?)

pulp_container/app/serializers.py
mdellweg: I see now, it is persisted in the viewset. Whichever place is better to do so I don't know.

pulp_container/app/viewsets.py
```diff
@@ -938,25 +939,31 @@ def build_image(self, request, pk):
...
-        containerfile = serializer.validated_data["containerfile_artifact"]
-        try:
-            containerfile.save()
-        except IntegrityError:
-            containerfile = Artifact.objects.get(sha256=containerfile.sha256)
-            containerfile.touch()
+        containerfile_pk = None
+        if containerfile := serializer.validated_data.get("containerfile_artifact", None):
+            try:
+                containerfile.save()
```
mdellweg, line 945: Just a note at this point. This will raise some warnings in the log stream, I believe and some people are kind of sensitive to the that. We should be prepared to change the logic to "look first, try to save then, except look again."
Don't change anything just yet.

pulp_container/app/viewsets.py
```diff
@@ -938,25 +939,31 @@ def build_image(self, request, pk):
...
+            containerfile_pk = str(containerfile.pk)
+
         tag = serializer.validated_data["tag"]
+        containerfile_name = serializer.validated_data.get("containerfile_name", None)
 
-        artifacts = serializer.validated_data["artifacts"]
-        Artifact.objects.filter(pk__in=artifacts.keys()).touch()
+        build_context_pk = None
+        if build_context := serializer.validated_data.get("build_context", None):
+            build_context_pk = build_context.pk
```
mdellweg, line 956: We need to put the build contexts repository (not the repository version) into the shared_resources for this task.

pulp_container/app/viewsets.py
mdellweg: Actually, i wonder now if the containerfile should be a PulpTemporaryFile instead of an Artifact after all.
@lubosmj WDYT?

lubosmj: Actually, this is good idea. If a user uploads a raw containerfile, we should store it into a PulpTemporaryFile, so it will be accessible from a task easily, no artifacts involved.

pulp_container/app/tasks/builder.py
lubosmj: In my opinion, the safest thing to do would be to leave the old API untouched (including the ViewSet action), introduce a new one, and then deprecate the old one after 2 releases. Otherwise, we would leave `artifacts` in place forever. Since this functionality is flagged as "tech-preview" and we completely overhauled the API, I call this concern unimportant.

Personally, I would rather implement a safety feature in pulpcore that will detect potential changes of the function signature after unpickling the function arguments: https://github.com/pulp/pulpcore/blob/e2f03715de68736b78942e5fa3aca98de2c318f7/pulpcore/tasking/tasks.py#L70-L75.

pulp_container/app/tasks/builder.py
```diff
@@ -68,15 +73,13 @@ def add_image_from_directory_to_repository(path, repository, tag):
     manifest_digest = calculate_digest(bytes_data)
     manifest_text_data = bytes_data.decode("utf-8")
 
-    manifest = Manifest(
+    manifest, _ = Manifest.objects.update_or_create(
```
mdellweg, line 76: Now this needs some explanation. Also i think we cannot do this. Content is immutable by design. (That is like a result of the number one Pulp architecture decisions that content is supposed to be shared.)
Did you mean `get_or_create`?

pulp_container/app/tasks/builder.py
```diff
@@ -68,15 +73,13 @@ def add_image_from_directory_to_repository(path, repository, tag):
...
         digest=manifest_digest,
         schema_version=2,
         media_type=MEDIA_TYPE.MANIFEST_OCI,
         data=manifest_text_data,
     )
-    manifest.save()
-    tag = Tag(name=tag, tagged_manifest=manifest)
-    tag.save()
+    tag, _ = Tag.objects.update_or_create(name=tag, tagged_manifest=manifest)
```
mdellweg, line 82: Same here.

pulp_container/app/tasks/builder.py
```diff
@@ -43,7 +48,7 @@ def get_or_create_blob(layer_json, manifest, path):
             artifact=layer_artifact, content=blob, relative_path=layer_json["digest"]
         ).save()
     if layer_json["mediaType"] != MEDIA_TYPE.CONFIG_BLOB_OCI:
-        BlobManifest(manifest=manifest, manifest_blob=blob).save()
+        BlobManifest.objects.update_or_create(manifest=manifest, manifest_blob=blob)
```
mdellweg, line 51: And i think it applies here too.

pulp_container/app/tasks/builder.py
mdellweg: Well, it's kind of safe, since an old task would just fail to start up on a new codebase. And we'd be like: "Hey yeah, tech preview. Try starting again." :shrug: .
More dangerous are changes, where the task didn't change the signature, but interprets it's parameters in a different way.

pulp_container/app/tasks/builder.py
git-hyagi: Thank you for the review and explaining about `Content` immutability (I didn't know that)!
I was trying to fix an issue, [found by Lubos++](https://github.com/pulp/pulp_container/pull/1687#pullrequestreview-2240031768), where trying to reuse a Containerfile in different build tasks would fail.
Since `Content` must be immutable, I agree that `get_or_create` should be used in these cases.

pulp_container/app/serializers.py
```diff
@@ -774,65 +774,90 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
+            # expects a "repo" or "repository_version" arguments, so we need to pass "build_context"
+            # as "repository_version" to be able to validate the permissions
+            if data.get("build_context", None):
+                data["repository_version"] = data["build_context"]
+
+            # return early, we don't need the remaining validations to check the permissions
+            return data
+
+        if containerfile := data.get("containerfile", None):
+            temp_file = PulpTemporaryFile.objects.update_or_create(file=containerfile)[0]
```
lubosmj, line 809: Is it safe to run update on this?

pulp_container/app/serializers.py
```diff
@@ -774,65 +774,90 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
+            content_artifacts = ContentArtifact.objects.filter(
+                content__pulp_type="file.file", content__in=repository_version.content
+            ).order_by("-content__pulp_created")
+            for content_artifact in content_artifacts.select_related("artifact").iterator():
+                if not content_artifact.artifact:
                     raise serializers.ValidationError(
-                        _("Relative path cannot start with '/'. " "{0}").format(relative_path)
+                        "It is not possible to use File content synced with on-demand "
+                        "policy without pulling the data first."
+                    )
```
lubosmj, line 823: Do we want to wrap this around by the gettext call?

pulp_container/app/serializers.py
```diff
@@ -774,65 +774,90 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
+                containerfile_name = data.get("containerfile_name", None)
+                if containerfile_name and content_artifact.relative_path == containerfile_name:
+                    artifact = Artifact.objects.get(pk=content_artifact.artifact.pk)
```
lubosmj, line 826: We can already access the artifact via the `content_artifact.artifact` attribute.

pulp_container/app/serializers.py
```diff
@@ -774,65 +774,90 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
-        if "artifacts" in data:
-            for url, relative_path in data["artifacts"].items():
-                if os.path.isabs(relative_path):
+
+        if self._is_permission_function_request():
+            # the "has_repo_or_repo_ver_param_model_or_obj_perms" permission condition function
+            # expects a "repo" or "repository_version" arguments, so we need to pass "build_context"
+            # as "repository_version" to be able to validate the permissions
+            if data.get("build_context", None):
+                data["repository_version"] = data["build_context"]
```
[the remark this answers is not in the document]
git-hyagi: This is because `has_repo_or_repo_ver_param_model_or_obj_perms` expects only `repository` or `repository_version` :
https://github.com/pulp/pulpcore/blob/91d158741ca790cff65c1cfd53579ed1737648b4/pulpcore/app/global_access_conditions.py#L433-L485

to use `build_context` "directly" we would need to modify the `has_repo_or_repo_ver_param_model_or_obj_perms` function in pulpcore.

pulp_container/tests/functional/api/test_build_images.py
```diff
@@ -29,35 +24,216 @@ def containerfile_name():
...
+
+        if not build_context:
+            build_context = b""
+
+        build_response = container_repository_api.build_image(
+            container_container_repository_href=repository,
+            containerfile=containerfile,
+            containerfile_name=containerfile_name,
+            build_context=build_context,
+        )
```
lubosmj, line 63:
```suggestion
        build_response = container_repository_api.build_image(
            container_container_repository_href=repository,
            containerfile=containerfile,
            containerfile_name=containerfile_name or "",
            build_context=build_context or "",
        )
```

pulp_container/app/tasks/builder.py
```diff
@@ -117,26 +125,29 @@ def build_image_from_containerfile(
         image and tag.
 
     """
-    containerfile = Artifact.objects.get(pk=containerfile_pk)
+    if containerfile_tempfile_pk:
+        containerfile = PulpTemporaryFile.objects.get(pk=containerfile_tempfile_pk)
+    else:
+        containerfile = Artifact.objects.get(pk=containerfile_artifact_pk)
```
lubosmj, line 131:
```suggestion
    elif containerfile_artifact_pk:
        containerfile = Artifact.objects.get(pk=containerfile_artifact_pk)
    else:
        raise RuntimeError("Neither an artifact or temporary file for the Containerfile was specified.")
```

pulp_container/tests/functional/api/test_build_images.py
```diff
@@ -29,35 +24,216 @@ def containerfile_name():
...
+    populated_file_repo,
+):
+    """Test with a non-existing Containerfile in file repository."""
+    with pytest.raises(ApiException) as e:
+        build_image(
+            repository=container_repo.pulp_href,
+            containerfile_name="Non_existing_file",
+            build_context=f"{populated_file_repo.pulp_href}versions/2/",
+        )
+    assert "Could not find the Containerfile" in e.value.body
```
lubosmj, line 208: Can we also check that this returns an HTTP 400 error?

CHANGES/479.removal
```diff
@@ -0,0 +1 @@
+Removed support for using raw artifacts as the build context and for the Containerfile.
```
lubosmj, line 1: Might be worth mentioning that this concerns the image building feature.

CHANGES/479.feature
```diff
@@ -0,0 +1,4 @@
+Replaced `artifacts` by `build_context` (i.e., a file plugin repository version HREF)
+as the parameter to provide the build context for a Containerfile.
+Added the `containerfile_name` as the parameter to provide the relative-path of a File Content
+from the provided `build_context`.
```
lubosmj, line 4: Might be worth mentioning that this concerns the image building feature.

CHANGES/479.feature
```diff
@@ -0,0 +1,5 @@
+Updated the image building feature:
+* Replaced `artifacts` by `build_context` (i.e., a file plugin repository version HREF)
+as the parameter to provide the build context for a Containerfile.
+* Added the `containerfile_name` as the parameter to provide the relative-path of a File Content
+from the provided `build_context`.
```
lubosmj, line 5: @pedro-psb, is this bullet-point list in the changelog going to be properly rendered by mkdocs?

pedro-psb: If the bullet paragraph has a 4 space indentation it will render correctly.

```
Updated the image building feature:
    * Replaced `artifacts` by `build_context` (i.e., a file plugin repository version HREF)
    as the parameter to provide the build context for a Containerfile.
    * Added the `containerfile_name` as the parameter to provide the relative-path of a File Content
    from the provided `build_context`.
```

Otherwise it'll become a flat list.

docs/admin/guides/build-image.md
```diff
@@ -6,47 +6,79 @@
...
-build container at `/pulp_working_directory`.
+Containerfile is the same as for a Dockerfile.
 
-## Create a Repository
+To make a file available during the build process, we need to add it to a versioned file repository,
+then pass it as the `build_context` query parameter of the build container API, and finally specify the
+```
+ADD/COPY <file relative-path> <location in container>
+```
+instruction in the Containerfile.
```
lubosmj, line 16: 1. What is a file? We should re-phrase it to something like "To pass arbitrary files or artifacts to the image building context, the `build_context` property can be provided in the request payload, containing a reference to a file repository."
2. `build_context` is not a [query parameter](https://docs.oracle.com/en/cloud/saas/cx-commerce/21b/ccdev/rest-api-query-parameters.html). It is a "JSON property" in the request body, see [the difference](https://stackoverflow.com/a/42462714).

docs/admin/guides/build-image.md
```diff
@@ -6,47 +6,79 @@
...
+```
+
+
 !!! warning
 
-    Non-staff users, lacking read access to the `artifacts` endpoint, may encounter restricted
-    functionality as they are prohibited from listing artifacts uploaded to Pulp and utilizing
-    them within the build process.
+    File repositories synced with on-demand policy will not automatically pull the missing artifacts.
+    Trying to build using a file that is not yet pulled will fail.
```
lubosmj, line 84:
```suggestion
    File repositories synced with the on-demand policy will not automatically download the missing artifacts.
    Trying to build an image using a file that has not yet been downloaded will fail.
```

pulp_container/app/serializers.py
```diff
@@ -774,65 +772,80 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
+                _("A 'build_context' must be specified when 'containerfile_name' is present.")
             )
-        artifacts = {}
-        if "artifacts" in data:
-            for url, relative_path in data["artifacts"].items():
-                if os.path.isabs(relative_path):
+
+        # the "has_repo_or_repo_ver_param_model_or_obj_perms" permission condition function
+        # expects a "repo" or "repository_version" arguments, so we need to pass "build_context"
+        # as "repository_version" to be able to validate the permissions
```
lubosmj, line 798:
```suggestion
        # the "has_repo_or_repo_ver_param_model_or_obj_perms" permission condition function
        # expects a "repository" or "repository_version" arguments, so we need to pass "build_context"
        # as "repository_version" to be able to validate the permissions
```

I find this comment redundant. It would be better to open a new feature request in pulpcore and add a TODO comment referencing it.

pulp_container/app/serializers.py
```diff
@@ -774,65 +772,80 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
+                        _(
+                            "It is not possible to use File content synced with on-demand "
+                            "policy without pulling the data first."
+                        )
+                    )
+                containerfile_name = data.get("containerfile_name", None)
+                if containerfile_name and content_artifact.relative_path == containerfile_name:
+                    data["containerfile_artifact_pk"] = str(content_artifact.artifact.pk)
+
+                data["content_artifact_pks"].append(content_artifact.pk)
```
lubosmj, line 827: When serializing data and dispatching a task from the viewset, there is a small window where someone could remove a file repository and we would end up with a list of non-existing keys. The task might fail when reading these keys. However, I am still unsure it is worth moving this code back to the task.

lubosmj (commented):
🥇

lubosmj (approved):
This looks great!

CHANGES/479.feature
```diff
@@ -0,0 +1,5 @@
+Updated the image building feature:
+    * Replaced `artifacts` by `build_context` (i.e., a file plugin repository version HREF)
+    as the parameter to provide the build context for a Containerfile.
+    * Added the `containerfile_name` as the parameter to provide the relative-path of a File Content
+    from the provided `build_context`.
```
mdellweg, line 5: Reminder to all of us:
Towncrier does not do Markdown validation. So we need to be on the lookout if this produces a nice and readable changelog in the end.

docs/admin/guides/build-image.md
```diff
@@ -6,47 +6,78 @@
...
-TASK_HREF=$(http --form POST :$REPO_HREF'build_image/' containerfile@./Containerfile \
-artifacts="{\"$ARTIFACT_HREF\": \"foo/bar/example.txt\"}"  | jq -r '.task')
+TASK_HREF=$(http --form POST ${BASE_ADDR}${CONTAINER_REPO}'build_image/' "containerfile@./Containerfile" \
+build_context=${FILE_REPO}versions/1/ | jq -r '.task')
 ```
 
+!!! note
+
+    If `TMPFILE_PROTECTION_TIME` is set to 0 (default value), the automatic cleanup is disabled,
+    meaning all the uploaded Containerfile used for the builds will be kept in `FILE_UPLOAD_TEMP_DIR`.
```
mdellweg, line 61: That is something we should change.
Tmpfiles should only be left over when a task failed or was canceled. After successful completion, the task should really clean up the kitchen. Don't bother the user with janitorial work.

pulp_container/app/viewsets.py
```diff
@@ -937,26 +937,29 @@ def build_image(self, request, pk):
...
             exclusive_resources=[repository],
+            shared_resources=[build_context_repo],
             kwargs={
-                "containerfile_pk": str(containerfile.pk),
+                "containerfile_artifact_pk": containerfile_artifact_pk,
+                "containerfile_tempfile_pk": containerfile_tempfile_pk,
                 "tag": tag,
                 "repository_pk": str(repository.pk),
-                "artifacts": artifacts,
+                "content_artifact_pks": content_artifact_pks,
```
mdellweg, line 962: Oh, I was thinking we were going to pass the build context and the containerfile_name to the task.

pulp_container/app/serializers.py
```diff
@@ -774,65 +772,78 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
-            for url, relative_path in data["artifacts"].items():
-                if os.path.isabs(relative_path):
+
+        # TODO: this can be removed after https://github.com/pulp/pulpcore/issues/5786
+        if data.get("build_context", None):
+            data["repository_version"] = data["build_context"]
+
+        return data
+
+    def deferred_validation(self, data):
```
mdellweg, line 802: Deferred validation will only run, if we call validate from the task context.
Actually on this kind of Serializer, there is no deferred_validation at all. Look for the Content upload serializer hierarchy to see where it's implemented.

pulp_container/app/viewsets.py
mdellweg: Maybe we should keep the old task untouched to be deprecated and removed in a few releases and design a new one with a clean interface next to it.

lubosmj:
Is this PR ready for review?

mdellweg (commented):
Some cosmetic suggestions.

Overall, looks very good!

pulp_container/app/serializers.py
```diff
@@ -774,65 +773,77 @@ class OCIBuildImageSerializer(ValidateFieldsMixin, serializers.Serializer):
...
-        if "containerfile" in data:
-            if "containerfile_artifact" in data:
-                raise serializers.ValidationError(
-                    _("Only one of 'containerfile' and 'containerfile_artifact' may be specified.")
-                )
-            data["containerfile_artifact"] = Artifact.init_and_validate(data.pop("containerfile"))
-        elif "containerfile_artifact" in data:
-            data["containerfile_artifact"].touch()
-        else:
+        if bool(data.get("containerfile", None)) == bool(data.get("containerfile_name", None)):
```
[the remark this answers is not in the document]
git-hyagi: 🤣

docs/admin/guides/build-image.md
```diff
@@ -6,47 +6,72 @@
...
 ```
 
+
 !!! warning
 
-    Non-staff users, lacking read access to the `artifacts` endpoint, may encounter restricted
-    functionality as they are prohibited from listing artifacts uploaded to Pulp and utilizing
-    them within the build process.
+    File repositories synced with the on-demand policy will not automatically download the missing artifacts.
+    Trying to build an image using a file that has not yet been downloaded will fail.
```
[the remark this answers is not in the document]
git-hyagi: I think it would be a good improvement.
We could “force” (or automate) the download by Pulp if the artifacts were not found, or maybe do not allow the usage of on-demand repositories for the build machinery.

CHANGES/479.feature
```diff
@@ -0,0 +1,2 @@
+Support for utilizing raw artifacts, as the build context and for the Containerfile, during image
```
lubosmj, line 1: We have still duplicated changelog entries. Please, resvole this.

lubosmj (approved):

PR #1707, opened 2024-07-21
Update CI files for branch 2.20

pulpbot:


lubosmj (dismissed):

lubosmj:
We will need to update the branch protection rules. From now on, only the ready-to-ship check should be mandatory.

lubosmj (approved):

Issue #1703, opened 2024-07-18
Requesting blob that doesn't exist results in 500


ianballou:
**Version**
pulpcore-selinux-2.0.1-1.el9.x86_64
python3.11-pulp-ostree-2.3.0-1.el9.noarch
python3.11-pulp-deb-3.2.0-1.el9.noarch
python3.11-pulp-rpm-3.25.3-1.el9.noarch
python3.11-pulp-python-3.11.1-1.el9.noarch
rubygem-smart_proxy_pulp-3.3.0-1.el9.noarch
**python3.11-pulp-container-2.20.0-1.el9.noarch**
python3.11-pulp-glue-0.25.3-1.el9.noarch
python3.11-pulpcore-3.49.10-1.el9.noarch
python3.11-pulp-ansible-0.21.5-1.el9.noarch
python3.11-pulp-cli-0.25.3-1.el9.noarch

**Describe the bug**
GETing a blob that doesn't exist triggers a 500:

```
Jul 18 15:18:09 centos9-proxy-devel pulpcore-api[2800]: pulp [24486f0709c64d4f8c3ec0f626abcfc7]: django.request:ERROR: Internal Server Error: /v2/default_organization-precipitation-container_push_view-buttermilk_biscuits-arianna3/blobs/sha256:cd55268b264d4d04e2eb2a95520656af57b1a2f8f8924522c71ff8b6fbdff538
Jul 18 15:18:09 centos9-proxy-devel pulpcore-api[2800]: Traceback (most recent call last):
Jul 18 15:18:09 centos9-proxy-devel pulpcore-api[2800]:  File "/usr/lib/python3.11/site-packages/pulp_container/app/registry_api.py", line 979, in handle_safe_method
Jul 18 15:18:09 centos9-proxy-devel pulpcore-api[2800]:    blob = models.Blob.objects.get(digest=pk, pk__in=repository_version.content)
Jul 18 15:18:09 centos9-proxy-devel pulpcore-api[2800]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jul 18 15:18:09 centos9-proxy-devel pulpcore-api[2800]:  File "/usr/lib/python3.11/site-packages/django/db/models/manager.py", line 87, in manager_method
[...]

PR #1713, opened 2024-07-24
Handles an exception for non-existing repository

git-hyagi:
fixes: #1703

### DIFF
--- a/CHANGES/1703.bugfix
+++ b/CHANGES/1703.bugfix
@@ -1,0 +1,2 @@
+Fixed an issue causing an HTTP 500 error when a GET request for a non-existing
+blob was made to a distribution with only a `repository_version` set.
--- a/pulp_container/tests/functional/api/test_pull_content.py
+++ b/pulp_container/tests/functional/api/test_pull_content.py
@@ -28,7 +28,7 @@ from pulp_container.tests.functional.constants import (
     REGISTRY_V2_REPO_HELLO_WORLD,
     PULP_HELLO_WORLD_LINUX_TAG,
 )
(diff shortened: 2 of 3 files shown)


CHANGES/1703.bugfix
```diff
@@ -0,0 +1,3 @@
+Fixed an issue that triggered a `NoneType` exception during a GET request when
+a `distribution` was created with only the `repository_version` (without specifying
+a `repository`).
```
lubosmj, line 3:
```suggestion
Fixed an issue causing an HTTP 500 error when a GET request for a non-existing blob was made to a distribution with only a `repository_version` set.
```
How about this? It does not talk about Python types. Users do not care about them.

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -279,6 +279,19 @@ def test_pull_nonexistent_image(self):
...
+        Verify that a GET request to a nonexistent BLOB will be properly handled
+        instead of outputting a stacktrace.
+        """
+        blob_path = "/v2/{}/blobs/{}".format(self.distribution_with_repo.base_path, EMPTY_JSON)
+        non_existing_blob_url = urljoin(self.cfg.get_base_url(), blob_path)
+
+        auth = get_auth_for_url(non_existing_blob_url)
+        content_response = requests.get(non_existing_blob_url, auth=auth)
+        with self.assertRaises(requests.exceptions.HTTPError):
+            content_response.raise_for_status()
```
lubosmj, line 293: Can we check that this returns 404 instead of 500?

PR #1725, opened 2024-07-29
Ignore tokens and keys in examples

lubosmj:
Even though these files have been already deprecated, some checkers still care about older branches.

[noissue]


.gitleaks.toml
```diff
@@ -1,5 +1,5 @@
 [allowlist]
-  description = "Our test install exports a test only MINIO ACCESS KEY"
-  paths = [
+  files = [
```
ggainey, line 2: Needs to stay "paths" - "files" is for specific-filenames (and is going to be deprecated soon anyway)

.gitleaks.toml
```diff
@@ -1,5 +1,5 @@
...
     ".github/workflows/scripts/install.sh",
+    "docs/workflows/authentication.rst",
```
ggainey, line 4: I think in this specific case, we're better served with a `regexes = [ exact-key-here ]` entry for the not-a-secret itself. Esp since this file is in history-only and won't be modifying the key ever again...

ggainey (changes requested):

ggainey:
Note for future investigators - the token in question is an **example** of a **temporary token** generated by pulp - it's never been able to give access to anything other than the transient test system that originally generated it.

ggainey (approved):

Issue #1700, opened 2024-07-12
Error when trying to pull an image with old manifest from pull-through cache


git-hyagi:
**Describe the bug**
Trying to pull an image with old manifest schema using pull-through caching will fail in the first execution:
```
('pulp [1d3e7a1770b046159c5efc4561c64d4d]: ::ffff:127.0.0.1 - - [12/Jul/2024:13:00:54 +0000] "GET /v2/docker-cache/pulp/redis/manifests/latest HTTP/1.0" 302 0 "-" "containers/5.28.0 (github.com/containers/image)"',)
[2024-07-12 13:00:55 +0000] [4736] [ERROR] Error handling request
Traceback (most recent call last):
...
   File "/src/pulp_container/pulp_container/app/registry.py", line 452, in init_pending_content
     if manifest_data["mediaType"] in (MEDIA_TYPE.MANIFEST_V2, MEDIA_TYPE.MANIFEST_OCI)
 KeyError: 'mediaType'
```

**To Reproduce**
* create a pull-through distribution:
```
REMOTE_HREF=$(curl -sH 'content-type: application/json' -XPOST http://localhost:5001/pulp/api/v3/remotes/container/pull-through/ -d '{"name": "foo", "url": "https://quay.io"}' | jq .pulp_href)
curl -H 'content-type: application/json'  -XPOST http://localhost:5001/pulp/api/v3/distributions/container/pull-through/ -d '{"remote": '${REMOTE_HREF}', "name": "docker-cache", "base_path": "docker-cache"}'
```

* try to pull an image with an old manifest
```
podman pull pulp-container:5001/docker-cache/pulp/redis:latest --tls-verify=false
```

**Expected behavior**
It should not raise the `KeyError` exception and work on the first try.

**Additional context**
Trying to pull the same content again works.

PR #1726, opened 2024-08-01
Fix a pull-through cache error with old manifests

git-hyagi:
fixes: #1700


CHANGES/1700.bugfix
```diff
@@ -0,0 +1 @@
+Fixed an issue with the pull-through cache that was causing exceptions when pulling old manifests.
```
lubosmj, line 1:
```suggestion
Resolved an issue with the pull-through cache that was causing errors when retrieving v2 schema 1 manifests.
```

lubosmj (approved):

Issue #1759, opened 2024-09-02
Signed content not being synced


git-hyagi:
**Describe the bug**
Trying to sync filtered content with cosign signature, from a remote without sigstore, and `"signed_only": true` can fail sometimes.

**To Reproduce**
```
pulp container repository create --name foo
pulp container remote create --name foo --url "https://quay.io" --upstream-name=curl/curl --include-tags='["8.9.1","sha256-7dd57efcae8c9c2a611816151d731a02a31fab5ab9fb5e0ff877f43009944a51.sig"]'
curl -H 'content-type:application/json' -u<user>:<pass> -X POST ${BASE_ADDR}$(pulp container repository show --name foo |jq .pulp_href -r)sync/ -d '{"signed_only": true, "remote": '$(pulp container remote show --name foo |jq .pulp_href)'}'
```
The above execution can download the signature tag, the '8.9.1' tag, or sometimes both.

**Expected behavior**
It should always sync the tag and the corresponding signature.

**Additional context**
I think this is happening because of 2 things:
**1- random tag_name checking**
Here https://github.com/pulp/pulp_container/blob/0406fa242c07829d77064d4fe6843c04d5ea7cea/pulp_container/app/tasks/sync_stages.py#L155

we are verifying the tag_name defined as the last element from loop:
https://github.com/pulp/pulp_container/blob/0406fa242c07829d77064d4fe6843c04d5ea7cea/pulp_container/app/tasks/sync_stages.py#L126-L134

which is updated in a later instruction:
https://github.com/pulp/pulp_container/blob/0406fa242c07829d77064d4fe6843c04d5ea7cea/pulp_container/app/tasks/sync_stages.py#L171

which can cause problems because, depending on the order returned from _to_download_artifact_, we would be checking the tag_name from one artifact and the digest from another:
https://github.com/pulp/pulp_container/blob/0406fa242c07829d77064d4fe6843c04d5ea7cea/pulp_container/app/tasks/sync_stages.py#L146C1-L146C72
and mistakenly think that the tag has no corresponding signature. For example:
* if the artifact for the current loop iteration is a signature and tag_name is still defined as '8.9.1', we will match the following condition
```
                    if (
                        not (tag_name.endswith(".sig") and tag_name.startswith("sha256-"))
                        and f"sha256-{digest.removeprefix('sha256:')}.sig" not in tag_list
                    ):
```
and the signature will not be downloaded.

[...]

PR #1760, opened 2024-09-02
Fix a sync filtered cosign without sigstore issue

git-hyagi:
fixes: #1759

### DIFF
--- a/CHANGES/1759.bugfix
+++ b/CHANGES/1759.bugfix
@@ -1,0 +1,2 @@
+Resolved an issue where syncing filtered content with a `cosign` signature from a remote source lacking `Sigstore`,
+and with `signed_only: true` enabled, could occasionally fail.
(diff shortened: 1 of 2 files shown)


CHANGES/1759.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed an issue where trying to sync a filtered content with cosign signature, from a remote
+without sigstore, and defined as `signed_only: true` could fail sometimes.
```
lubosmj, line 2: Can we go for something along these lines?

```suggestion
Resolved an issue where syncing filtered content with a cosign signature from a remote source lacking Sigstore, and with signed_only: true enabled, could occasionally fail.
```

pulp_container/app/tasks/sync_stages.py
```diff
@@ -147,13 +147,14 @@ async def run(self):
                 content_data, raw_text_data, response = await artifact
 
                 digest = calculate_digest(raw_text_data)
+                tag_name = response.url.split("/")[-1]
```
lubosmj, line 150: Was this change necessary? It looks like the variable is only used on line 172. The code tends to be a lot more readable if you keep the declaration and the usage of the variable close together.

lubosmj (approved):

Issue #1729, opened 2024-08-05
Disallow removing pull-through remotes when referenced from a pull-through distribution


git-hyagi:
**Version**
```
    {                                                                                                                                                                                                                                                                                     
      "component": "core",                                                                                                                                                                                                                                                                
      "version": "3.57.0.dev",                                                                                                                                                                                                                                                            
      "package": "pulpcore",                                                                                                                                                                                                                                                              
      "module": "pulpcore.app",                                                                                                                                                                                                                                                           
      "domain_compatible": true                                                                                                                                                                                                                                                           
    },                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
    {                                                                                                                                                                                                                                                                                     
      "component": "container",                                                                                                                                                                                                                                                           
      "version": "2.22.0.dev",                                                                                                                                                                                                                                                            
      "package": "pulp-container",                                                                                                                                                                                                                                                        
      "module": "pulp_container.app",                                                                                                                                                                                                                                                     
      "domain_compatible": false                                                                                                                                                                                                                                                          
    }
```

**Describe the bug**
Removing a pull-through remote that is already associated with a pull-through distribution will cause errors during an image push:
https://github.com/pulp/pulp_container/pull/1727#discussion_r1704442586
```
...
   File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 637, in get
     raise self.model.DoesNotExist(
 pulp_container.app.models.ContainerDistribution.DoesNotExist: ContainerDistribution matching query does not exist.
...
   File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 637, in get
     raise self.model.DoesNotExist(
 pulp_container.app.models.ContainerPullThroughRemote.DoesNotExist: ContainerPullThroughRemote matching query does not exist.
```

**To Reproduce**
```
REMOTE_HREF=$(curl -sH 'content-type: application/json' -uadmin:password -XPOST http://localhost:5001/pulp/api/v3/remotes/container/pull-through/ -d '{"name": "foo", "url": "https://quay.io"}' | jq .pulp_href)
curl -H 'content-type: application/json' -uadmin:password -XPOST http://localhost:5001/pulp/api/v3/distributions/container/pull-through/ -d '{"remote": '${REMOTE_HREF}', "name": "quay-cache", "base_path": "quay-cache"}'
[...]

PR #1768, opened 2024-09-13
Add a check for pull-through remote removal

git-hyagi:
If a pull-through remote is linked with a distribution, warn users that removing it can cause problems to pull non-cached content.

closes: #1729

### DIFF
--- a/pulp_container/app/viewsets.py
+++ b/pulp_container/app/viewsets.py
@@ -511,6 +511,25 @@ class ContainerPullThroughRemoteViewSet(RemoteViewSet, RolesMixin):
         ],
     }
 
+    @extend_schema(
+        description="Trigger an asynchronous delete task",
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def destroy(self, request, pk, **kwargs):
+        """
+        Delete a model instance
+        """
+        # check if there is no distribution associated with this remote before removing it
+        linked_distribution = models.ContainerPullThroughDistribution.objects.filter(remote_id=pk)
+        if linked_distribution.exists():
+            log.info(
+                f"The pull-through distribution {linked_distribution[0].name} has a reference to "
+                f"the remote being removed ({request._request.path}). "
+                f"After removing it, trying to pull content from {linked_distribution[0].base_path}"
+                " pull-through distribution can lead to errors."
+            )
+        return super().destroy(request, pk, **kwargs)
+
 
 class TagOperationsMixin:
     """
--- a/CHANGES/1729.feature
+++ b/CHANGES/1729.feature
@@ -1,0 +1,2 @@
+Added an informational message to alert users that removing a remote pull-through associated with
+a distribution may lead to problems when trying to pull non-cached content.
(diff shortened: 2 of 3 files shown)


pulp_container/app/viewsets.py
```diff
@@ -511,6 +511,25 @@ class ContainerPullThroughRemoteViewSet(RemoteViewSet, RolesMixin):
...
+        responses={202: AsyncOperationResponseSerializer},
+    )
+    def destroy(self, request, pk, **kwargs):
+        """
+        Delete a model instance
+        """
+        # check if there is no distribution associated with this remote before removing it
+        linked_distribution = models.ContainerPullThroughDistribution.objects.filter(remote_id=pk)
+        if linked_distribution.exists():
+            log.info(
```
mdellweg, line 525: This is basically too late. When this gets logged we still destroy the thing, so the damage has been done.
What if instead we mark the FK-relation `delete.PROTECT` instead?
Then the user needs to unset the remote on the distribution first. And that should not be an accidental step.

git-hyagi: When I opened the issue I thought that it would be fine to block the remote removal, then, when I started to work on the PR, I changed my mind and thought that blocking the removal and "forcing" users to first remove the associated distribution would be kind of "annoying" for users who really knows what they are doing and wanted to delete it.


>  _What if instead we mark the FK-relation delete.PROTECT instead?_

hum... to do so we would need to change the `remote` attribute from pulpcore Distribution which would impact all the other plugins, right?
https://github.com/pulp/pulpcore/blob/1672b3f405170d0a45de9c4b8cc2b059a066a162/pulpcore/app/models/publication.py#L659

I tried to override it in pulp_container
https://github.com/pulp/pulp_container/blob/bb7e18c090f50e75387cae2fbbe6f3025f901ae1/pulp_container/app/models.py#L688
but it seems like django does not allow it:
https://docs.djangoproject.com/en/5.1/topics/db/models/#field-name-hiding-is-not-permitted

> _"In normal Python class inheritance, it is permissible for a child class to override any attribute from the parent class. In Django, this isn’t usually permitted for model fields. If a non-abstract model base class has a field called author, you can’t create another model field or define an attribute called author in any class that inherits from that base class."_

mdellweg: > hum... to do so we would need to change the `remote` attribute from pulpcore Distribution which would impact all the other plugins, right?

That may not be the worst idea...

git-hyagi:
From pulpcore PR https://github.com/pulp/pulpcore/pull/2836, the safeguard of removing remotes seems to be a complex issue that involves more than just modifying the ON_DELETE constraint or adding a log message. I'm closing this PR and pulp_container issue #1729, since this should be handled by pulpcore https://github.com/pulp/pulpcore/issues/1975

lubosmj:
How does the removal of a distribution relate to the removal of a remote? The issue in the referenced PR was that we could not clearly evaluate all possible references from the remote and display them to the user when she was trying to delete it.

Issue #1767, opened 2024-09-13
Retrieve manifest image size and architecture from API


ianballou:
**Is your feature request related to a problem? Please describe.**
Katello would like to begin showing container image sizes and related architecture information. 
We'd also like to show more info about the type of image, but that seems to be covered by https://github.com/pulp/pulp_container/issues/1751

**Describe the solution you'd like**
- A new field for compressed image size on manifests
- A new field for architecture on manifests

**Describe alternatives you've considered**
None so far.

**Additional context**
This is part of an effort for Katello to revamp its container registry UI. Our goal is to show the information that a user would expect from a general container registry UI. If there is other data not yet in Pulp that would be helpful in a UI, lets discuss that here as well.

PR #1782, opened 2024-09-25
Add manifest arch, os, and compressed layers size fields

git-hyagi:
closes: #1767


lubosmj (changes requested):
A couple of remarks:

1. If we need to store the "architecture" and "os" fields on the Manifest model, we should try to fetch the values from a manifest list first. An OCI Index has the "architecture" and "os" fields required. There is no need to read the data from a config blob if a manifest is listed within an index (https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions).
2. Currently, there are three separate migrations. I would like to squash them. Every migration takes time to run some boilerplate which is transparent to us.
3. Currently, there are three commits referencing the same issue. I would like to squash them. Such a separation is not necessary.
4. If we decided to create another django-admin command, we should align with Katello to see if it makes sense for them to run a management command reading data from the storage.

ianballou:
> 4\. If we decided to create another django-admin command, we should align with Katello to see if it makes sense for them to run a management command reading data from the storage.

From the Katello side, we're okay with the command being the same as before or a new one. Choose whatever is most efficient from the Pulp side.

ianballou:
Also I wanted to clarify to be super sure -- OCI Image Index manifests will always have a `null` arch, os, and size?

pulp_container/app/management/commands/container-repair-manifest-metadatas.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def update_manifests(self, manifests_qs):
+        manifests_updated_count = 0
+        manifests_to_update = []
+        for manifest in manifests_qs.iterator():
+            # suppress non-existing/already migrated artifacts and corrupted JSON files
+            with suppress(ObjectDoesNotExist, JSONDecodeError):
+                manifest_data = json.loads(manifest.data)
+                manifest.init_architecture(manifest_data)
+                manifest.init_compressed_layers_size(manifest_data)
```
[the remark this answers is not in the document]
git-hyagi: This is not a migration, we can update this command as we update the model itself.

git-hyagi:
> Also I wanted to clarify to be super sure -- OCI Image Index manifests will always have a `null` arch, OS, and size? I'm not super clear since, according to https://specs.opencontainers.org/image-spec/image-index/?v=v1.0.1, the optional platform property can have an arch and an OS,

Thank you for bringing this!
We reviewed and discussed the `os` and `architecture` fields from manifest list, and we will work on it in this PR.
If the manifest list (or oci index) contains the `platform` (an optional) field, we will populate pulp manifest with `arch` and `os` (whenever `platform` is defined, `os` and `arch` are required).

git-hyagi:
After re-reading the specs, `manifestlist` or `oci-index` do not have a `platform` field. Actually, `platform` is a field from `manifests`. In this link, we can see it better: https://github.com/opencontainers/image-spec/blob/main/image-index.md (`platform` is a bullet inside `manifests`).

So, please, ignore my last comment and yes, "_OCI Image Index manifests will always have a null arch, OS, and size_".

ianballou:
> After re-reading the specs, I realized that `manifestlist` or `oci-index` do not have a `platform` field. Actually, `platform` is a field from `manifests`. In this link, we can see it better: https://github.com/opencontainers/image-spec/blob/main/image-index.md (`platform` is a bullet inside `manifests`).
> 
> So, please, ignore my last comment and yes, "_OCI Image Index manifests will always have a null arch, OS, and size_".

Nice catch! I didn't notice it was under manifests, so that's perfect.

git-hyagi:
> _If we need to store the "architecture" and "os" fields on the Manifest model, we should try to fetch the values from a manifest list first. An OCI Index has the "architecture" and "os" fields required. There is no need to read the data from a config blob if a manifest is listed within an index (https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions)._

I did some investigation in each workflow, and here are my findings:
* for the **sync** and **pull-through** tasks, the first `declarative content` that we pass to the next pipeline stage is/are `blobs`/`configblobs`
https://github.com/pulp/pulp_container/blob/1a8fe634dba25b6adacd17459da3be931e37ae74/pulp_container/app/tasks/sync_stages.py#L330-L345
  * after that, in the `resolve_flush` method, we then send the `manifests` and `manifest lists` in this order:
https://github.com/pulp/pulp_container/blob/1a8fe634dba25b6adacd17459da3be931e37ae74/pulp_container/app/tasks/sync_stages.py#L274-L306

* for the **push** workflow, we first push the `blobs` and the corresponding `manifest` for that image, and, just after pushing all images, the `manifest-list`/`oci-index` is pushed


I couldn't find how to fetch these values from the manifest list in advance. I'm not sure if I overlooked something or misunderstood the code workflow.

lubosmj:
- Mirroring:
  I believe you can extract the information about os/architecture inside the `create_listed_manifest` method: https://github.com/pulp/pulp_container/blob/1a8fe634dba25b6adacd17459da3be931e37ae74/pulp_container/app/tasks/sync_stages.py#L480.

- Pushing:
  When it comes to pushing, we may want to get this information from the already uploaded config blob: https://github.com/pulp/pulp_container/blob/1a8fe634dba25b6adacd17459da3be931e37ae74/pulp_container/app/registry_api.py#L1295.

  If we are fine (we are mostly not) with updating existing manifests, we could potentially update the objects here: https://github.com/pulp/pulp_container/blob/1a8fe634dba25b6adacd17459da3be931e37ae74/pulp_container/app/registry_api.py#L1221.

git-hyagi:
Thank you for the help and the suggestions/optimizations!

ianballou:
cc @sjha4 @qcjames53 we should keep an eye on this and integrate with it in Katello sooner rather than later to avoid excess reindexing of container manifests.

pulp_container/app/models.py
```diff
@@ -103,6 +108,9 @@ class Manifest(Content):
 
     annotations = models.JSONField(default=dict)
     labels = models.JSONField(default=dict)
+    architecture = models.TextField(null=True)
+    os = models.TextField(null=True)
+    compressed_layers_size = models.IntegerField(null=True)
```
ipanova, line 113: how about compressed_image_size

lubosmj:
@git-hyagi, please, rebase this PR against the main branch and prepare it for the final review.

pulp_container/app/registry_api.py
```diff
@@ -1300,6 +1300,8 @@ def put(self, request, path, pk=None):
             config_blob = found_config_blobs.first()
             manifest = self._init_manifest(manifest_digest, media_type, raw_text_data, config_blob)
             manifest.init_metadata(manifest_data=content_data)
+            manifest.init_architecture_and_os()
+            manifest.init_compressed_image_size()
```
mdellweg, line 1304: Do we now always need to call them adjacent to init_metadata?
Would it be useful to combine them into one to prevent holding it wrong?

pulp_container/app/tasks/sync_stages.py
```diff
@@ -629,19 +640,19 @@ def _post_save(self, batch):
...
+                        manifest_list=manifest_dc.content,
+                        image_manifest=dc.content,
                     )
+                    if platform := listed_manifest.get("platform"):
+                        manifest_list_manifest.architecture = platform["architecture"]
+                        manifest_list_manifest.os = platform["os"]
+                        manifest_list_manifest.features = platform.get("features")
+                        manifest_list_manifest.variant = platform.get("variant")
+                        manifest_list_manifest.os_version = platform.get("os.version")
+                        manifest_list_manifest.os_features = platform.get("os.features")
```
mdellweg, line 653: This is at least the third time i'm seeing this pattern now.

Maybe a dataclass together with `asdict` can help here? (Maybe not in this PR.)
https://docs.python.org/3/library/dataclasses.html

mdellweg (commented):
I believe the migration is zero downtime safe. And also i don't see any code failing before the fields get populated (The results may be incomplete, but you are supposed to run the command anyway.).
None of the comments should be blocking here.

pulp_container/app/registry_api.py
git-hyagi: hmm... during the build and mirroring tasks these fields are not handled together because we don't have all the information to set them at the same time (they are gathered from different resources). So I don't know if merging them would be straightforward.

mdellweg: Yeah, let's call it food for thought.

mdellweg (dismissed):

pulp_container/app/registry_api.py
git-hyagi: We discussed this and agreed on opening a task issue to investigate this later.

pulp_container/app/tasks/sync_stages.py
git-hyagi: We discussed this and agreed on opening a task issue to investigate this later.

pulp_container/app/registry_api.py
```diff
@@ -1219,17 +1219,17 @@ def put(self, request, path, pk=None):
...
-                    os_version=platform.get("os.version", ""),
-                    os_features=platform.get("os.features", ""),
                 )
+                if platform := manifests[manifest.digest]:
+                    manifest_to_list.architecture = (platform["architecture"],)
+                    manifest_to_list.os = (platform["os"],)
+                    manifest_to_list.features = (platform.get("features", ""),)
+                    manifest_to_list.variant = (platform.get("variant", ""),)
+                    manifest_to_list.os_version = (platform.get("os.version", ""),)
+                    manifest_to_list.os_features = (platform.get("os.features", ""),)
```
lubosmj, line 1232: I do not understand the purpose of the encapsulating brackets here. What is it needed for? This does not look good to me.
```
>>> platform = {"os": "blabla", "architecture": "amd"}
>>> (platform.get("architecture"),)
('amd',)
```
```
In [16]: ManifestListManifest.objects.filter().update(os=("amd",))
Out[16]: 1

In [17]: ManifestListManifest.objects.get().os
Out[17]: "('amd',)"
```

pulp_container/tests/functional/conftest.py
```diff
@@ -474,3 +474,14 @@ def _check_manifest_fields(**kwargs):
...
     return _check_manifest_fields
+
+
+@pytest.fixture
+def check_manifest_arch_os_size():
+    def _check_manifest_arch_os_size(manifest):
+        manifests = manifest.to_dict()["results"]
+        assert any("amd64" in manifest["architecture"] for manifest in manifests)
+        assert any("linux" in manifest["os"] for manifest in manifests)
+        assert any(manifest["compressed_image_size"] > 0 for manifest in manifests)
```
lubosmj, line 485: Your tests are passing on `"('amd64',)"` because `"amd64"` is really stored inside `"('amd64',)"`. Try matching the exact value, please.

pulp_container/app/tasks/sync_stages.py
lubosmj: I do not understand why this cannot be addressed in this PR. Also, notice how in some cases you are initializing `os_version` as `p.get("os.version")` and in some cases as `p.get("os.version", "")`. Not to mention that you even do `(p.get("os.version", ""),)`. I am not a fan of such discrepancies. Please, resolve this.

lubosmj (changes requested):

pulp_container/app/tasks/sync_stages.py
git-hyagi: I tried to use the `@dataclass` decorator, but I had trouble to make it work (probably because of the `image_manifest` and `manifest_list` foreignkey fields).

lubosmj (approved):

Issue #1751, opened 2024-08-30
As a user I want to differentiate cosign signatures when looking at the image manifests


ipanova:
**Is your feature request related to a problem? Please describe.**

it might be beneficial to differentiate cosign signature wrapped into a manifest from the rest of the images. It is a similar ask to https://github.com/pulp/pulp_container/issues/1437, as result the manifest will be marked of `cosign signature` type.

Right now we have 2 fields on the maifest model `is_bootable` and `is_flatpak`. Does it make sense to consolidate them into a single field `type` and have `bootable`, `flatpak`, `cosign signature`, `artifact`, `unknown` ( for still unidentified types, that we can gradually add in future with incoming requests).

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Add any other context about the feature request here.

lubosmj:
I am leaning towards the idea of having a single type with different allowed values. We will need to run a migration (that will potentially read config blobs (again)). A new CharField filed (without choices constraints) will be defined on the Manifest table (+type artifact, image, volume, ...).

ipanova:
The only 2 examples i am aware of that would require reading from config blob and look at labels is flatpak and bootable containers. We already have performed a migration that would extact this information into `is_flatpack` and `is_bootable` fields. So this time you will look at the field directly and not at the parsed config.json file.

For the rest of 'types' we would look at 'mediaType' ( on config/layer) and 'artifactType' which is set on the manifest.json file

lubosmj:
Is there going to be defined a mapping between mediaTypes and types? Are we the ones who define that? Or, are we going to introduce a setting for that like we did for allowed artifact types?

ianballou:
From a Pulp user standpoint, having a single type field makes a lot of sense. An image won't be flatpak, bootable, and an artifact at the same time.
In Katello we're looking to also start differentiating if an image is an artifact or not, so we could potentially take this new field and use it directly in our API/UI/etc.

PR #1794, opened 2024-10-10
Add the `type` field to the Manifest model

git-hyagi:
closes: #1751


lubosmj (changes requested):
When should we deprecate `is_flatpak` and `is_bootable` in favour of `type/nature`? Is it fine to deprecate them now?

pulp_container/app/models.py
```diff
@@ -164,17 +171,41 @@ def init_manifest_list_nature(self):
...
+
+                # set custom attributes (is_flatpak, is_bootable) to keep compatibility
+                if manifest_type.get("custom", None):
+                    custom_key_name = list(manifest_type["custom"])[0]
+                    custom_key_value = manifest_type["custom"][custom_key_name]
+                    setattr(self, custom_key_name, custom_key_value)
+
+                return True
+
+        return False
```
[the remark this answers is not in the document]
lubosmj: This is not possible because we cannot determine the type of the image from `media_type`. For some of the natures, we still need to look into the config blob (e.g., for helm charts). 🧷

git-hyagi:
> _When should we deprecate `is_flatpak` and `is_bootable` in favour of `type/nature`? Is it fine to deprecate them now?_

Well remembered! I talked to the Katello's team, and they are fine with deprecating these fields in favor of the new `type/nature`.

pulp_container/constants.py
```diff
@@ -71,3 +73,12 @@
...
 SIGNATURE_API_EXTENSION_VERSION = 2
+
+MANIFEST_TYPE = SimpleNamespace(
+    IMAGE="image",
+    BOOTABLE="bootable",
+    FLATPAK="flatpak",
+    HELM="helm",
+    SIGNATURE="signature",
+    UNKNOWN="unknown",
+)
```
mdellweg, line 84: These types are mutually exclusive, right?

pulp_container/app/tasks/sync_stages.py
```diff
@@ -648,4 +649,4 @@ def _post_save(self, batch):
         # it is possible to initialize the nature of the corresponding manifest lists
         for ml in manifest_lists:
             if ml.init_manifest_list_nature():
-                ml.save(update_fields=["is_bootable", "is_flatpak"])
+                ml.save(update_fields=["is_bootable", "is_flatpak", "nature"])
```
mdellweg, line 652: _post_save runs in the same transaction as the things are created, right?

I still wonder, can we not know all this before we even save the manifests the first time?

pulp_container/app/registry_api.py
```diff
@@ -1238,7 +1238,7 @@ def put(self, request, path, pk=None):
             # once relations for listed manifests are established, it is
             # possible to initialize the nature of the manifest list
             if manifest.init_manifest_list_nature():
-                manifest.save(update_fields=["is_bootable", "is_flatpak"])
+                manifest.save(update_fields=["is_bootable", "is_flatpak", "nature"])
```
mdellweg, line 1241: Not your fault, apparently, but this appears to be unsafe. (It mutates content, which is both wrong in general and prone to creating deadlocks.)
However, i believe we need a transaction here anyway, because we need to save the manifest list before we can attach the listed manifests, but having an "empty" manifest list committed into the database is an invalid state.
I think the transaction needs to start before the manifest_list safe call.

We may think about this in a different PR though.

pulp_container/app/models.py
```diff
@@ -154,39 +159,85 @@ def init_image_nature(self):
...
+
+    @validate_json_field("layers")  # Check for 'layers' field for cosign images
+    def is_cosign(self, json_manifest):
+        return any(
+            layer.get("mediaType", None) == MEDIA_TYPE.COSIGN for layer in json_manifest["layers"]
+        )
+
+    @validate_json_field("config")  # Check for 'config' field for helm images
+    def is_helm_image(self, json_manifest):
+        return json_manifest.get("config").get("mediaType") == MEDIA_TYPE.HELM
```
mdellweg, line 240: So `manifest.data` needs to be a string field, because different formatting would bust the manifest checksums, right? But we can safely assume it contains valid json.
Let me suggest a different (maybe more pythonic) approach here:
We can add a `json_manifest` property.
```python
@property
def json_manifest(self):
    return json.loads(self.data)
```

Then we can do:
```python
def is_helm_image(self):
    # should this actually be `is_helm_chart`?
    try:
        return self.json_manifest["config"]["mediaType"] == MEDIA_TYPE.HELM
    except KeyError:
        return False
```

mdellweg: Or just
```python
def is_helm_image(self):
    # should this actually be `is_helm_chart`?
    try:
        return json.loads(self.data)["config"]["mediaType"] == MEDIA_TYPE.HELM
    except KeyError:
        return False
```
in the first place.

pulp_container/app/models.py
```diff
@@ -154,39 +159,85 @@ def init_image_nature(self):
...
+            MANIFEST_TYPE.IMAGE: self.is_manifest_image,
+        }
 
     def is_bootable_image(self):
         if (
             self.annotations.get("containers.bootc") == "1"
             or self.labels.get("containers.bootc") == "1"
         ):
+            # DEPRECATED: is_bootable is deprecated and will be removed in a future release.
+            self.is_bootable = True
```
mdellweg, line 204: This sounds like a weird side effect of a function that claims to only calculate whether a condition is met.

pulp_container/app/models.py
```diff
@@ -154,39 +159,85 @@ def init_image_nature(self):
...
+        return False
+
+    def known_types(self):
+        return {
+            MANIFEST_TYPE.BOOTABLE: self.is_bootable_image,
+            MANIFEST_TYPE.FLATPAK: self.is_flatpak_image,
+            MANIFEST_TYPE.HELM: self.is_helm_image,
+            MANIFEST_TYPE.SIGNATURE: self.is_cosign,
+            MANIFEST_TYPE.IMAGE: self.is_manifest_image,
+        }
```
mdellweg, line 196: I'd prefer the if elif structure, even for readability.

pulp_container/app/models.py
```diff
@@ -154,39 +159,85 @@ def init_image_nature(self):
...
     def init_manifest_list_nature(self):
+        updated_nature = False
+        if not self.nature:
+            self.nature = MANIFEST_TYPE.UNKNOWN
+            updated_nature = True
+
         for manifest in self.listed_manifests.all():
             # it suffices just to have a single manifest of a specific nature;
             # there is no case where the manifest is both bootable and flatpak-based
             if manifest.is_bootable:
```
mdellweg, line 170: If I understand this correctly, you are relying on the deprecated fields here.
If you are doing this for zero downtime upgrade reasons, please add a big comment explaining the line of thought. If not, I think it needs to be rewritten.

pulp_container/constants.py
```diff
@@ -71,3 +73,12 @@
 SIGNATURE_PAYLOAD_MAX_SIZE = 4 * MEGABYTE
 
 SIGNATURE_API_EXTENSION_VERSION = 2
+
+MANIFEST_TYPE = SimpleNamespace(
```
mdellweg, line 77: Should we call this MANIFEST_NATURE?

mdellweg (commented):
Lots of suggestions here...

CHANGES/1751.feature
```diff
@@ -0,0 +1,2 @@
+Introduced the `nature` field on the Manifests endpoint to enable easier differentiation of image
```
ipanova, line 1: I suggest going with simple 'type'

pulp_container/app/models.py
```diff
@@ -154,39 +159,85 @@ def init_image_nature(self):
...
     def is_flatpak_image(self):
-        return True if self.labels.get("org.flatpak.ref") else False
+        if self.labels.get("org.flatpak.ref"):
+            # DEPRECATED: is_flatpak is deprecated and will be removed in a future release.
+            self.is_flatpak = True
+            return True
+        return False
+
+    def is_manifest_image(self):
+        return self.media_type in (MEDIA_TYPE.MANIFEST_OCI, MEDIA_TYPE.MANIFEST_V2)
```
ipanova, line 217: what about schema1? we stopped conversion on the fly but still support its mirroring from remote registries.

pulp_container/constants.py
```diff
@@ -19,6 +19,8 @@
     FOREIGN_BLOB_OCI_TAR_GZIP="application/vnd.oci.image.layer.nondistributable.v1.tar+gzip",
     FOREIGN_BLOB_OCI_TAR_ZSTD="application/vnd.oci.image.layer.nondistributable.v1.tar+zstd",
     OCI_EMPTY_JSON="application/vnd.oci.empty.v1+json",
+    HELM="application/vnd.cncf.helm.config.v1+json",
```
ipanova, line 22: rather name it config_blob_helm

pulp_container/constants.py
```diff
@@ -19,6 +19,8 @@
...
+    COSIGN="application/vnd.dev.cosign.simplesigning.v1+json",
```
ipanova, line 23: cosign_blob to be more precise?

ipanova: what about attestations and sboms?
these are also cosgn produced artifacts? how about you split these into cosign_sig_blob, cosign_att_blob and cosign_sbom_blob?

ipanova: see more examples here https://quay.io/repository/sallyom/hello-go?tab=tags
for sboms one can have spdx type and cyclonedx but you should rather check what's the situation as of today, my info might alreayd be outdated.

pulp_container/constants.py
```diff
@@ -71,3 +73,12 @@
 SIGNATURE_PAYLOAD_MAX_SIZE = 4 * MEGABYTE
 
 SIGNATURE_API_EXTENSION_VERSION = 2
+
+MANIFEST_TYPE = SimpleNamespace(
+    IMAGE="image",
+    BOOTABLE="bootable",
+    FLATPAK="flatpak",
+    HELM="helm",
+    SIGNATURE="signature",
```
ipanova, line 82: i suggest to be more specific - cosign_signature

pulp_container/app/models.py
```diff
@@ -72,6 +73,7 @@ class Manifest(Content):
         digest (models.TextField): The manifest digest.
         schema_version (models.IntegerField): The manifest schema version.
         media_type (models.TextField): The manifest media type.
+        nature (models.TextField): The manifest's type (flatpak, bootable, signature, etc.).
```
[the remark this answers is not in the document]
git-hyagi: > How is this related to the OCI specification?
> Is it a pure invention on our side?

This field is not part of OCI spec. The idea came not only from us, but also as a request from Katello's team: https://github.com/pulp/pulp_container/issues/1751#issuecomment-2349236940

> Is the value easily derived from the "media_type"? Should it maybe be a choice field with MANIFEST_TYPE as the choices?

Unfortunately not for all types. To identify _flatpak_ we need to check the manifest **labels**, whereas _bootable_ images can be recognized through **labels** or **annotations**, _helm_ by checking the media_type from **configblob**, _cosigned images_ through the media_type from **layers**, and, so far, only "_common" images_ are identified through the **manifest media_type**. This is why we have the different methods to identify each type.

pulp_container/constants.py
git-hyagi: Right.

pulp_container/app/tasks/sync_stages.py
git-hyagi: I updated the current PR to address this concern, but I didn't modify the current implementation `ml.save(update_fields=["is_bootable", "is_flatpak"])` to avoid introducing bugs (and this block will be removed "soon" since these fields are now deprecated).

pulp_container/app/models.py
git-hyagi: ops... my bad, I thought schema1 was deprecated on our side as well.

pulp_container/constants.py
git-hyagi: Thank you for pointing the missing types! I also found this doc which was helpful to define the other cosign "derived types":  https://github.com/sigstore/cosign/blob/main/specs/SBOM_SPEC.md

pulp_container/app/models.py
```diff
@@ -154,40 +164,93 @@ def init_image_nature(self):
...
-            if manifest.is_bootable:
+            if manifest.type == MANIFEST_TYPE.BOOTABLE:
+                self.type = MANIFEST_TYPE.BOOTABLE
                 self.is_bootable = True
                 return True
-            elif manifest.is_flatpak:
+            elif manifest.type == MANIFEST_TYPE.FLATPAK:
+                self.type = MANIFEST_TYPE.FLATPAK
                 self.is_flatpak = True
                 return True
```
lubosmj, line 182: Can a manifest list be of type "helm"?

pulp_container/app/models.py
```diff
@@ -154,40 +164,93 @@ def init_image_nature(self):
...
+    def get_cosign_type(self, media_type):
+        if media_type in MEDIA_TYPE.COSIGN_SBOM:
+            return MANIFEST_TYPE.COSIGN_SBOM
+
+        cosign_types_mapping = {
+            MEDIA_TYPE.COSIGN_BLOB: MANIFEST_TYPE.COSIGN_SIGNATURE,
+            MEDIA_TYPE.COSIGN_ATTESTATION: MANIFEST_TYPE.COSIGN_ATTESTATION,
+            MEDIA_TYPE.COSIGN_ATTESTATION_BUNDLE: MANIFEST_TYPE.COSIGN_ATTESTATION_BUNDLE,
+        }
+        return cosign_types_mapping.get(media_type)
```
lubosmj, line 246: We should improve the readability aspect of this method. Is such a level of granularity even needed? If yes, let's move this mapping to `constants.py`.

pulp_container/app/registry_api.py
```diff
@@ -1213,6 +1214,11 @@ def put(self, request, path, pk=None):
                 ManifestInvalid(digest=manifest_digest)
 
             manifest_list = self._init_manifest(manifest_digest, media_type, raw_text_data)
+            manifest_list.type = MANIFEST_TYPE.UNKNOWN
+            for manifest in found_manifests:
+                if manifest.type in [MANIFEST_TYPE.BOOTABLE, MANIFEST_TYPE.FLATPAK]:
+                    manifest_list.type = manifest.type
+                    break
```
lubosmj, line 1221: Again, can the manifest list be of type "helm"?

pulp_container/app/tasks/sync_stages.py
```diff
@@ -287,9 +288,21 @@ async def resolve_flush(self):
...
 
         for manifest_list_dc in self.manifest_list_dcs:
+            manifest_list_type = MANIFEST_TYPE.UNKNOWN
             for listed_manifest in manifest_list_dc.extra_data["listed_manifests"]:
                 # Just await here. They will be associated in the post_save hook.
                 await listed_manifest["manifest_dc"].resolution()
+
+                # if manifestlist type is not defined as "unknown" we already checked its nature,
+                # in this case, there is no need to verify the other manifests from list
+                if manifest_list_type != MANIFEST_TYPE.UNKNOWN and listed_manifest[
```
lubosmj, line 298: This `if` condition is always evaluated to `False`, right? What is the rationale behind this procedure?

pulp_container/constants.py
```diff
@@ -71,3 +75,34 @@
...
 SIGNATURE_API_EXTENSION_VERSION = 2
+
+MANIFEST_TYPE = SimpleNamespace(
+    IMAGE="image",
+    BOOTABLE="bootable",
+    FLATPAK="flatpak",
+    HELM="helm",
+    COSIGN_SIGNATURE="cosign_signature",
+    COSIGN_ATTESTATION="cosign_attestation",
+    COSIGN_ATTESTATION_BUNDLE="cosign_attestation_bundle",
```
lubosmj, line 86: I am unsure we should differentiate between these. What would be a typical use case for seeing a difference between the bundle and regular attestation? Please, elucidate further.

pulp_container/constants.py
```diff
@@ -71,3 +75,34 @@
 SIGNATURE_PAYLOAD_MAX_SIZE = 4 * MEGABYTE
 
 SIGNATURE_API_EXTENSION_VERSION = 2
+
+MANIFEST_TYPE = SimpleNamespace(
+    IMAGE="image",
```
lubosmj, line 80: Maybe a residue, but I see you asserting this value in the tests. Did not we agree on "unkown" for images of unknown types? What is the purpose of having the "image" type if all images are images by their nature?

lubosmj (commented):
Overall, looking great! 🦆  I only left out a few clarifying questions.

New quest: Would you mind testing the migration command more deeply? For example, (1) sync content with pulp-contianer 2.19, (2) upgrade to 2.21 (without dropping the DB), (3) run the django-management command, (4) checkout back to this branch and upgrade, (5) run the migration command, (6) and manually verify the values of types and labels.

pulp_container/app/models.py
ipanova: Manifest list should stay as `list`. Nowadays I've seen lists containing artifacts and images, so it's a mixture of things..

pulp_container/constants.py
ipanova: Regular image that adheres to the specs is friendly to the container runtimes, meaning that it can be instantiated by `podman run`, flatpak image, for example cannot be instantiated(podman run will fail), hence you have a special cli to manage flatpak images and not podman.
So ideally if the image fully adheres to the specs, meaning that manifest is of the v2 type, config and layers are regular ones then it's a regular application image. For the rest, if there are deviations like in the config type or layer type that is not part or the constants we have then it's an 'unkown' type of image (for us).
On the thing that all images are images, you are right. The question whether we want to distinguish them between those 2 cases i've described.

pulp_container/app/tasks/sync_stages.py
git-hyagi: > This if condition is always evaluated to False, right?

Right, but for the first iteration execution only. For the following iterations, the `manifest_list_type` may have been changed to `flatpak` or `bootable`.


> What is the rationale behind this procedure?

This is to avoid having to define the `manifest_list_type` in every iteration of the loop. If we already identified it as `flatpak` or `bootable`, we can skip checking the `listed_manifest["manifest_dc"].content.type`.

pulp_container/constants.py
git-hyagi: I was thinking along the same lines as Ina. I thought the type `image` could be useful to differentiate "common" **container** images from other oci artifacts (like flatpak, or helm charts, etc).

pulp_container/app/models.py
lubosmj: Aah, good to know! I initially thought that if any of the listed manifests is of type "bla", then, we can treat the entire manifest list as being of type "bla". My goal was to reduce the number of API calls for our integrators, so they would not need to check each listed manifest individually to determine the type. By treating the manifest list as the "list" type, they would need to iterate through all listed manifests to get the right type and potentially present it in the UI (because users should see only the tagged manifests/manifest lists in the UI).

It might even make sense to allow filtering manifests by their referencing manifest list. Meaning that only one additional API call will be needed at most.

lubosmj: Also, when it comes to different types of listed manifests, we might even consider prioritizing one over the other. The type that has the highest priority will be displayed in the view.

pulp_container/constants.py
lubosmj: If the image fully adheres to the specs, it's media type is already shown in the Manifests endpoint as either "docker.manifest.v2" or "oci.manifest", correct? Clients do not read our API endpoints. So, this type thingy is irrelevant to them. However, the integrators can still benefit from the information we provide. For example, they can determine that if media_type is one of the above and type is set to "uknown", it can be treated as a regular image. How does that sound to you?

pulp_container/app/models.py
git-hyagi: So if we decide to set manifestlist as `type: list`, Pulp integrators would need to check and identify the corresponding manifests types by themselves. But if we set the type based on a single manifest from the list (for example, `flatpak`) there is a risk of providing the wrong type since we can now have a mixture of things. :thinking:

pulp_container/app/tasks/sync_stages.py
git-hyagi: Actually, you are right! My bad! :man_facepalming: 
The condition should be `if manifest_list_type == MANIFEST_TYPE.UNKNOWN and ...`

pulp_container/app/models.py
lubosmj: If we decide to take a middle ground, it would be best to "allow filtering manifests by their referencing manifest list". Another PR. Also, note that this "type" field is pretty flexible. We can add or remove new types later.

lubosmj: Unless, we do not enforce these types on the DB level...

ipanova: Initially manifest list was introduced for the multi-arch builds of the same image, nowadays it can also be treated as a list of images collected together. there is no enforcement in the specs around this.
Here's a practical example `skopeo inspect --raw docker://quay.io/nalind/testing:podman-machine-unified|jq`

I would not do any heuristics and just leave type as `list`. Integrators can take their own risks and parse and prioritize the types the consider more important on their own.

ipanova: @ianballou FYI the rationale behind ^

pulp_container/constants.py
ipanova: Not sure what would be best approach. Concept of a manifest is a more low level thingy, and the fact that it splits into an image/artifact/list. End users usually are familiar with `container image concept`. So if we keep `list`, types we parse like (flatpak, bootable) then maybe it makes sense to leave `image` for the rest instead of `unknown`

ipanova: speaking of.. i did not really see this PR handling the `artifact`type, was it intentional?

lubosmj: Okay, let's leave the `image` and `list` in place. We can still remove it later as needed. @git-hyagi, please, consider adding the `artifact` type to the list.

pulp_container/app/models.py
ipanova: @lubosmj  you should have inspected each child manifest and you would have seen that 2 of them are artifacts and 2 are different arches for a regular image.

lubosmj (approved):
Let's give it a shot! 🔫

PR #1797, opened 2024-10-24
Fix Any serializer type regression breaking Ruby bindings

pedro-psb:
See context: https://github.com/pulp/pulp_rpm/issues/3639


lubosmj (approved):
Thank you for submitting this! 🦜

lubosmj:
These fields were introduced in 2.19. I am backporting the change only to 2.20 and 2.2.

PR #1802, opened 2024-10-30
Add artifact as a new manifest type

git-hyagi:


pulp_container/app/models.py
```diff
@@ -244,6 +247,20 @@ def is_helm_chart(self):
...
 
+    def is_artifact(self):
+        # artifact is valid only for OCI spec
+        if self.media_type != MEDIA_TYPE.MANIFEST_OCI:
+            return False
+
+        if self.json_manifest.get("artifactType", None):
+            return True
+
+        manifest_config_media_type = self.json_manifest["config"]["mediaType"]
```
mdellweg, line 258: We parse the same json twice here. And probably more times in the chain of all the other `is_anything` functions. What is the easiest way to get some memoization without messing up the cache invalidation?

lubosmj: One of the ways would be to create an internal instance variable `_json_manifest` that will be initialized once.
```
def __init__(self):
    self._json_data = None

@property
def json_manifest(self):
    if not self._json_manifest:
         self._json_manifest = json.loads(bla)
    return self._json_manifest
```

@git-hyagi, try the above ^.

https://docs.djangoproject.com/en/5.1/ref/models/instances/#django.db.models.Model

lubosmj (changes requested):
Caching the json value is indeed useful.

lubosmj (approved):

PR #1804, opened 2024-11-01
Use Containerfile as default if none provided

git-hyagi:
When a build_context is provided without a containerfile_name, this commit will verify if a file called Containerfile exists in the provided file repository (build_context) and use it in the build task.

[noissue]


ipanova:
Quick suggestion is to also look for Dockerfile, these two should be interchangeable. Podman will prefer Containerfile over Dockerfile if both are present in the context dir.

Issue #1812, opened 2024-11-05
Cannot communicate with Sonatype Nexus docker repository, fails with 401 with correct credentials


Soraph:
**Version**
The problem has been found with the following scenario:

Installed via Automation Platform 2.4
```
python39-pulp-container.noarch     2.15.5-1.el8ap
python39-pulpcore.noarch           3.28.32-1.el8ap
```

Installed via Automation Platform 2.5
```
python3.11-pulp-container.noarch   2.19.3-1.el9ap
python3.11-pulpcore.noarch         3.49.19-1.el9ap
```

**Describe the bug**
When attempting to communicate with a registry of type Sonatype Nexus the authentication step does not happen correctly.

Same credentials used with `podman` allow proper login and operations.

**To Reproduce**

Testing with Automation Hub 2.4
1. Create a new registry with the Sonatype Nexus URL and credentials
2. Define a new Execution Environment definition (image)
3. Attempt to synchronize it
4. 401 error
```
pulpcore-worker[70580]: pulp [fa3fba1ec0eb4238b421812448c70e22]: pulpcore.tasking.tasks:INFO: Starting task 0192fb6b-5231-70e9-a1ad-9be63c8f897c
pulpcore-worker[70580]: pulp [fa3fba1ec0eb4238b421812448c70e22]: pulp_container.app.tasks.synchronize:INFO: Synchronizing: repository=test remote=test
pulpcore-worker[70580]: Giving up download_wrapper(...) after 1 tries (aiohttp.client_exceptions.ClientResponseError: 401, message='Unauthorized', url=URL('http://nexus.domain/v2/'))
pulpcore-worker[70580]: pulp [fa3fba1ec0eb4238b421812448c70e22]: backoff:ERROR: Giving up download_wrapper(...) after 1 tries (aiohttp.client_exceptions.ClientResponseError: 401, message='Unauthorized', url=URL('http://nexus.domain/v2/'))
pulpcore-worker[70580]: pulp [fa3fba1ec0eb4238b421812448c70e22]: pulpcore.tasking.tasks:INFO: Task 0192fb6b-5231-70e9-a1ad-9be63c8f897c failed (Connection closed)
pulpcore-worker[70580]: pulp [fa3fba1ec0eb4238b421812448c70e22]: pulpcore.tasking.tasks:INFO:   File "/usr/lib/python3.9/site-packages/pulpcore/tasking/tasks.py", line 66, in _execute_task
[...]

PR #1830, opened 2024-11-22
Fix an issue with auth header validation

git-hyagi:
fixes: #1812


mdellweg:
@git-hyagi Please don't skip supported branches in backports. Otherwise they retroactively regress.

PR #1868, opened 2024-12-17
Remove sudo from post_before_script in CI

gerrod3:


.github/workflows/scripts/post_before_script.sh
```diff
@@ -6,7 +6,7 @@ fi
 
 # add the copied certificates from install.sh to the container's trusted certificates list
 if [[ "$TEST" = "azure" ]]; then
-  cmd_prefix sudo trust anchor /etc/pki/tls/cert.pem
+  cmd_prefix trust anchor /etc/pki/tls/cert.pem
 else
-  cmd_prefix sudo trust anchor /etc/pulp/certs/pulp_webserver.crt
+  cmd_prefix trust anchor /etc/pulp/certs/pulp_webserver.crt
```
mdellweg, line 11: This _will_ be challenged hard by the plugin template...

mdellweg (approved):

.github/workflows/scripts/post_before_script.sh
gerrod3: No it won't, this script isn't templated by the plugin template. All the pre & post scripts are custom per plugin.

PR #1890, opened 2025-01-14
[PULP-255] Remove backward compatibility for manifest with artifact

### LINKED ISSUE #1621: Stop supporting manifests with artifacts and enforce the data migration

dralley:
This commit removes support for manifests storing their data inside an artifact instead of using the recently introduced  Manifest.data text field.

closes #1621

### DIFF
--- a/pulp_container/app/tasks/sync_stages.py
+++ b/pulp_container/app/tasks/sync_stages.py
@@ -36,7 +36,6 @@ from pulp_container.app.utils import (
     validate_manifest,
     calculate_digest,
     filter_resources,
-    get_content_data,
 )
 
 log = logging.getLogger(__name__)
@@ -79,25 +78,11 @@ class ContainerFirstStage(Stage):
 
         digest = response.headers.get("docker-content-digest")
 
-        if (
-            manifest := await Manifest.objects.prefetch_related("contentartifact_set")
-            .filter(digest=digest, pulp_domain=get_domain())
-            .afirst()
-        ):
-            if raw_text_data := manifest.data:
-                content_data = json.loads(raw_text_data)
-
-            # TODO: BACKWARD COMPATIBILITY - remove after fully migrating to artifactless manifest
-            elif saved_artifact := await manifest._artifacts.afirst():
-                content_data, raw_bytes_data = await sync_to_async(get_content_data)(saved_artifact)
-                raw_text_data = raw_bytes_data.decode("utf-8")
-            # if artifact is not available (due to reclaim space) we will download it again
-            else:
-                content_data, raw_text_data, response = await self._download_manifest_data(
-                    response.url
-                )
-            # END OF BACKWARD COMPATIBILITY
-
+        if manifest := await Manifest.objects.filter(
+            digest=digest, pulp_domain=get_domain()
+        ).afirst():
+            raw_text_data = manifest.data
+            content_data = json.loads(raw_text_data)
         else:
             content_data, raw_text_data, response = await self._download_manifest_data(response.url)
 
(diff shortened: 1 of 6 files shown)


pulp_container/app/registry_api.py
```diff
@@ -1120,7 +1120,7 @@ def handle_safe_method(self, request, path, pk):
                     self.fetch_manifest(remote, pk)
                     return redirects.redirect_to_content_app("manifests", pk)
                 elif manifest:
-                    return redirects.issue_manifest_redirect(manifest)
+                    return Response(manifest.data)
```
dralley, line 1123: TODO: https://github.com/pulp/pulp_container/pull/1644#discussion_r1630995990

dralley: @git-hyagi @gerrod3 Any idea which headers specifically he was referring to?

git-hyagi: I was checking the [OCI spec](https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md#pulling-manifests) and found this:

> _"In a successful response, the Content-Type header will indicate the type of the returned manifest. The Content-Type header SHOULD match what the client pushed as the manifest's Content-Type. If the manifest has a mediaType field, clients SHOULD reject unless the mediaType field's value matches the type specified by the Content-Type header."_
> _"A successful response SHOULD contain the digest of the uploaded blob in the header Docker-Content-Digest."_


and searching the sample (`return web.Response(...)`) provided by Lubos, I found these headers definitions (which comply with the oci spec):
https://github.com/pulp/pulp_container/blob/18dab985c11d289f3a222f1fce04e7eb3e47cb1d/pulp_container/app/registry.py#L137-L142

dralley: I'm going to treat removing this redirect as a separate issue.

dralley: https://github.com/pulp/pulp_container/issues/1974

I've assigned myself to that work.  But this PR should probably go ahead and proceed as-is, w/ passing CI.

pulp_container/app/tasks/sync_stages.py
```diff
@@ -79,25 +77,9 @@ async def _check_for_existing_manifest(self, download_tag):
...
-                content_data, raw_bytes_data = await sync_to_async(get_content_data)(saved_artifact)
-                raw_text_data = raw_bytes_data.decode("utf-8")
-            # if artifact is not available (due to reclaim space) we will download it again
-            else:
-                content_data, raw_text_data, response = await self._download_manifest_data(
-                    response.url
-                )
-            # END OF BACKWARD COMPATIBILITY
-
+        if manifest := await Manifest.objects.filter(digest=digest).afirst():
```
gerrod3, line 80:
```suggestion
        if manifest := await Manifest.objects.filter(digest=digest, pulp_domain=get_domain()).afirst():
```

pulp_container/app/tasks/sync_stages.py
```diff
@@ -460,26 +440,10 @@ async def create_listed_manifest(self, manifest_data):
...
-            else:
-                content_data, manifest = await self._download_and_instantiate_manifest(
-                    manifest_url, digest
-                )
-            # END OF BACKWARD COMPATIBILITY
-        else:
-            content_data, manifest = await self._download_and_instantiate_manifest(
-                manifest_url, digest
-            )
+        if manifest := await Manifest.objects.filter(digest=digest).afirst():
```
gerrod3, line 443:
```suggestion
        if manifest := await Manifest.objects.filter(digest=digest, pulp_domain=get_domain()).afirst():
```

gerrod3 (changes requested):
Two small fixes then lgtm!

gerrod3 (approved):

Issue #1887, opened 2025-01-08
500 when accessing flatpak index after space reclamation


ianballou:
**Version**
```
      "versions": {
        "deb": "3.3.1",
        "rpm": "3.27.2",
        "core": "3.63.6",
        "file": "3.63.6",
        "ostree": "2.4.4",
        "python": "3.12.5",
        "ansible": "0.22.3",
        "certguard": "3.63.6",
        "container": "2.22.0"
      }
```

**Describe the bug**
```
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:  File "/usr/lib/python3.11/site-packages/pulp_container/app/registry_api.py", line 680, in get
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:    response = super().get(request)
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:               ^^^^^^^^^^^^^^^^^^^^
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:  File "/usr/lib/python3.11/site-packages/pulp_container/app/registry_api.py", line 639, in get
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:    with storage.open(manifest.config_blob._artifacts.get().file.name) as file:
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:  File "/usr/lib/python3.11/site-packages/django/db/models/manager.py", line 87, in manager_method
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:    return getattr(self.get_queryset(), name)(*args, **kwargs)
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:  File "/usr/lib/python3.11/site-packages/django/db/models/query.py", line 637, in get
Jan  8 21:30:26 centos9-katello-devel-stable pulpcore-api[3002]:    raise self.model.DoesNotExist(
[...]

PR #1899, opened 2025-01-21
Fix 500 error when using flatpak with reclaimed content [PULP-262]

gerrod3:
After looking through the history and current state of the Manifest model I came up with this solution to try to get the metadata off the model if it is there and then fallback to the config blob. If the config blob is not there during a fallback then the image will be skipped from the returned results of the index call. This bugfix can only really be back ported to 0.22 since it relies on the new manifest fields. Flatpak index was added in 0.16, so a different solution would be needed to be backported for these inbetween releases. Thoughts on whether I should create these other backports?

fixes: #1887

### DIFF
--- a/CHANGES/1887.bugfix
+++ b/CHANGES/1887.bugfix
@@ -1,0 +1,1 @@
+Fixed flatpak index returning 500 when Manifest content was on_demand or had been reclaimed.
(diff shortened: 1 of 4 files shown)


ianballou (commented):
I tested this on a Katello machine running pulp container 2.22.0, and it seems to be working well. I switch some repos from immediate to on demand and ran reclaim space. After that, the index still remained.

git-hyagi (approved):

PR #1919, opened 2025-02-06
Remove pulp-smash from test_recursive_add

gerrod3:


decko (approved):

pulp_container/tests/functional/api/test_recursive_add.py
```diff
@@ -1,670 +1,702 @@
...
-        api_client = gen_container_client()
-        cls.repositories_api = RepositoriesContainerApi(api_client)
-        cls.remotes_api = RemotesContainerApi(api_client)
-        cls.versions_api = RepositoriesContainerVersionsApi(api_client)
-        cls.tags_api = ContentTagsApi(api_client)
-        cls.manifests_api = ContentManifestsApi(api_client)
+        remote = container_remote_factory(upstream_name=PULP_FIXTURE_1)
+        repo = container_repository_factory()
+        container_sync(repo, remote)
+        return container_bindings.RepositoriesContainerApi.read(repo.pulp_href)
```
mdellweg, line 327: If that is exactly the same as above (it looks to me like that), you can define it outside of the classes (still being class scoped.)

mdellweg (approved):
One suggestion for improvement, but easy to do in another PR.

PR #1921, opened 2025-02-06
Remove pulp-smash from test_pull_content

gerrod3:


pulp_container/app/serializers.py
```diff
@@ -159,6 +159,13 @@ class BlobSerializer(SingleArtifactContentSerializer):
 
     digest = serializers.CharField(help_text="sha256 of the Blob file")
 
+    def __init__(self, *args, **kwargs):
+        """Fix for bindings to allow for on-demand blobs."""
+        # TODO: Move into pulpcore
+        super().__init__(*args, **kwargs)
+        if "artifact" in self.fields:
+            self.fields["artifact"].allow_null = True
+
```
mdellweg, line 168: So if I read correctly, this allows to serialize on-demand blobs.
Is this serializer ever used to create blobs via the api? It may be OK for this piece of content, but i doubt, we want it generally for serializers that are also used to create content.

mdellweg: Can you just add these last two sentences as comments?

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -309,94 +206,44 @@ def setUpClass(cls):
...
-    @classmethod
-    def tearDownClass(cls):
-        """Clean class-wide variable."""
-        for cleanup_function, args in reversed(cls.teardown_cleanups):
-            cleanup_function(args)
-
-    def test_api_returns_same_checksum(self):
+        monitor_task(
+            pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 0}).task
+        )
```
mdellweg, line 211: We do have the orphan_cleanup_pre fixture for this I believe.
And it even prevents you from adding the parallel mark.

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -309,94 +206,44 @@ def setUpClass(cls):
...
+        remote = container_remote_factory(policy="on_demand")
+        container_sync(repo, remote)
+        type(self).repo = container_bindings.RepositoriesContainerApi.read(repo.pulp_href)
+        type(self).distribution_with_repo = container_distribution_factory(
+            repository=repo.pulp_href
+        )
+        type(self).distribution_with_repo_version = container_distribution_factory(
+            repository_version=self.repo.latest_version_href
+        )
+        type(self).artifact_count = pulpcore_bindings.ArtifactsApi.list().count
```
mdellweg, line 223: Do you mind emplyoing the `from_repo` like pattern here too?

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -58,126 +42,56 @@ def setUpClass(cls):
 
         * `Pulp #4460 <https://pulp.plan.io/issues/4460>`_
```
dralley, line 43: Whereever we have plan.io links, we can probably ditch those.

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -58,126 +42,56 @@ def setUpClass(cls):
...
         """
         # Get local checksums for content synced from the remote registry
-        checksums = [
-            content["digest"]
-            for content in get_content(self.repo.to_dict())[CONTAINER_CONTENT_NAME]
+        blob_checksums = [
+            c.digest
+            for c in container_bindings.ContentBlobsApi.list(
+                repository_version=self.repo.latest_version_href
+            ).results
```
dralley, line 68: Style nitpick, this would be a bit cleaner if the query was broken out and assigned to a variable.

You don't actually have to fix this. It's merely a suggestion.

dralley: Also if the primary operation being done here is a lookup, a set comprehension would be better.  But again I don't care that much.

pulp_container/tests/functional/api/test_pull_content.py
dralley: It's called `delete_orphans_pre`, yeah?

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -309,94 +206,44 @@ def setUpClass(cls):
...
-
-    def test_api_returns_same_checksum(self):
+        monitor_task(
+            pulpcore_bindings.OrphansCleanupApi.cleanup({"orphan_protection_time": 0}).task
+        )
+        registry_client.rmi("-a", "-f")
+        repo = container_repository_factory()
+        remote = container_remote_factory(policy="on_demand")
+        container_sync(repo, remote)
+        type(self).repo = container_bindings.RepositoriesContainerApi.read(repo.pulp_href)
```
dralley, line 216: Can you explain why `type(self)` is needed?

pulp_container/tests/functional/api/test_pull_content.py
```diff
@@ -186,117 +100,100 @@ def test_create_empty_blob_on_the_fly(self):
...
+        non_existing_blob_url = urljoin(bindings_cfg.host, blob_path)
 
         auth = get_auth_for_url(non_existing_blob_url)
         content_response = requests.get(non_existing_blob_url, auth=auth)
-        assert content_response.status_code, 404
+        assert content_response.status_code == 404
 
 
-class PullOnDemandContentTestCase(unittest.TestCase):
+class TestPullOnDemandContent:
```
dralley, line 182: Is it not possible for `TestPullContent` and `TestPullOnDemandContent` to share some of this code instead of duplicating so much of it?  It looks like much of it was copied and pasted at one point.

dralley: This shouldn't be interpreted as a blocker to the immediate goal of dropping pulp-smash ^

pulp_container/tests/functional/api/test_pull_content.py
gerrod3: I already tried, it is not allowed because of scope mistmatch.

dralley (approved):

dralley (commented):
Ack, didn't notice the CI turned red again

gerrod3:
@dralley Can you reapprove, got the CI green now.

dralley (approved):

PR #1923, opened 2025-02-07
Remove pulp-smash from test_content_cache & test_crud_distributions

gerrod3:


pulp_container/tests/functional/api/test_crud_distributions.py
```diff
@@ -1,297 +1,49 @@
...
-            distribution = self.distribution_api.read(task.created_resources[0])
-            self.addCleanup(self.namespace_api.delete, distribution.namespace)
-
-    def test_invalid_base_paths(self):
-        """Test the validation for base paths."""
-        self._try_create_distribution(base_path="i/am/an/very:bad:repo/name.py")
-        self._try_create_distribution(base_path="invalid_")
-        self._try_create_distribution(base_path="_invalid")
-        self._try_create_distribution(base_path="another/invalid..reponame")
-        self._try_create_distribution(base_path="also/invalid___reponame")
```
mdellweg, line 279: I think these tests (or maybe the whole class) got lost. Were they redundant?

gerrod3: Yeah, we already test these in pulpcore afaik.

PR #1924, opened 2025-02-07
Remove pulp-smash from test_recursive_remove

gerrod3:


pulp_container/tests/functional/api/test_recursive_remove.py
```diff
@@ -1,465 +1,467 @@
...
-        cls.repositories_api.delete(cls.from_repo.pulp_href)
-        cls.remotes_api.delete(cls.remote.pulp_href)
-        delete_orphans()
-
-    def test_repository_only_no_latest_version(self):
+        repo = container_repository_factory()
+        remote = container_remote_factory(upstream_name=PULP_FIXTURE_1)
+        container_sync(repo, remote)
+        type(self).from_repo = container_bindings.RepositoriesContainerApi.read(repo.pulp_href)
+        type(self).latest_from_version = self.from_repo.latest_version_href
```
mdellweg, line 29: I see, why we cannot make these class attributes easily. But I fear this method of injecting the hrefs into the tests is a bit hard to understand for our future selves.
Have you thought about creating a setup fixture that provides a typed namespace?

gerrod3: Technically the trick with type is that we are making them class attributes :). I could make the setup fixture just return the needed values or a namespace object.

mdellweg: I think (really just personal preference) that'd be easier to read.

pulp_container/tests/functional/api/test_recursive_remove.py
```diff
@@ -1,465 +1,488 @@
...
-        cls.remotes_api.delete(cls.remote.pulp_href)
-        delete_orphans()
-
-    def test_repository_only_no_latest_version(self):
+        repo = container_repository_factory()
+        remote = container_remote_factory(upstream_name=PULP_FIXTURE_1)
+        container_sync(repo, remote)
+        from_repo = container_bindings.RepositoriesContainerApi.read(repo.pulp_href)
+        latest_from_version = from_repo.latest_version_href
+        return from_repo, latest_from_version
```
mdellweg, line 30: Yet another copy of the `from_repo` fixture? maybe it should live in conftest...

PR #1952, opened 2025-03-03
Update CI files for branch 2.24

pulpbot:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers=[
 requires-python = ">=3.9"
 dependencies = [
   "jsonschema>=4.4,<4.24",
-  "pulpcore>=3.71,<3.85",
+  "pulpcore>=3.72.0,<3.85",
```
dralley, line 28: @gerrod3 This is an existing release branch, I presume it should not be bumping the lower bound here?

dralley: The CI update PR for `main` also did this, but it's not against the rule to make this change there so I approved it - it may not have been necessary though if it was accidental.  I clicked this one second.

gerrod3: The problem is that we have a policy that the lowerbound of any release branch should be a supported pulpcore release. Since there are no supported pulpcore branches greater than 3.63 this value is going to keep getting bumped to the latest pulpcore version till there is one. The original value was 3.68.2. We should discuss this in the pulpcore meeting, don't approve this one.

dralley: >This is an existing release branch

To be clear I meant "this branch (2.24) is an existing release branch for pulp_container", not pulpcore.  Bumping y-release pulpcore versions in z-releases of plugins is atypical was my point.

You probably figured that out, I'm just clarifying because I realized it wasn't obvious from the way I phrased it.

I agree we should bring it up at the pulpcore meeting, +1 to that.

gerrod3 (approved):

Issue #1926, opened 2025-02-11
Tag overwrite issue when syncing multiple images with the same tag in Pulp


VrindaMarwah:
When syncing multiple container images that have the same tag in Pulp, the tag is only associated with the latest manifest synced. This overwrites any previous manifests associated with the tag, causing inconsistencies when pulling images. As a result, pulling both images leads to identical checksums, even though the images should be different.

**Steps to Reproduce:**
1) Sync two or more container images with the same tag into Pulp. For example:
**Image 1:** registry.k8s.io/kube-apiserver:v1.31.4
**Image 2:** registry.k8s.io/kube-proxy:v1.31.4

2) Inspect the tag associations in Pulp after the sync using command: ``` pulp container repository content list --repository foo ```
3) Pull both images using the tag.

**Expected Behavior:**
Each image should retain its unique manifest, and the tag should point to the correct manifest corresponding to the image being pulled, without overwriting.

**Actual Behavior:**
The tag is associated with only one manifest (the last one synced), overwriting others. Pulling both images results in identical checksums despite the images being different.

Pulp should introduce support for versioned tagging or maintain a history of manifests associated with the same tag so that the user can prevent conflicts/checksum overwrite issues while pulling multiple images having the same tag. Optionally, Pulp can provide an API or CLI feature to list all manifests associated with a specific tag, enabling better transparency and control.

gerrod3:
I think you might be using pulp_container incorrectly, it behaves a bit differently then other Pulp plugins. In container land a repository = an image with all its versions and tags. While we allow users to sync multiple images into one repository, you should really only be syncing one per repository. So in your case kube-apiserver and kube-proxy should both get their own repository + distribution, that way their tags are correctly separated.

VrindaMarwah:
Appreciate the quick response and clarification. I was able to sync each image into its own repository and distribution to keep the tags properly separated. Thanks for pointing this out!

VrindaMarwah:
@gerrod3 Pulp container remote create command doesn't allow to sync images with "latest" tag. 
E.g:
Created remote using below command:
```pulp container remote update --name remote_docker.io_kubeflowkatib_pytorch-mnist-cpu_latest --url https://registry-1.docker.io --upstream-name kubeflowkatib/pytorch-mnist-cpu --policy immediate --include-tags '["latest"]' ```

Synced image using below command:
[...]

PR #1957, opened 2025-03-07
Fix overflow error on images >2GB

gerrod3:
fixes: #1926


gerrod3:
@mdellweg The affected field was added in 2.22, but only 2.24 is supported (since it is the latest release, but it will probably become a full supported branch with the latest katello), so my plan is to backport this fix to 2.24 and then yank the earlier 2.24 releases. Wdyt?

mdellweg:
There's a migration. Backporting migrations is a bad idea. And yanking older releases does not help people that already installed them. Actually i don't see any reason for yanking here. There'd be a newer version with a fix, that should be sufficient.

Is 0.24 the currently latest branch? In that case we may be able to make an exception. But it's still dangerous and strategically speaking it will probably lead to asking for more exceptions.

Reminder: The issue with backporting migrations is nothing about the complexity of the migrations itself, but only about inconsistent migration history in the installed system.

grzleadams:
Is there an ETA on this? Our Pulp instance is basically useless for containers right now until this is fixed.

VrindaMarwah:
Meanwhile, could you provide us with the details on which Pulp Container image tag has the fix for this issue so that we will be able to pull the specific image and utilize the fix for images > 2GB. 

Additionally, with the latest pulp image from dockerhub, post deployment of pulp container, is it possible to upgrade the pulp container version to 2.22? If yes, could you provide with the appropriate steps for the same.

mdellweg (approved):
The fix sounds easy enough.

Issue #1534, opened 2024-03-05
Add support for domains


lubosmj:
Domains is an optional feature that allows multi-tenancy.

In the Pulp Container plugin, we need to deal with the fact that we have a single registry (a single source of truth, a remote server) that can be accessed by anyone. An open question is whether we should have a different namespace per each domain when serving the content or not. This would induce the requirement to reconsider redirects. Usually, `CONTENT_PATH_PREFIX` is a subject to change when dealing with domains' scopes.

https://docs.pulpproject.org/pulpcore/workflows/domains-multi-tenancy.html
https://docs.pulpproject.org/pulpcore/plugin_dev/plugin-writer/concepts/domains/domains_compatibility.html

Plugins adding domains support:
https://github.com/pulp/pulp_rpm/pull/3106
https://github.com/pulp/pulp_ostree/pull/346

lubosmj:
Right now, we are issuing redirects, like this:
- v2/repo/manifests/bla -> pulp/content/manifests/id/

What would we like to implement?
1. v2/repo/manifests/bla -> pulp/default/content/manifests/id/ <----- bad, we do not know the domain ID from api-app
2. v2/default/repo/manifests/bla -> pulp/default/content/manifests/id/ <----- bad, we are breaking the standardized Registry API

Needs more thinking...

lubosmj:
The expansion to the domains' path will need to be identified on the pull level:

podman pull pulp.example.com/ipanova/busybox -> podman pull pulp.example.com/default/ipanova/busybox

podman pull pulp.example.com/redhat/ipanova/busybox -> podman pull pulp.example.com/redhat/ipanova/busybox

The content-app path will be then explicitly retrieved from the API, like pulp/redhat/manifests/blabla.

PR #1965, opened 2025-03-26
Add full domain support

gerrod3:
fixes: #1534
https://issues.redhat.com/browse/PULP-344

### DIFF
--- a/docs/admin/learn/domain.md
+++ b/docs/admin/learn/domain.md
@@ -1,0 +1,54 @@
+# Domain support
+
+Enabling domain support in pulp_container works a bit differently than other plugins due to the 
+nature of the Registry API. Each domain is scoped as a unique registry by the domain name. This 
+means that all images will include the domain name as a prefix to its repository path.
+
+## Examples
+
+### Sync and Pull
+
+Here's an example of syncing and hosting an image (`pulp/pulp`) in the domain `foo`.
+
+```bash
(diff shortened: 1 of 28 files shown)


docs/admin/learn/domain.md
```diff
@@ -0,0 +1,54 @@
...
+docker pull localhost:24817/foo/bar:latest
+```
+
+!!! note
+
+    Objects are still prohibited from being used across domains even if you have permissions for both.
+
+!!! note
+
+    The Flatpak endpoints will only work within the default domain, even with domains enabled.
```
dralley, line 53: Is that a temporary or permanent limitation?

gerrod3: Temporary, I hope. I need to do more research into flatpak and how it works (the docs aren't very helpful about using flatpak with an oci-registry) to see if we can somehow add the domain-name to the urls that flatpak uses (currently `/index/static/` and `/index/dynamic/`

dralley:
@gerrod3 Can you give me some idea of what changed since the last time I looked it over (when that comment was made)?

dralley:
Or should I just do a full look through again (if a lot has changed)

gerrod3:
@dralley The only thing I changed was the tests. I fixed them for the CI (properly take into the account the different settings the CI is tested under).

dralley (approved):

PR #1966, opened 2025-03-27
Added set/unset_label support to tag/blob/manifest/manifestsignature.

### LINKED ISSUE #1954: [PULP-350] Teach container to allow access to the new set/unset_label options for Content
pulpcore introduce labels-on-content with PR https://github.com/pulp/pulpcore/pull/5787

It requires an addition to the RBAC rules to allow access to the new set/unset_label API calls, as noted here : https://github.com/pulp/pulpcore/pull/5787/files#diff-dc946cef2bc4fef289c4a2d72dde9e612e92efc300edb9273a4e01b914092401R85

ggainey:
closes #1954.

### DIFF
--- a/pulp_container/tests/functional/api/test_rbac_repo_content.py
+++ b/pulp_container/tests/functional/api/test_rbac_repo_content.py
@@ -4,6 +4,8 @@ import pytest
 
 
 from pulp_container.tests.functional.constants import PULP_FIXTURE_1, REGISTRY_V2_REPO_PULP
+from pulpcore.client.pulp_container.exceptions import ForbiddenException
+from pulpcore.client.pulp_container import SetLabel, UnsetLabel
 
 
 def test_rbac_repository_content(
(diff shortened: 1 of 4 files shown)


pulp_container/tests/functional/api/test_rbac_repo_content.py
```diff
@@ -182,3 +183,85 @@ def test_rbac_repository_content(
...
+        assert labels["key_1"] == "XXX"
+
+        # Unset a label
+        sl = UnsetLabel(key="key_1")
+        the_binding.unset_label(the_content.pulp_href, sl)
+        content2 = the_binding.read(the_content.pulp_href)
+        assert "key_1" not in content2.pulp_labels
+
+    if pulp_settings.TOKEN_AUTH_DISABLED:
+        pytest.skip("RBAC cannot be tested when token authentication is disabled")
```
gerrod3, line 224:
```suggestion
```
This RBAC check is only for the Registry API. Since this is the Pulp API we should be good.

gerrod3 (changes requested):
One little change, the rest looks good.

pulp_container/tests/functional/api/test_rbac_repo_content.py
gerrod3: @ggainey You removed the wrong lines. That is why the s3 test is failing now.

ggainey: ....ha!  Code in haste, repent at leisure :)

gerrod3 (approved):

PR #1978, opened 2025-04-02
Fix overflow error when syncing/pushing images >2gb [2.22]

gerrod3:
https://issues.redhat.com/browse/PULP-449

### DIFF
--- a/pulp_container/app/models.py
+++ b/pulp_container/app/models.py
@@ -31,7 +31,7 @@ from pulpcore.plugin.util import gpg_verify
 
 
 from . import downloaders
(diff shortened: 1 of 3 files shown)


ianballou (commented):
I've tested this on a Katello machine running Pulp Container 2.22, and it works well! I was able to sync a large AI model. The large images have a compressed_image_size of `null` now looking at the Pulp API.

git-hyagi (approved):

dralley (approved):

ggainey (approved):
Nice - clever way to fix the problem when the "real" fix can't be backported!

PR #2029, opened 2025-06-16
Debug nightly hanging

pedro-psb:


pedro-psb:
`pycares==4.9.0` was the faulty dep.
Closing because I'll add that to plugin_template.

PR #2043, opened 2025-08-12
Bump pulpcore upperbound to <3.100

gerrod3:


gerrod3:
Interestingly I didn't have to squash the migrations for container since the core migration removing the `BaseDistribution` gets run after all the container migrations run. I think it's an alphabet thing, where container gets scheduled first, but it depends on some core migrations so those partially run till container migrations can fully run, then the core migrations resume.

ggainey (approved):

PR #2045, opened 2025-08-13
Check for file in ENABLED_PLUGINS

mdellweg:


pulp_container/app/checks.py
```diff
@@ -6,6 +6,17 @@
 def container_settings_check(app_configs, **kwargs):
     errors = []
 
+    # pulp_container depends on pulp_file so if ENABLED_PLUGINS is used, it must be contained.
+    if ENABLED_PLUGINS := getattr(settings, "ENABLED_PLUGINS", None) is not None:
```
evgeni, line 10:
```suggestion
    if (ENABLED_PLUGINS := getattr(settings, "ENABLED_PLUGINS", None)) is not None:
```

otherwise `ENABLED_PLUGINS` is a boolean ;)

PR #2059, opened 2025-09-03
Fixed a bug on distribution create

mdellweg:
Due to a recent pulpcore change, the distribution serializer must be able to serialize in the task context where the request is set to none.


pedro-psb (approved):

mdellweg:
general_create needs to be able to serialize the task result in the task.

PR #2064, opened 2025-09-13
Fix replicate not creating distributions

gerrod3:
Forgot to change the distribution extra fields method name as required by this pulpcore change for 3.70: https://github.com/pulp/pulpcore/commit/32a24b18936b82c752c30780ef63ea0827e4f017. Since we didn't add tests for replicate (the feature was added before we had domain support) we didn't catch this breaking change, so I added some tests. Current pulpcore requirement is >=3.73,<3.100 so this should only need to be backported to the latest release.


pulp_container/tests/functional/api/test_domains.py
```diff
@@ -319,3 +323,63 @@ def test_cross_domain_pulp_apis(
             "future_base_path": "test",
         },
     )
+
+
+@pytest.mark.parallel
+def test_domain_content_replication(
```
ggainey, line 329: :+1:

ggainey (approved):

PR #2140, opened 2026-01-06
Update to allow core/3.100.

ggainey:


ggainey:
#2141 is merged, the azure failure is unrelated to core/3.100 - we can force-merge this once approved, and *then* do a CI-Update to fix/pin azure.

PR #2219, opened 2026-02-23
Fix monolithic push memory issues

gerrod3:
According to the spec there are two ways to perform monolithic blob pushes:
1. On the initial POST request with the digest query parameter
2. Empty (normal) body POST, then all in the PUT

There's an unofficial third way that podman can apparently do according to our comments which is: 
1. Normal empty POST
2. Entire chunk in one PATCH with no Range header
3. Empty PUT

The code should now call our special large chunk handler for all three cases, we were forgetting case number 2. Note that depending on the client they could still cause the server to go OOM if they send ridiculously large chunks in the normal chunked upload path since we read the entire chunk into memory before saving it. There is nothing we can do here as part of the code lives in pulpcore, but honestly what is that client thinking sending such large chunks!

### 📜 Checklist

### DIFF
--- a/CHANGES/+mono-push.bugfix
+++ b/CHANGES/+mono-push.bugfix
@@ -1,0 +1,1 @@
+Fixed memory usage when pushing large images with monolithic upload.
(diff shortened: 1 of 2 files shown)


gerrod3:
The most we can do is maybe introduce a setting that checks the size of the chunk and if over a certain amount return a 4XX error that the chunk is too large for a PATCH request, use a smaller size.

Issue #2080, opened 2025-10-07
Some models can hit maximum value of sequence (32 bit integer primary keys) errors during normal operation


dralley:
**Version**

Any

**Describe the bug**

Several models, such as `BlobManifest`, which do not inherit from `Content` and thus use the standard Django sequential 32bit integer primary key type, can potentially exhaust the address range of that 32 bit integer (actually 31 since negatives don't count) during normal Pulp operation

**To Reproduce**
See Jira

**Expected behavior**

**Additional context**
https://issues.redhat.com/browse/PULP-890

PR #2235, opened 2026-03-03
Migration to make two sequences BIGINT.

ggainey:
This addresses the failure noted in https://issues.redhat.com/browse/PULP-890 .

### DIFF
--- a/CHANGES/2080.bugfix
+++ b/CHANGES/2080.bugfix
@@ -1,0 +1,3 @@
+Altered several id-fields and their related sequences.
+
+INTEGER AutoField sequences can "run out" on large/old instances, update to BIGINT.
--- a/pulp_container/app/__init__.py
+++ b/pulp_container/app/__init__.py
@@ -1,4 +1,21 @@
 from pulpcore.plugin import PulpPluginAppConfig
+from django.db import connection
+from django.db.models.signals import post_migrate
+
+update_sequences_to_bigint = """
+ALTER TABLE container_blobmanifest ALTER COLUMN id TYPE bigint;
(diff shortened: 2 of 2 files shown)


gerrod3:
Is this related to this issue: https://github.com/pulp/pulp_container/issues/2080. I see they share the same JIRA.

ggainey:
> Is this related to this issue: #2080. I see they share the same JIRA.

It is - somehow didn't notice #2080 when looking for open issues :(

dralley:
@ggainey @gerrod3 going forwards, how do we ensure that non-BaseModel-derived models use the correct sequence type?

We can update these specific fields to BigAutoField, but we should probably also ensure that new such fields are set to BigAutoField by default via: 
https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-DEFAULT_AUTO_FIELD
https://docs.djangoproject.com/en/6.0/topics/db/models/#automatic-primary-key-fields

And/or migrate to UUID pks via BaseModel

pulp_container/app/migrations/0047_bigint_sequences.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 5.2.6 on 2025-09-23 11:54
+
+from django.db import migrations, models
+
+
+# This is an IDEMPOTENT change - it can be backported, because applying it
+# multiple times is harmless.
+update_sequences_to_bigint = """
+ALTER SEQUENCE container_blobmanifest_id_seq AS BIGINT;
+ALTER SEQUENCE container_manifestlistmanifest_id_seq AS BIGINT;
```
dralley: Does Django still recognize this as AutoField, or does it see it as BigAutoField?  I suppose even if the former is true it still works?

PR #2326, opened 2026-04-15
Add a fingerprint to ManifestSignature

dralley:
closes #2261

### 📜 Checklist


pulp_container/app/serializers.py
```diff
@@ -186,6 +186,7 @@ class ManifestSignatureSerializer(NoArtifactContentSerializer):
     digest = serializers.CharField(help_text="sha256 digest of the signature blob")
     type = serializers.CharField(help_text="Container signature type, e.g. 'atomic'")
     key_id = serializers.CharField(help_text="Signing key ID")
+    fingerprint = serializers.CharField(help_text="Signing key fingerprint")
```
dralley: We could use `PgpKeyFingerprintField` here, if you're comfortable with that 

https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/app/serializers/repository.py#L91-L96

It's a bit different though.  Prefixed with type, etc.

gerrod3: Would we need to bump the pulpcore lowerbound? Does it give us some benefits?

dralley: The primary benefit is consistency across pulp, and it would be possible to migrate to it exclusively long-term, because the prefix includes the possibility of `keyid:<keyid>` for keys that **only** have a keyid and don't have a fingerprint.  But yes you'd need to bump the minimum bound

I guess you could question whether we want to keep the key_id field long-term, and if the answer is yes than the `PpgKeyFingerprintField` provides less value, and if the answer is no then it provides more value.

gerrod3: Choose what you think is best, I don't have a strong opinion either way.

dralley:
I am tempted to leave it as-is.

pulp_container/app/migrations/0048_manifestsignature_fingerprint.py
```diff
@@ -0,0 +1,75 @@
...
+        if fingerprint:
+            assert sig.key_id == keyid_from_fingerprint(fingerprint), (
+                f"Signature {sig.pk}: key_id {sig.key_id} does not match "
+                f"fingerprint {fingerprint}"
+            )
+            sig.fingerprint = fingerprint
+            updated.append(sig)
+
+    if updated:
+        ManifestSignature.objects.bulk_update(updated, ["fingerprint"], batch_size=1000)
```
gerrod3: Do we need to clear the updated array throughout the loop or is the size of the objects negligible?

dralley: I think the signatures *should* be OK but I can make it defensive if you want.  I guess PQC signatures are potentially somewhat large in comparison to traditional ones.

gerrod3 (approved):

PR #2358, opened 2026-05-06
Remove content app redirect for manifest fetching

### LINKED ISSUE #1974: Omit redirects to content-app
Manifest data is now available on the content unit model as opposed to needing to separately redirect to the content app.  That means that in theory we should be able to eliminate this redirect entirely and serve the data directly from the registry.

Unfortunately there seems to be some issues with the caching machinery getting in the way of this.

See the discussions on previous PRs:

https://github.com/pulp/pulp_container/pull/1890#discussion_r1914231150
https://github.com/pulp/pulp_container/pull/1644

gerrod3:
Fixes https://github.com/pulp/pulp_container/issues/1974
https://redhat.atlassian.net/browse/PULP-486

To properly remove the redirect for manifest fetching I had to remove the PullThroughCaching redirect logic as well. Now the entirety of PullThroughCaching lives on the registry_api side, no more content app logic needed. These are the three scenarios we were handling for pull-through caching:

1. Manifest tag fetch:
Upon first fetch you need to create the tag and download the manifest and initiate a task to download the rest of the image in the background (`tag=None, local_manifest=None`). If the upstream manifest already exists on Pulp you simply need to create the tag and add it and the image's content to the repo (`tag=None, local_manifest is not None`). On the second fetch of a tag with pull-through you need to check to see if the tag has been updated on the remote. There are three scenarios for this: **one** - new upstream manifest that needs to be downloaded to Pulp (`tag is not None, local_manifest=None`), **two** - new upstream manifest already in Pulp needs to be added to repo (`tag is not None, local_manifest is not None`), and **three** - no changes on upstream (`tag is not None, tag.manifest.digest == local_manifest.digest`). If the tag doesn't exist on the remote, but we have the tag locally then we should serve our local copy (`tag is not None, fetch_manifest raised ApiException`). 
2. Manifest digest fetch:
When fetching a manifest by digest we either have it or we don't. If we don't, we check the upstream if it's there. If yes, then we check if the manifest is in Pulp locally and just needs to be added to the repo, or if we need to download the whole image in the background.
3. Blob fetch:
When fetching a blob and we don't have it in the repository we check to see if the remote has the blob. If so, we create the blob as an on-demand content and redirect to the content-app for it to finish streaming the blob from the upstream. We also add it to the `pending_blobs` to speed up future requests. 

### 📜 Checklist

### DIFF
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -24,6 +24,7 @@ from django.db.models import F, Func, TextField, Value
 from django.forms.models import model_to_dict
 from django.shortcuts import get_object_or_404
 from rest_framework.exceptions import (
+    APIException,
     AuthenticationFailed,
     NotAuthenticated,
     ParseError,
@@ -44,7 +45,7 @@ from rest_framework.viewsets import ViewSet
 from pulpcore.plugin import pulp_hashlib
 from pulpcore.plugin.exceptions import TimeoutException
 from pulpcore.plugin.files import PulpTemporaryUploadedFile
-from pulpcore.plugin.models import Artifact, ContentArtifact, UploadChunk
+from pulpcore.plugin.models import Artifact, ContentArtifact, RemoteArtifact, UploadChunk
 from pulpcore.plugin.tasking import dispatch
 from pulpcore.plugin.util import get_domain, get_objects_for_user, get_url
 
@@ -72,7 +73,7 @@ from pulp_container.app.redirects import (
     FileStorageRedirects,
     S3StorageRedirects,
 )
-from pulp_container.app.tasks import aadd_and_remove
+from pulp_container.app.tasks import aadd_and_remove, download_image_data
 from pulp_container.app.token_verification import (
     RegistryAuthentication,
     RegistryPermission,
@@ -80,9 +81,11 @@ from pulp_container.app.token_verification import (
     TokenPermission,
 )
 from pulp_container.app.utils import (
+    calculate_digest,
     determine_media_type,
     extract_data_from_signature,
     filter_resource,
+    get_accepted_media_types,
     get_full_path,
     has_task_completed,
     validate_manifest,
@@ -90,6 +93,7 @@ from pulp_container.app.utils import (
 from pulp_container.constants import (
     EMPTY_BLOB,
     MANIFEST_TYPE,
+    MEDIA_TYPE,
     MEGABYTE,
     SIGNATURE_API_EXTENSION_VERSION,
     SIGNATURE_HEADER,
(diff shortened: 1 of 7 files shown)


pulp_container/app/registry_api.py
```diff
@@ -1069,21 +1112,70 @@ def handle_safe_method(self, request, path, pk):
...
-        except models.Blob.DoesNotExist:
-            if pk == EMPTY_BLOB:
-                return redirects.redirect_to_content_app("blobs", pk)
-            repository = repository.cast()
-            try:
-                blob = repository.pending_blobs.get(digest=pk)
-                blob.touch()
-            except models.Blob.DoesNotExist:
-                raise BlobNotFound(digest=pk)
+        if pk == EMPTY_BLOB:
```
dralley, line 1117: can you explain (and maybe document) this a bit better?  What is EMPTY_BLOB and why does it redirect?

pulp_container/app/registry_api.py
```diff
@@ -1069,21 +1112,70 @@ def handle_safe_method(self, request, path, pk):
...
+                    remote = distribution.remote.cast()
+                    blob_url = self.fetch_blob(remote, pk)
+                    blob = self.create_on_demand_blob(pk, remote, blob_url)
+                    repository.pending_blobs.add(blob)
+                else:
+                    raise BlobNotFound(digest=pk)
+            blob.touch()
         return redirects.issue_blob_redirect(blob)
 
+    def create_on_demand_blob(self, digest, remote, blob_url):
```
dralley, line 1140: Missing doc comments - would be nice if all functions added here had them

pulp_container/app/registry_api.py
gerrod3: It's a layer that has nothing in it. Each layer(blob) is a gzip-compressed tar and the empty blob is a no-op layer used in images, so we handle it specially and have the empty compressed bytes ready to serve if requested. See the `_empty_blob()` method in `registry.py`, there are still bytes that need to be served so that is why it was originally redirected to the content app.

dralley: The name `pk` is a bit confusing because the EMPTY_BLOB constant seems to be a checksum string

pulp_container/app/registry_api.py
dralley: Also, even in the pull-through cache, you are using on-demand?

pulp_container/app/registry_api.py
```diff
@@ -1112,39 +1204,80 @@ def handle_safe_method(self, request, path, pk):
...
+                and distribution.pull_through_distribution
+                and permission_checker.has_pull_through_permissions(distribution)
+            ):
+                # check if the tag is on the upstream remote and if it has been updated
+                remote = distribution.remote.cast()
+                try:
+                    local_manifest, response = self.fetch_manifest(remote, pk)
+                except APIException as e:
+                    if not tag:
+                        raise e
```
dralley, line 1230: This could use more documentation probably.  I see the bit in the PR description, but it's difficult to trace, especially since fake_init_manifest doesn't have a docstring

pulp_container/app/registry_api.py
```diff
@@ -1212,16 +1327,32 @@ def get_content_units_to_add(self, manifest, tag):
             add_content_units.extend(manifest.blobs.values_list("pk", flat=True))
         return add_content_units
 
+    def fake_init_manifest(self, response, pk):
```
dralley, line 1349: Needs dscription

pulp_container/app/registry_api.py
gerrod3: Do you want me to rename `pk`? It's just a path parameter of the url and it's equal to the blob's `digest` field which is the unique field of the blob, but not its `pulp_id` (db pk).

pulp_container/app/registry_api.py
gerrod3: Yep, because we redirect to the content app we can use the content app's ability to stream the artifact from the remote while saving it to Pulp. So that is why we create the blob as on-demand before redirecting to the content app. Technically the content app also has the ability to create the content for pull-through at request time, but that logic is complicated and we already overwrite the content app for pulp-container and have special operations for `pending_blobs`, so I didn't think it was worth it to try to integrate into pulpcore's pull-through machinery.

pulp_container/app/registry_api.py
dralley: It's not needed necessarily, but it would be good to distinguish from `pk` in the database sense.  Not blocking

dralley (approved):

PR #2366, opened 2026-05-19
Switch push created repo to ContainerRepository

gerrod3:
https://redhat.atlassian.net/browse/PULP-1748
Assisted by: cursor composer 2

### 📜 Checklist


pulp_container/app/viewsets.py
```diff
@@ -186,33 +147,74 @@ def get_content_qs(self, qs, push_perm, mirror_perm):
...
+    """
+    domain = domain or get_domain()
+    namespaces = get_objects_for_user(
+        user,
+        ns_perm,
+        models.ContainerNamespace.objects.filter(pulp_domain=domain),
+    )
+    ns_repository_pks = models.ContainerDistribution.objects.filter(
+        namespace__in=namespaces,
+        pulp_domain=domain,
```
dralley, line 192: Completely unrelated to this PR, but it would be nice if we could make have queries automatically detect `domain` from a context var and filter based on that domain.  It's tedious and error prone to have to include the domain in every query.  Maybe there's an ORM middleware for that.

docs/admin/learn/rbac.md
```diff
@@ -28,464 +273,226 @@ Role based access control (RBAC) is configured using access policies for the fol
...
+    ```
+
+There are two action statements for `pull`, only one of them needs to be true for the user to be granted permission to pull an image. The order of statements is the order the statements are checked. So first Pulp checks if the distribution is public, i.e. `not is_private`. If the distribution is private than we check if the user has permission on the namespace or distribution. See the section below describing the custom conditions available for `pulp-container`.
 
 !!! note
+    If no statement evaluates to true then the request is denied. Admins always bypass any checks.
+
+#### `docker/podman push`
+
+When pushing an image to Pulp there are three different scenarios a user can find themselves in which determines the permissions that are checked for during the push.
```
dralley: This section is a bit too indirect and can be made more straightforwards.  I would also maybe reverse the order of the scenarios if you're going to talk about scenario 3 first.

docs/admin/learn/rbac.md
```diff
@@ -28,464 +273,226 @@ Role based access control (RBAC) is configured using access policies for the fol
...
+!!! note
+    Access polices can be reset to their default using the reset endpoint, e.g: pulp access-policy reset --href "$AP_HREF"
+
+### Docker API Permission Policies
+
+The main roles and permission policies for the Docker API described before are found under the `distributions/container/container` and `pulp_container/namespace` viewsets. Specifically the Container Distribution access policy contains special action statements that are used when determining the permissions a user has to perform a Docker action.
+
+#### `docker/podman pull`
+
+When pulling an image Pulp will check the `pull` action on the Container Distribution access policy.
```
dralley, line 317: Is it correct to say that all users must have a Pulp user in order to pull content?  I was trying to figure out what direction to go in with the pulp_rust plugin, since the normal protocol is that each user is assigned a secret API token which is then used for auth, but strictly speaking auth could probably go through the Pulp user permissions / RBAC system.

gerrod3: No, the `pull` action checks to see if the distribution is not private first, if it is "public" then any user, authenticated or not, can pull the image. If the distro is private then the user needs to be logged in as Pulp user.

docs/admin/learn/rbac.md
```diff
@@ -355,59 +355,49 @@ There are two action statements for `pull`, only one of them needs to be true fo
...
-When pushing an image to Pulp there are three different scenarios a user can find themselves in which determines the permissions that are checked for during the push.
+When pushing `namespace/image:tag`, Pulp checks whether the distribution and namespace already exist. The permission check depends on what is already in Pulp.
 
-- Scenario 1: Pushing a new image and a new namespace
-- Scenario 2: Pushing a new image inside an existing namespace
-- Scenario 3: Pushing a new tag/manifest for an existing image
+##### Push to a new namespace
 
-Let's start with Scenario 3 which requires the least amount of permissions and only checks the `push` action on the Container Distribution access policy.
+If the namespace does not exist, Pulp checks the `create` action on the Namespace access policy. The user needs `container.add_containernamespace`, or the namespace name must match their username. On success, they are granted the Namespace *Owner* role for the new namespace, which in turn grants them permission to create the distribution for the new image.
```
dralley, line 362: > The user needs `container.add_containernamespace`, **or** the namespace name must match their username.

Does this mean that users could create a namespace that matches a different username?

dralley (approved):

docs/admin/learn/rbac.md
gerrod3: Yes if they have the Namespace creator role or are admin.

PR #2369, opened 2026-05-29
Add OCI manifest to allowed media types of docker manifest list

gerrod3:
Assisted by: cursor-compose-2.5

### 📜 Checklist

### DIFF
--- a/pulp_container/tests/unit/test_json_schemas.py
+++ b/pulp_container/tests/unit/test_json_schemas.py
@@ -1,4 +1,5 @@
 import json
+from pathlib import Path
 
 from django.test import TestCase
 from jsonschema import Draft7Validator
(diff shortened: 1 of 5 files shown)


pulp_container/tests/unit/test_json_schemas.py
```diff
@@ -197,3 +198,33 @@ def test_valid_manifest_with_invalid_layer_media_type(self):
...
+        self.assertIn(MEDIA_TYPE.MANIFEST_OCI, entry_media_types)
+
+        try:
+            validate_manifest(
+                manifest,
+                MEDIA_TYPE.MANIFEST_LIST,
+                "sha256:31b31269d06603366cbfd0284708dcd2e281e8a4188e53fce3d3304439d0df3d",
+            )
+        except ManifestInvalid:
+            self.fail()
```
dralley, line 228: presumably just letting it throw the exception would fail the test, right?  why catch it then?  seems like this would just pass if something *else* went wrong and not this specific exception.

gerrod3: That's not how try/except works though. If another exception happen it wouldn't be caught and it would should up as an error in the pytest output. If ManifestInvalid is thrown it gets caught and shows as a failure in the output. The AI just followed the pattern in the other tests

dralley (approved):

PR #2381, opened 2026-06-02
Readd manifest artifact backwards compatibility

gerrod3:
### 📜 Checklist


pulp_container/app/tasks/sign.py
```diff
@@ -101,10 +102,23 @@ async def create_signature(manifest, reference, signing_service):
     """
     async with semaphore:
         # download and write file for object storage
-        async with tempfile.NamedTemporaryFile(dir=".", mode="wb", delete=False) as tf:
-            await tf.write(manifest.data.encode("utf-8"))
-            await tf.flush()
-        manifest_path = tf.name
+        if not manifest.data:
+            # TODO: BACKWARD COMPATIBILITY - remove after fully migrating to artifactless manifest
```
mdellweg, line 106: "Fully migrating" at this point should include making the manifest.data field not null, so the migration actually fails if it's not yet done.

mdellweg (dismissed):
So this adds the compatibility code back to 2.26 specifically.
Should we still make the sync autorepair in main?

gerrod3:
> So this adds the compatibility code back to 2.26 specifically. Should we still make the sync autorepair in main?

Yes, I think we maybe also want to autorepair on this PR. This just enables the old behavior, but at the end it still requires the user to run management command to fully transition to our new field. It might make sense to just run the data migration for each manifest missing the `.data` field during the sync, of course it won't catch all the manifests in the system.

dralley:
Do we want this in the "official" 2.26 branch?  Will it negatively impact any other consumers? 

Should we backport to any other branches for consistency?

pulp_container/app/tasks/sign.py
dralley: Yes, and this feels very insufficient https://github.com/pulp/pulp_container/blob/main/pulp_container/app/migrations/0039_manifest_data.py

pulp_container/app/management/commands/container-handle-image-data.py
```diff
@@ -68,6 +68,13 @@ def handle(self, *args, **options):
             self.style.SUCCESS("Successfully updated %d manifests." % manifests_updated_count)
         )
 
+        if self._broken_manifests:
+            self.stdout.write(
+                self.style.WARNING("Found %d broken manifests. PKs:" % len(self._broken_manifests))
+            )
+            for manifest in self._broken_manifests:
+                self.stdout.write(self.style.WARNING("  %s" % manifest.pk))
```
dralley: Would it be worth also printing the manifest digest and/or any other metadata about them (e.g. associated images)?

Maybe which repository they are in, if any, or whether it's an orphan?

pulp_container/app/tasks/sync_stages.py
```diff
@@ -101,6 +101,9 @@ async def _check_for_existing_manifest(self, download_tag):
                 content_data, raw_text_data, response = await self._download_manifest_data(
                     response.url
                 )
+            if manifest.data is None:
+                manifest.data = raw_text_data
+                await manifest.asave(update_fields=["data"])
```
dralley, line 106: I consider this a low risk compare to the existing issues, but in theory since we are mutating content, it could potentially deadlock between two parallel syncs.

Not a blocker, just a note.

gerrod3: Well hopefully after running the command, this piece won't be executed often. Also since manifests are per image and each image gets its own repo, it's unlikely to have parallel syncs of the same two images going at the same time.

dralley: I agree

pulp_container/app/tasks/sync_stages.py
```diff
@@ -533,12 +536,16 @@ async def create_listed_manifest(self, manifest_data):
...
             # TODO: BACKWARD COMPATIBILITY - remove after fully migrating to artifactless manifest
             elif saved_artifact := await manifest._artifacts.afirst():
-                content_data, _ = await sync_to_async(get_content_data)(saved_artifact)
+                content_data, raw_bytes_data = await sync_to_async(get_content_data)(saved_artifact)
+                manifest.data = raw_bytes_data.decode("utf-8")
+                await manifest.asave(update_fields=["data"])
             # if artifact is not available (due to reclaim space) we will download it again
             else:
-                content_data, manifest = await self._download_and_instantiate_manifest(
+                content_data, new_manifest = await self._download_and_instantiate_manifest(
```
dralley, line 544: So before, we were taking the returned `manifest` value directly (which AFAICT is an unsaved model object) and throwing away the previous saved `manifest`.  Presumably later on the `QueryExistingContents` stage would swap the model *back*, unless it's actually meaningfully different.  And so substantively it's only doing a download of `content_data` 

 And in the new version, we are taking an existing `manifest` from above, overwriting the `data` field, and then re-saving.  And so it's still mostly about the download, but instead of the manifest getting swapped with the original later we're just updating the original now and passing it along.

Right?  This is subtle, not sure if I have it correct.

gerrod3: Yeah that's right, we don't have the artifact so we need to download the data and update the manifest.

pulp_container/app/management/commands/container-handle-image-data.py
```diff
@@ -68,6 +70,23 @@ def handle(self, *args, **options):
...
+                repos = manifest.repositories.all()
+                if repos:
+                    for repo in repos:
+                        broken_by_repo[get_url(repo)].append(get_url(manifest))
+                else:
+                    broken_by_repo["orphaned"].append(get_url(manifest))
+            for repo_url, manifests in broken_by_repo.items():
+                self.stdout.write(self.style.WARNING("  %s" % repo_url))
+                for manifest_url in manifests:
+                    self.stdout.write(self.style.WARNING("    %s" % manifest_url))
```
dralley, line 88: We might still want to get a list of the unique affected manifest PKs directly.

Maybe it's easiest to just print that in a separate loop at the end.

dralley (dismissed):

dralley (approved):

PR #2385, opened 2026-06-04
Update migrations 39-45 to perform the manifest data migration for new fields [2.26]

gerrod3:
Still need to test, but after each migration that adds a new manifest field (after the last 2.19 migration of 0038), run a data migration to populate the field that container-handle-image-data would handle.

0039 - Adds `data`; populates `data`, `annotations` and `labels` (the last two were added in 0038, but not filled in)
0042 - Adds `type`; populates `type`, `is_bootable` and `is_flatpak` (the last two were also added in 0038)
0043 - Adds `os`, `architecture`, `compressed_image_size`; removed the handle-image-data warning
0045 - Fixes `compress_image_size`; populates `os`, `architecture`, `compressed_image_size`; doesn't make sense to populate image_size till we fix the field in this migration

I can also change this to do the entire migration in the last migration 0045, so that we aren't doing multiple passes over the Manifest table. I don't delete the backing artifacts of the manifests, so this should be pretty safe.  I added a check to see if the backing artifact is there and skip over manifests that didn't get `data` migrated in later migrations. There is still the possibility that it could fail if the config_blob artifact disappears, maybe I should also add safe guards around that, but not sure how common that is (there are no safeguards in our main model code that I copied over from...)

### 📜 Checklist


pulp_container/app/migrations/0039_manifest_data.py
```diff
@@ -1,15 +1,61 @@
...
+    'is_bootable' and 'is_flatpak' fields are not initialized since they require annotations and 
+    labels to be initialized first.
+    """
+    Manifest = apps.get_model("container", "Manifest")
+    to_update = []
+    manifests = Manifest.objects.filter(
+        contentartifact__artifact__isnull=False
+    ).distinct().prefetch_related("_artifacts").select_related("config_blob")
+    for manifest in manifests.iterator(chunk_size=1000):
+        manifest_artifact = manifest._artifacts.get()
```
dralley, line 32: While it sounds like the missing artifact thing may have been a rare issue, do you think we should try to either A) skip and print a warning or B) raise a proper error if an artifact (exactly 1) doesn't exist?

pulp_container/app/migrations/0039_manifest_data.py
```diff
@@ -1,15 +1,61 @@
...
+def init_annotations(manifest, manifest_data):
+    # annotations are part of OCI only
+    if manifest.media_type not in (MEDIA_TYPE.MANIFEST_OCI, MEDIA_TYPE.INDEX_OCI):
+        return False
+
+    manifest.annotations = manifest_data.get("annotations", {})
+
+def init_labels(manifest):
+    if manifest.config_blob:
+        config_artifact = manifest.config_blob._artifacts.get()
```
dralley: Same

pulp_container/app/migrations/0039_manifest_data.py
gerrod3: The `contentartifact__artifact__isnull=False` skips over manifests that don't have their backing artifact. However, the `config_blob` which is a Blob content unit could be on-demand or missing its artifact, so I think I need to add a check for that.

gerrod3:
Closing as not sure we want to modify the migrations for the older versions.

PR #2406, opened 2026-07-08
[PULP-1749] Add migrate endpoint for push repositories

gerrod3:
## Summary
- Add a `migrate` action on push repositories that dispatches a task to convert `ContainerPushRepository` into `ContainerRepository`
- Support optional `copy_versions` to preserve full repository version history (defaults to copying only the latest version)
- Reassociate distributions, return the created repository in the task result, and add functional test coverage

## Test plan
- [ ] Run `pytest pulp_container/tests/functional/api/test_migrate_push_repository.py`
- [ ] Verify `POST /pulp/api/v3/repositories/container/container-push/{pk}/migrate/` returns 202 and completes successfully
- [ ] Confirm migrated repository appears under container repositories with preserved content and distribution link
- [ ] Test `copy_versions: true` when multiple repository versions exist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pulp_container/app/tasks/migrate_push_repository.py
+++ b/pulp_container/app/tasks/migrate_push_repository.py
@@ -1,0 +1,73 @@
+from django.db import transaction
+
+from pulpcore.plugin.models import CreatedResource
+
+from pulp_container.app.models import (
+    ContainerDistribution,
+    ContainerPushRepository,
+    ContainerRepository,
+)
+from pulp_container.app.serializers import ContainerRepositorySerializer
+
+
+def migrate_push_repository(push_repository_pk, copy_versions=False):
+    """
+    Convert a ContainerPushRepository into a ContainerRepository.
+
+    Creates a new container repository with the same name and metadata, copies
+    content from the push repository, reassociates any distributions, and deletes
+    the original push repository.
+
+    Args:
+        push_repository_pk (str): The primary key for the push repository to migrate.
+        copy_versions (bool): If True, replay the full repository version history into the
+            new repository (new version numbers/timestamps). If False, only copy content
+            from the latest repository version.
+    """
+    push_repository = ContainerPushRepository.objects.get(pk=push_repository_pk)
+
+    original_name = push_repository.name
+    temp_name = f"{original_name}__migrating__{push_repository.pk}"
+
+    with transaction.atomic():
+        push_repository.name = temp_name
+        push_repository.save(update_fields=["name"])
+
+        container_repository = ContainerRepository(
+            name=original_name,
+            pulp_domain=push_repository.pulp_domain,
+            pulp_labels=push_repository.pulp_labels,
+            description=push_repository.description,
+            retain_repo_versions=push_repository.retain_repo_versions,
+            retain_checkpoints=push_repository.retain_checkpoints,
+            user_hidden=push_repository.user_hidden,
+            manifest_signing_service=push_repository.manifest_signing_service,
+        )
+        container_repository.save()
+
+        container_repository.pending_blobs.set(push_repository.pending_blobs.all())
+        container_repository.pending_manifests.set(push_repository.pending_manifests.all())
+
+        if copy_versions:
+            for push_version in push_repository.versions.complete().order_by("number"):
+                if push_version.number == 0 and not push_version.content.exists():
+                    continue
+                with container_repository.new_version() as new_version:
+                    new_version.set_content(push_version.content.all())
+        else:
+            latest_version = push_repository.latest_version()
+            if latest_version:
+                with container_repository.new_version() as new_version:
+                    new_version.set_content(latest_version.content.all())
+
+        ContainerDistribution.objects.filter(repository=push_repository).update(
+            repository=container_repository
+        )
+
+        push_repository.delete()
+
+    CreatedResource(content_object=container_repository).save()
+
+    return ContainerRepositorySerializer(
+        instance=container_repository, context={"request": None}
+    ).data
--- a/pulp_container/app/tasks/migrate_push_repository.py
+++ b/pulp_container/app/tasks/migrate_push_repository.py
@@ -1,70 +1,130 @@
-from django.db import transaction
+from django.contrib.contenttypes.models import ContentType
+from django.db import connection, transaction
 
 from pulpcore.plugin.models import CreatedResource
+from pulpcore.plugin.models.role import GroupRole, UserRole
 
-from pulp_container.app.models import (
-    ContainerDistribution,
-    ContainerPushRepository,
-    ContainerRepository,
-)
+from pulp_container.app.models import ContainerPushRepository, ContainerRepository
 from pulp_container.app.serializers import ContainerRepositorySerializer
 
 
-def migrate_push_repository(push_repository_pk, copy_versions=False):
+def migrate_push_repository(push_repository_pk):
     """
-    Convert a ContainerPushRepository into a ContainerRepository.
+    Convert a ContainerPushRepository into a ContainerRepository in place.
 
-    Creates a new container repository with the same name and metadata, copies
-    content from the push repository, reassociates any distributions, and deletes
-    the original push repository.
+    Swaps the multi-table-inheritance child row and updates `pulp_type` on the
+    parent repository so the primary key is preserved. Repository versions,
+    distributions, and other FKs to `core.Repository` remain valid.
 
     Args:
         push_repository_pk (str): The primary key for the push repository to migrate.
-        copy_versions (bool): If True, replay the full repository version history into the
-            new repository (new version numbers/timestamps). If False, only copy content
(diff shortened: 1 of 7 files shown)


mdellweg:
Do you plan to add a management command to migrate all remaining repositories?

Do you think we need to keep the old href's around as an alias? (Or by returning a clever redirect?)

gerrod3:
> Do you plan to add a management command to migrate all remaining repositories?

Yeah. I think I'll do it in a separate PR once I get this right.

> Do you think we need to keep the old href's around as an alias? (Or by returning a clever redirect?)

No, this is one downside of the migration is that it's a completely new object, old hrefs won't work. I think it won't be a big deal since these repos are auto-created on push and auto-deleted when the distro is removed, so I don't expect many users are keeping track of these hrefs. Adding a compatibility layer is a lot of effort so my plan is to make the docs explicit on how the migration works so users understand what is happening.

pulp_container/app/tasks/migrate_push_repository.py
```diff
@@ -0,0 +1,73 @@
...
+        copy_versions (bool): If True, replay the full repository version history into the
+            new repository (new version numbers/timestamps). If False, only copy content
+            from the latest repository version.
+    """
+    push_repository = ContainerPushRepository.objects.get(pk=push_repository_pk)
+
+    original_name = push_repository.name
+    temp_name = f"{original_name}__migrating__{push_repository.pk}"
+
+    with transaction.atomic():
```
mdellweg: Is there any structural difference between the ContainerRepository and the ContainerPushRepository table?
(I am wondering if we could do something clever in SQL to just change the DNA of the existing object.)

As far as I know the inheritance here is managed with a one-to-one relation so that the pk of the parent model is also the pk of the child model. And since the content models in those repositories as well as the repository versions are compatible (as far as I know), this could be a single create and delete in a transaction.

mdellweg: For example, we used this code in an old pulp_file migration:
```
def migrate_data_from_old_model_to_new_model_up(apps, schema_editor):
    """ Move objects from FileDistribution to NewFileDistribution."""
    FileDistribution = apps.get_model('file', 'FileDistribution')
    NewFileDistribution = apps.get_model('file', 'NewFileDistribution')
    for file_distribution in FileDistribution.objects.all():
        with transaction.atomic():
            NewFileDistribution(
                pulp_id=file_distribution.pulp_id,
                pulp_created=file_distribution.pulp_created,
                pulp_last_updated=file_distribution.pulp_last_updated,
                pulp_type=file_distribution.pulp_type,
                name=file_distribution.name,
                base_path=file_distribution.base_path,
                content_guard=file_distribution.content_guard,
                remote=file_distribution.remote,
                publication=file_distribution.publication
            ).save()
            file_distribution.delete()
```
Maybe the "trick" here is that the raw Django migration level objects don't "know" about the pulp specific MasterDetailModel and can carefully work around that.

mdellweg: e.g. to turn a file_repository into a pgp keyring (yes, we should not do that becuase they are not as compatible...) i needed to call these three commands:
```
insert into core_openpgpkeyring (repository_ptr_id) VALUES ('019fd0fc-103f-775a-bb16-f92440ba53dc');
update core_repository set pulp_type = 'core.openpgp' where pulp_id = '019fd0fc-103f-775a-bb16-f92440ba53dc';
delete from file_filerepository where repository_ptr_id = '019fd0fc-103f-775a-bb16-f92440ba53dc';
```

gerrod3: Ok, I've tested this out and yeah the ORM code in the migration behaves differently from normal ORM code and runs into issues with the multi-table inheritance. But we can still do the SQL migrate manually and we can preserve mostly everything. See my latest commit.

pulp_container/app/tasks/migrate_push_repository.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Args:
+        push_repository_pk (str): The primary key for the push repository to migrate.
+    """
+    with transaction.atomic():
+        push_repository = ContainerPushRepository.objects.select_for_update().get(
+            pk=push_repository_pk
+        )
+        if push_repository.pulp_type != ContainerPushRepository.get_pulp_type():
+            raise RuntimeError(
```
mdellweg, line 35: This is a sanity check, since we expect the transition down below to happen in a transaction.
Maybe with a comment we can prevent future readers to panic.

pulp_container/app/tasks/migrate_push_repository.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Swaps the multi-table-inheritance child row and updates ``pulp_type`` on the
+    parent repository so the primary key is preserved. Repository versions,
+    distributions, and other FKs to ``core.Repository`` remain valid.
+
+    Args:
+        push_repository_pk (str): The primary key for the push repository to migrate.
+    """
+    with transaction.atomic():
+        push_repository = ContainerPushRepository.objects.select_for_update().get(
```
mdellweg, line 30: I'm not sure what the `select_for_update` is supposed to protect here, but why not play it extra safe...
OTOH, does this "select for update" the base `Repository` row too?

pulp_container/app/tasks/migrate_push_repository.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+
+            # Deleting the push child cascades its pending M2M rows; parent repo is kept.
+            cursor.execute(
+                """
+                DELETE FROM container_containerpushrepository
+                WHERE repository_ptr_id = %s
+                """,
+                [push_repository_pk],
+            )
```
mdellweg: Is Django really setting up cascade delete in the database?
I guess this is just something to keep looking out for.

pulp_container/app/tasks/migrate_push_repository.py
```diff
@@ -0,0 +1,110 @@
...
+
+        UserRole.objects.filter(content_type=push_ct, object_id=str(push_repository_pk)).update(
+            content_type=container_ct
+        )
+        GroupRole.objects.filter(content_type=push_ct, object_id=str(push_repository_pk)).update(
+            content_type=container_ct
+        )
```
mdellweg, line 125: Good call!

pulp_container/app/tasks/migrate_push_repository.py
```diff
@@ -0,0 +1,110 @@
...
+            cursor.execute(
+                """
+                INSERT INTO container_containerrepository
+                    (repository_ptr_id, manifest_signing_service_id)
+                VALUES (%s, %s)
+                """,
+                [push_repository_pk, signing_service_id],
+            )
+
+            cursor.execute(
```
mdellweg, line 75: Hah, I guess it wouldn't be dramatic to loose the pending items accidentally, but yes, good call too.

Reminds me that "pending content" should probably be a pulpcore thing by now.

pulp_container/app/tasks/migrate_push_repository.py
```diff
@@ -0,0 +1,110 @@
...
+            raise RuntimeError(
+                f"Repository {push_repository_pk} has pulp_type "
+                f"{push_repository.pulp_type!r}, expected a push repository."
+            )
+        if ContainerRepository.objects.filter(pk=push_repository_pk).exists():
+            raise RuntimeError(
+                f"Container repository child row already exists for {push_repository_pk}."
+            )
+
+        signing_service_id = push_repository.manifest_signing_service_id
```
mdellweg, line 44: This is the one "normal" field we have to transfer from the child table, right?

pulp_container/app/tasks/migrate_push_repository.py
mdellweg: How about adding `cursor.execute("set transaction isolation level serializable")` as the very first statement in this transaction?

It could be the single one measure to protect all the reads happening here, including the GenericForeignKeys.

It may be a hit on performance, and could lead to `SerializationErrors` but for this kind of once in a lifetime migration operation it's probably a good tradeoff.

mdellweg (commented):
I very much like this approach!

pulp_container/app/tasks/migrate_push_repository.py
gerrod3: Correct.

pulp_container/app/tasks/migrate_push_repository.py
gerrod3: Apparently not, good call out, fixed.

pulp_container/app/tasks/migrate_push_repository.py
gerrod3: Yes, `select_for_update` locks both parents and childs row. Added the `SET` transaction level.

mdellweg (approved):

PR #2424, opened 2026-07-22
Fix 404 on tag fetch due to overly strict Accept header matching

### LINKED ISSUE #2417: Domain-scoped container registry: tag-based manifest fetch 404s, digest-based works (pulpcore 3.113.0 / pulp_container 2.28.0)
## Problem

With Domains enabled, every domain-scoped container registry pull **by tag** returns `404 MANIFEST_UNKNOWN`, while pulling the identical manifest **by digest** succeeds (`200`). Reproduced across all 7 on-demand container repos in our `default` domain.

## Expected outcome

`GET /v2/<domain>/<repo>/manifests/<tag>` should return the manifest, same as `GET /v2/<domain>/<repo>/manifests/<digest>` does.

## Versions (via `/pulp/api/v3/status/`)

- pulpcore: 3.113.0
- pulp_container: 2.28.0 (`domain_compatible: true`)
- Storage: S3 (RustFS, S3-compatible), `redirect_to_object_storage: False`

## Reproduction

```
BASE=https://pulp.example.com

TOKEN=$(curl -s -u "$USER:$PASS" \
  "$BASE/token/?service=pulp.example.com & scope=repository:default/quay/prometheus/prometheus:pull" \
  | python3 -c 'import sys,json;print(json.load(sys.stdin)["token"])')

# tags/list works
curl -s -H "Authorization: Bearer $TOKEN" \
  $BASE/v2/default/quay/prometheus/prometheus/tags/list
# {"name":"quay/prometheus/prometheus","tags":["latest"]}

# tag-based manifest fetch — 404
curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $TOKEN" \
  -H "Accept: application/vnd.docker.distribution.manifest.v2+json" \
  $BASE/v2/default/quay/prometheus/prometheus/manifests/latest
# 404

# digest-based manifest fetch (identical content, digest obtained from Pulp's own
# content/container/tags/ API for the same tag) — 200
curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $TOKEN" \
  -H "Accept: application/vnd.docker.distribution.manifest.v2+json" \
  $BASE/v2/default/quay/prometheus/prometheus/manifests/sha256:<digest>
# 200
```

- Pulp's own object API confirms the tag exists and points at exactly the manifest that 200s by digest: `GET /pulp/default/api/v3/content/container/tags/` returns `{"name": "latest", "tagged_manifest": "<href>"}`.
- The 404 is a **direct response from `pulp-api`** — no `Location` redirect header, Django access log shows `WARNING: Not Found: /v2/default/.../manifests/latest`.
- Ruled out: ingress/ nginx routing (we already have the `/pulp/container/` → content-app proxy rule some older-version reports needed), missing `Accept` header for manifest lists (sent all 4 media types, no change), token/auth (works fine, `tags/list` also works fine).
- All affected distributions have `repository_version: null` (follow-latest) and remote `policy: on_demand`.

## Suspected cause

2.28.0's changelog states manifest data is now served directly from the registry API instead of issuing a redirect to the content app (#1974). That's exactly the code path failing here. Reading `Manifests.handle_safe_method` in `pulp_container/app/registry_api.py` at the `2.28.0` tag, the tag-lookup and digest-lookup branches both filter against the same `repository_version.content` queryset resolved earlier in the view — digest lookup succeeds, tag lookup on the identical content does not. This points at the `Tag` resolution path specifically under Domains + on-demand + follow-latest-version distributions.

We haven't been able to confirm this is a regression introduced by #1974 specifically (attempted an isolated repro against a pre-2.28.0 image but hit an unrelated environment issue before getting a clean comparison) — flagging based on the changelog correlation and the tag/digest split, not a confirmed bisection.

Full writeup with more context: internal incident doc (happy to share more detail if useful) covers the same investigation in more depth.

Happy to provide more logs/detail if it'd help narrow this down.

gerrod3:
fixes: #2417
Generated by: cursor-composer-2.5

### 📜 Checklist

### DIFF
--- a/CHANGES/2417.bugfix
+++ b/CHANGES/2417.bugfix
@@ -1,0 +1,1 @@
+Fixed tag-based manifest fetches returning 404 when the client Accept header included q-values or when the manifest media type did not exactly match the parsed header value.
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -95,6 +95,7 @@ from pulp_container.app.utils import (
     get_accepted_media_types,
     get_full_path,
     has_task_completed,
+    is_media_type_accepted,
     validate_manifest,
 )
 from pulp_container.constants import (
@@ -166,11 +167,11 @@ class ManifestResponse(Response):
         manifest_media_type = manifest.media_type
         if request:
             accepted_media_types = get_accepted_media_types(request.headers)
-            if (
-                manifest_media_type not in accepted_media_types
-                and manifest_media_type != MEDIA_TYPE.MANIFEST_V1
+            if accepted_media_types and not is_media_type_accepted(
+                accepted_media_types, manifest_media_type
             ):
-                raise ManifestNotFound(reference=tag.name)
+                if manifest_media_type != MEDIA_TYPE.MANIFEST_V1:
+                    raise ManifestNotFound(reference=tag.name)
 
         # Schema v1 manifests are always returned with the signed content type
         if manifest_media_type == MEDIA_TYPE.MANIFEST_V1:
--- a/pulp_container/app/utils.py
+++ b/pulp_container/app/utils.py
@@ -42,6 +42,12 @@ def get_full_path(base_path, pulp_domain=None):
     return base_path
 
 
+def _parse_accept_entry(value):
+    """Return the media type from an Accept header entry, without parameters such as q-values."""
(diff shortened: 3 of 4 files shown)


gerrod3:
No longer needed.

PR #2438, opened 2026-07-26
Fix tag-based manifest pulls 404ing while digest-based pulls succeed

### LINKED ISSUE #2417: Domain-scoped container registry: tag-based manifest fetch 404s, digest-based works (pulpcore 3.113.0 / pulp_container 2.28.0)

wussh:
## Summary

- `ManifestResponse.from_tag` rejected requests based on the client `Accept` header, raising `ManifestNotFound` when the tag's manifest media type wasn't explicitly listed. `from_manifest` (the digest lookup path) never had this check, so the identical content served `200` by digest and `404` by tag — most visible for manifest-list/OCI-index tags whose media type clients often omit from `Accept`.
- Regression from #1974 (`8479d527`), which moved manifest serving from a content-app redirect (that ignored `Accept`) into the registry API view.
- `from_tag` now mirrors `from_manifest`: it always serves the tagged manifest, only rewriting schema v1 to the signed content type.

Fixes #2417

## Test plan

- [x] Reworked `test_pull_content.py::test_api_performes_schema_conversion` (previously asserted a 4xx with the comment "I don't understand what this is testing") into `test_api_serves_tag_regardless_of_accept_header`, which pulls the same manifest by tag and by digest with a narrow `Accept` header and asserts identical `200` responses.
- [x] **Reproduced end-to-end on the live domain-scoped, on-demand, follow-latest environment that originally reported #2417** (Domains enabled, S3-backed storage, pulpcore 3.113.0 / pulp_container 2.28.0). Pulled the exact live `registry_api.py` out of the running `pulp-api` pod, applied only the `from_tag` diff from this PR on top of it (no unrelated version drift), wrote it back into the running container, and reloaded gunicorn workers (no pod restart). Results:
  - Before patch: tag pull `404`, digest pull `200` (matches the issue exactly).
  - After patch: tag pull `200`, digest pull `200`, `tags/list` `200`, and the tag/digest response bodies are byte-identical.
  - Verified pod stayed healthy throughout, then reverted the pod back to the original unpatched image and re-confirmed the original `404` baseline — no lasting changes were left on the cluster.

### DIFF
--- a/CHANGES/2417.bugfix
+++ b/CHANGES/2417.bugfix
@@ -1,0 +1,1 @@
+Fixed tag-based manifest pulls returning 404 when the client's `Accept` header listed the tagged manifest's media type together with parameters such as q-values, since the media type comparison did not strip them and had no support for the `*/*` wildcard.
--- a/pulp_container/app/utils.py
+++ b/pulp_container/app/utils.py
@@ -42,6 +42,12 @@ def get_full_path(base_path, pulp_domain=None):
     return base_path
 
 
+def _parse_accept_entry(value):
+    """Return the media type from an Accept header entry, without parameters such as q-values."""
+    media_type = value.strip().split(";", maxsplit=1)[0].strip()
+    return media_type or None
+
+
 def get_accepted_media_types(headers):
     """
     Returns a list of media types from the Accept headers.
@@ -56,10 +62,20 @@ def get_accepted_media_types(headers):
     """
     accepted_media_types = []
     for header, values in headers.items():
-        if header == "Accept":
-            values = [v.strip() for v in values.split(",")]
-            accepted_media_types.extend(values)
+        if header.lower() == "accept":
+            for value in values.split(","):
+                if media_type := _parse_accept_entry(value):
+                    accepted_media_types.append(media_type)
     return accepted_media_types
+
+
+def is_media_type_accepted(accepted_media_types, media_type):
+    """
+    Return whether a manifest media type satisfies an Accept header.
+
+    Supports exact media type matches and the `*/*` wildcard used by some clients.
+    """
+    return any(accepted in ("*/*", media_type) for accepted in accepted_media_types)
 
 
 def urlpath_sanitize(*args):
--- a/CHANGES/+container-build-error-sanitization.bugfix
+++ b/CHANGES/+container-build-error-sanitization.bugfix
@@ -1,0 +1,1 @@
+Raised a proper `PulpException` subclass instead of a bare `Exception` when building or pushing an OCI image fails, so the error is not sanitized away by pulpcore in a future release.
--- a/pulp_container/app/exceptions.py
+++ b/pulp_container/app/exceptions.py
@@ -1,5 +1,7 @@
 from rest_framework.exceptions import APIException, NotFound, ParseError
 
+from pulpcore.plugin.exceptions import PulpException
+
 
 class BadGateway(APIException):
     status_code = 502
@@ -160,6 +162,19 @@ class ManifestSignatureInvalid(ParseError):
                 ]
             }
         )
+
+
+class ContainerBuildError(PulpException):
+    """Exception to signal that building or pushing an OCI image failed."""
+
+    error_code = "CON0001"
+
+    def __init__(self, message):
+        """Initialize the exception with the decoded stderr of the failed podman command."""
+        self.message = message
+
+    def __str__(self):
+        return self.message
 
 
 class InvalidRequest(ParseError):
--- a/pulp_container/app/registry_api.py
+++ b/pulp_container/app/registry_api.py
@@ -95,6 +95,7 @@ from pulp_container.app.utils import (
     get_accepted_media_types,
     get_full_path,
     has_task_completed,
+    is_media_type_accepted,
     validate_manifest,
 )
 from pulp_container.constants import (
@@ -166,11 +167,11 @@ class ManifestResponse(Response):
         manifest_media_type = manifest.media_type
         if request:
             accepted_media_types = get_accepted_media_types(request.headers)
-            if (
-                manifest_media_type not in accepted_media_types
-                and manifest_media_type != MEDIA_TYPE.MANIFEST_V1
+            if accepted_media_types and not is_media_type_accepted(
+                accepted_media_types, manifest_media_type
             ):
-                raise ManifestNotFound(reference=tag.name)
+                if manifest_media_type != MEDIA_TYPE.MANIFEST_V1:
+                    raise ManifestNotFound(reference=tag.name)
 
         # Schema v1 manifests are always returned with the signed content type
         if manifest_media_type == MEDIA_TYPE.MANIFEST_V1:
--- a/pulp_container/app/tasks/builder.py
+++ b/pulp_container/app/tasks/builder.py
@@ -13,6 +13,7 @@ from pulpcore.plugin.models import (
 )
 from pulpcore.plugin.util import get_domain
 
+from pulp_container.app.exceptions import ContainerBuildError
 from pulp_container.app.models import (
     Blob,
     BlobManifest,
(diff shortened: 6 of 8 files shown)


wussh:
Verified this fix end-to-end against the live cluster that originally hit #2417 (Domains + on-demand + follow-latest distribution, S3 storage, pulpcore 3.113.0 / pulp_container 2.28.0):

- **Before**: `GET /v2/default/quay/prometheus/prometheus/manifests/latest` → `404`; the same manifest by digest → `200`.
- **After** (applying only this PR's `from_tag` diff to the live `registry_api.py`, no unrelated changes, then reloading gunicorn workers): tag pull → `200`, digest pull → `200`, `tags/list` → `200`, and the tag/digest response bodies are byte-identical.
- Cluster was left in its original (pre-fix) state afterward — no permanent changes.

gerrod3:
@wussh Did you see my comment on the original issue? I'm not sure we should be removing the accept header check. In fact we should probably add it to the digest route for consistency. I went and fix some bugs with the accept header check #2424

lint_requirements.txt
```diff
@@ -8,5 +8,5 @@
 bump-my-version
 check-manifest
 packaging
-ruff
+ruff<0.16
```
jobselko: This should not be needed if you rebase on main

wussh:
Addressed the mechanical review feedback:
- Dropped the `ruff<0.16` pin — rebased on main, where the 0.16 lint findings were already fixed elsewhere, so it's no longer needed.
- `ContainerBuildError.error_code` renamed to `CON0001` per the naming convention you and @jobselko settled on.
- Changelog fragment now uses single backticks (renders as markdown).
- `cdomain_factory` now uses your version of the namespace cleanup from #2424 (suppress + `monitor_task`) instead of mine.

Still not addressed: your point about the accept-header check itself (removing it from `from_tag` vs. adding it to the digest route for consistency, and #2424 fixing the underlying matching bugs). Wanted to get the mechanical stuff out of the way first — let me know how you'd like to proceed on that part, happy to rework or close this in favor of #2424 if that's the better fix.

dralley:
Squash the commits please

CHANGES/2417.bugfix
```diff
@@ -0,0 +1 @@
+Fixed tag-based manifest pulls returning 404 when the client's ``Accept`` header did not list the tagged manifest's media type, while the identical digest-based pull succeeded.
```
gerrod3: Single tick marks.

pulp_container/app/utils.py
```diff
@@ -56,12 +62,25 @@ def get_accepted_media_types(headers):
...
+                if media_type := _parse_accept_entry(value):
+                    accepted_media_types.append(media_type)
     return accepted_media_types
 
 
+def is_media_type_accepted(accepted_media_types, media_type):
+    """
+    Return whether a manifest media type satisfies an Accept header.
+
+    Supports exact media type matches and the ``*/*`` wildcard used by some clients.
```
gerrod3: Single tick marks.

pulp_container/app/utils.py
```diff
@@ -56,12 +62,25 @@ def get_accepted_media_types(headers):
...
+    """
+    for accepted in accepted_media_types:
+        if accepted in ("*/*", media_type):
+            return True
+    return False
```
gerrod3: Looking at this I think we can do this:
```suggestion
    return accepted in ("*/*", *accepted_media_types)
```

gerrod3 (changes requested):
Looking pretty good. Few changes and we should be good to merge.

CHANGES/2417.bugfix
wussh: Fixed — single backticks now, both here and in the docstring.

pulp_container/app/utils.py
wussh: Fixed.

pulp_container/app/utils.py
wussh: Applied, with one tweak: your literal snippet references `accepted` outside the loop it was defined in, and checking `media_type` against a tuple of `("*/*", *accepted_media_types)` doesn't actually catch the wildcard case (it only matches if `media_type` itself equals `"*/*"`, which it never does). Collapsed the loop into `return any(accepted in ("*/*", media_type) for accepted in accepted_media_types)` instead — same semantics as before, just a one-liner. Pushed in 01f7832d.

wussh:
@gerrod3 addressed all three review comments in 01f7832d — single backticks in the changelog and docstring, and `is_media_type_accepted` collapsed to a one-liner (see inline reply for a small correction to the suggested snippet, which had an unbound variable and missed the wildcard case).

Also confirming the earlier `lint_requirements.txt` comment is moot now — that file no longer diverges from `main` after the rebase.

gerrod3:
@wussh Can you squash and repush? Hopefully github actions are working again, then I can approve and merge.

wussh:
Squashed into a single commit and repushed — ab61e13d. Also dropped some stray `CHANGES.md`/`pyproject.toml` noise that had crept in from an earlier rebase (unrelated bot commits, verified they weren't part of this fix's diff).

gerrod3 (approved):
Thanks for all the work on this!

PR #2439, opened 2026-07-27
Raise TaskResourceNotFound instead of bare DoesNotExist in tasks

wussh:
## Summary

- pulpcore's task executor treats any task exception that isn't a `PulpException` subclass as deprecated: it logs a `pulpcore.deprecation` warning today and will sanitize the message away entirely in pulpcore 3.130.
- Every task that looks up its arguments' referenced objects by pk (repository, remote, manifest, signing service, artifact, ...) raises a bare Django `DoesNotExist` when that object was deleted between dispatch and execution. This is non-deterministic — a different one surfaces on each CI run depending on test/teardown timing (observed both `ManifestSigningService` and `ContainerDistribution` `DoesNotExist` warnings on unrelated PRs' CI, e.g. #2438).
- Adds `TaskResourceNotFound(PulpException)` and a `get_task_resource_or_raise()` helper (`app/utils.py`), used at the top-of-task object lookups in `tag.py`, `untag.py`, `recursive_add.py`, `recursive_remove.py`, `synchronize.py`, `download_image_data.py`, `sign.py`, and `builder.py`.
- Left the intentional `try/except DoesNotExist` in `builder.py`'s `get_or_create_blob` alone — that's expected control flow (create-on-miss), not an unhandled lookup failure.
- Also pulled in two related, independently-discovered `deprecations`-noise fixes so this PR is self-contained (same root cause class — non-`PulpException` raised in a task — found while chasing #2417's CI): `builder.py`'s podman build/push failure now raises `ContainerBuildError(PulpException)` instead of bare `Exception`, and `test_domains.py`'s `cdomain_factory` fixture now cleans up `ContainerNamespace` (a `PROTECT` FK to `Domain`) before teardown deletes the domain, instead of leaving it to raise `ProtectedError`.

**Not addressed** (out of scope for this PR): pulpcore's own generic tasks (e.g. `general_multi_delete`, which can raise `ContainerDistribution.DoesNotExist`) are outside pulp_container's control, so the `deprecations` CI job may still show occasional pulpcore-origin warnings even after this merges. This is a noise-reduction / forward-compat cleanup for pulp_container's own task code, not a guarantee of a fully green `deprecations` job.

## Test plan

- [x] `ruff check` / `ruff format --check` clean on all touched files.
- [x] Verified by grep that no other unhandled `Model.objects.get(pk=...)` task-entry lookups were missed in `app/tasks/`.
- [x] CI (lint/build/test/docs/sanity/ready-to-ship) — all green; `deprecations` job re-checked, remaining warnings confirmed unrelated to this PR's changes.

### DIFF
--- a/pulp_container/app/exceptions.py
+++ b/pulp_container/app/exceptions.py
@@ -1,5 +1,7 @@
 from rest_framework.exceptions import APIException, NotFound, ParseError
 
+from pulpcore.plugin.exceptions import PulpException
+
 
 class BadGateway(APIException):
     status_code = 502
@@ -160,6 +162,26 @@ class ManifestSignatureInvalid(ParseError):
                 ]
             }
         )
+
+
+class TaskResourceNotFound(PulpException):
+    """Exception to signal that a resource a task depends on no longer exists.
+
+    Tasks look up their arguments' referenced objects by pk. If that object was
+    deleted between dispatch and execution (e.g. by a racing delete), a bare Django
+    DoesNotExist is not a PulpException, which pulpcore's task executor logs as
+    deprecated and will sanitize away in a future release. Raise this instead so the
+    real reason is preserved on the task result.
+    """
+
+    error_code = "CON0001"
+
+    def __init__(self, message):
+        """Initialize the exception with a description of the missing resource."""
+        self.message = message
+
+    def __str__(self):
+        return self.message
 
 
 class InvalidRequest(ParseError):
--- a/CHANGES/+task-resource-not-found.bugfix
+++ b/CHANGES/+task-resource-not-found.bugfix
@@ -1,0 +1,1 @@
+Raised a proper `PulpException` subclass instead of a bare Django `DoesNotExist` when a task's referenced object (repository, remote, manifest, signing service, etc.) no longer exists, so the error is not sanitized away by pulpcore in a future release.
--- a/pulp_container/app/tasks/builder.py
+++ b/pulp_container/app/tasks/builder.py
@@ -13,6 +13,7 @@ from pulpcore.plugin.models import (
 )
 from pulpcore.plugin.util import get_domain
 
+from pulp_container.app.exceptions import TaskResourceNotFound
 from pulp_container.app.models import (
     Blob,
     BlobManifest,
@@ -138,9 +139,21 @@ def build_image(
         raise RuntimeError("Neither a name nor temporary file for the Containerfile was specified.")
 
     if containerfile_tempfile_pk:
-        containerfile_artifact = PulpTemporaryFile.objects.get(pk=containerfile_tempfile_pk)
+        try:
+            containerfile_artifact = PulpTemporaryFile.objects.get(pk=containerfile_tempfile_pk)
+        except PulpTemporaryFile.DoesNotExist:
+            raise TaskResourceNotFound(
+                f"PulpTemporaryFile matching pk={containerfile_tempfile_pk} does not exist. "
+                "It may have been deleted after this task was dispatched."
+            ) from None
 
-    repository = ContainerRepository.objects.get(pk=repository_pk)
+    try:
+        repository = ContainerRepository.objects.get(pk=repository_pk)
+    except ContainerRepository.DoesNotExist:
+        raise TaskResourceNotFound(
+            f"ContainerRepository matching pk={repository_pk} does not exist. It may "
+            "have been deleted after this task was dispatched."
+        ) from None
     name = str(uuid4())
     with tempfile.TemporaryDirectory(dir=".") as working_directory:
         working_directory = os.path.abspath(working_directory)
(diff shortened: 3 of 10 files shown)


CHANGES/+container-build-error-sanitization.bugfix
```diff
@@ -0,0 +1 @@
+Raised a proper ``PulpException`` subclass instead of a bare ``Exception`` when building or pushing an OCI image fails, so the error is not sanitized away by pulpcore in a future release.
```
gerrod3: The changelogs are rendered in markdown so use single ticks.

pulp_container/tests/functional/api/test_domains.py
```diff
@@ -22,6 +22,14 @@ def _domain_factory(*args, **kwargs):
     yield _domain_factory
 
     for domain in domains:
+        # ContainerNamespace has a PROTECT FK to Domain, so it must be cleared before
```
gerrod3: I would copy the fix I am adding here: https://github.com/pulp/pulp_container/pull/2424/changes#diff-5287a8b917dfa3076a52a1b5bdfc8ce254b5d74ecbddf6ad08179904c5e91649R30

pulp_container/app/exceptions.py
```diff
@@ -162,6 +164,39 @@ def __init__(self, digest):
...
+        self.message = message
+
+    def __str__(self):
+        return self.message
+
+
+class ContainerBuildError(PulpException):
+    """Exception to signal that building or pushing an OCI image failed."""
+
+    error_code = "PLP_CONTAINER_BUILD_0001"
```
gerrod3: @jobselko What did we decide for the nomenclature of error codes?

gerrod3 (changes requested):
Not sold on the retrieve or raise logic/exception. The other builder exception looks good.

pulp_container/app/exceptions.py
jobselko: 3 uppercase letters/_ , followed by 4 leading-zero digits, so `CON0001` for pulp container

jobselko:
@wussh I see this PR duplicates some code from #2438. It would be helpful to put those changes in a separate commit and clearly mark them to make the PR easier to review.

wussh:
Addressed:
- `error_code`s renamed to the `CON00XX` convention: `TaskResourceNotFound` -> `CON0001`, `ContainerBuildError` -> `CON0002`.
- Changelog fragments now use single backticks (markdown).
- `cdomain_factory` now uses @gerrod3's version of the namespace cleanup from #2424 (suppress + `monitor_task`) instead of mine.

@jobselko the duplicated `ContainerBuildError`/`cdomain_factory` changes from #2438 are already in their own separate commits (`Raise ContainerBuildError instead of bare Exception in image builder`, `Clean up ContainerNamespace before domain deletion in domain tests`) — let me know if you'd like them called out any more clearly than that.

@gerrod3 re: "not sold on the retrieve or raise logic/exception" — could you elaborate on what you'd prefer instead? Happy to rework `get_task_resource_or_raise` once I understand the concern.

wussh:
@gerrod3 @jobselko rebased this branch directly onto \`main\` and re-did the fix as a single commit (517637ef), which resolves both open items:

- **Duplicate code / diff noise**: the branch no longer stacks on top of #2438, so the \`ContainerBuildError\` commit that was showing up here is gone — this PR's diff is now just the \`TaskResourceNotFound\` work.
- **"Not sold on the retrieve or raise logic/exception"**: dropped \`get_task_resource_or_raise\` entirely. Every call site now does an explicit inline \`try/except ... raise TaskResourceNotFound(...) from None\`, mirroring the \`ContainerBuildError\` pattern you approved on #2438 — no generic wrapper.
- Error code stays \`CON0001\` (3 letters + 4 digits, per the convention confirmed above) since \`ContainerBuildError\` lives exclusively on #2438 now and there's no collision on this branch.
- The \`test_domains.py\` namespace-cleanup fix (mirroring your #2424 diff) landed on #2438 instead, since that's where the file is actually touched — this PR's scope no longer includes it.
- Old changelog-formatting comment on \`+container-build-error-sanitization.bugfix\` is moot; that fragment doesn't exist on this branch anymore (moved to #2438). The new fragment (\`+task-resource-not-found.bugfix\`) already uses single backticks.

Force-pushed the rebased branch just now — CI is green apart from the known flaky azure job and an infra-only podman config error on lowerbounds, both unrelated to this diff.

gerrod3:
Is this PR still needed? How often do you actually experience resources in tasks disappearing before they are used? This shouldn't happen with task resource locking.

wussh:
@gerrod3 Yeah, still needed — but the motivation isn't "resources disappearing in production despite locking," it's pulpcore's task-exception deprecation:

- pulpcore's task executor already logs a `pulpcore.deprecation` warning for any task exception that isn't a `PulpException` subclass, and starting 3.130 it'll sanitize the message away entirely. A bare `DoesNotExist` is exactly that case.
- The `DoesNotExist`s I'm actually seeing aren't from a real locking gap — they're CI/test-teardown timing (object deleted between dispatch and execution in test fixtures), which is why the same lookup surfaces `ManifestSigningService.DoesNotExist` on one run and `ContainerDistribution.DoesNotExist` on another (see #2438's CI history).
- So this PR isn't "fix a race" — it's "don't let task code raise a bare, soon-to-be-sanitized exception," converting it into a proper `PulpException` subclass (`TaskResourceNotFound`) with a useful message, consistent with the `ContainerBuildError` pattern already merged on #2438.

If you'd rather not carry this until it's actually causing pain, happy to close and revisit if/when the deprecation warning turns into an error in 3.130 — just let me know which you'd prefer.

PR #2484, opened 2026-08-31
Add PQC (ML-DSA) coverage to manifest signing tests

### LINKED ISSUE #2280: Test a container plugin signing service with PQC
The tooling at the moment is still a bit raw and experimental, but eventually we should have a test that uses a PQC key to sign manifests

### LINKED ISSUE #2281: Test pushing PQC signed content to pulp_container

dralley:
Parameterize the existing manifest signing and push-signature tests over both a traditional GPG key and a post-quantum ML-DSA/Sequoia key, covering signing with either a primary key or a dedicated signing subkey. Replaces python-gnupg verification with pysequoia.

closes #2280
closes #2281

Assisted-By: Claude Opus 4.8


dralley:
I can squash the commits once done, I wanted to retain the history for now.

dralley:
The tests have decided to break all of a sudden, but it's not specific to this PR, it's the nightly job too.

https://github.com/pulp/pulp_container/actions/runs/33714487446

Maybe something to do with the new pulpcore release.

pulp_container/tests/functional/api/test_sign_manifests.py
```diff
@@ -1,50 +1,118 @@
...
+    Runs against both primary-key and subkey signing.
+    """
+    service, fingerprint, keyid = manifest_signing_service
+    sign_data = {"manifest_signing_service": service.pulp_href}
 
     response = container_repository_api.sign(distribution.repository, sign_data)
     created_resources = monitor_task(response.task).created_resources
 
-    tags = container_tag_api.list(repository_version=created_resources[0])
+    repository_version = created_resources[0]
```
gerrod3, line 108: What is the output of the sign task? Is it the repository version? I would like to start moving towards using this field over `created_resource`.

pulp_container/tests/functional/conftest.py
```diff
@@ -37,20 +128,48 @@ class RegistryClient:
...
-        self.inspect = lambda *args: self._dispatch_command("inspect", *args)
-        self.import_ = lambda *args: self._dispatch_command("import", *args)
-        self.images = lambda *args: self._dispatch_command("images", "--format", "json", *args)
-        self.rmi = lambda *args: self._dispatch_command("rmi", *args)
-        self.tag = lambda *args: self._dispatch_command("tag", *args)
+    def logout(self, *args):
+        return self._dispatch_command("logout", *args)
+
+    @contextmanager
+    def authenticated(self, registry, username, password):
```
gerrod3, line 149: Nifty!

pulp_container/tests/functional/conftest.py
```diff
@@ -97,8 +216,12 @@ def tls_verify(bindings_cfg):
 
 
 @pytest.fixture(scope="session")
-def registry_client(tls_verify):
+def registry_client(tls_verify, tmp_path_factory, worker_id):
     """Fixture for a container registry client."""
+    # tmp_path_factory's counter is process-local, so include the xdist worker
+    # ID to prevent workers from selecting the same auth file.
+    auth_dir = tmp_path_factory.mktemp(f"podman-auth-{worker_id}")
+    os.environ["REGISTRY_AUTH_FILE"] = str(auth_dir / "auth.json")
```
gerrod3, line 224: Interesting. I assume we want to backport this change to our other branches? Have we seen our parallel tests failing because the podman auth was being swapped out underneath?

gerrod3 (commented):
Some questions before final approval. Looks good.

pulp_container/tests/functional/conftest.py
dralley: If you want.  At various points I did run into tests that were failing sporadically and it seems plausible that this could be the cause.  I had no other particularly good explanation for why they were flaky.

Of course, the refactor is completely separate, that's just something that I thought could use some improvement while investigating.  I can separate the two if you want.

I do want to backport the overall PR to the latest release but I wasn't really planning to go further, unless you want this specific patch backported further.

pulp_container/tests/functional/api/test_sign_manifests.py
dralley: Since a `ManifestSignature` is a model, yeah, it creates new content in the repo and therefore a new version 

What do you mean by "this field"?

gerrod3: `task.result` https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/task.py#L145. I keep calling it output, but it's named result. We should make sure every task produces a result. The output of the signing-task is `None` since it doesn't return anything: https://github.com/pulp/pulp_container/blob/main/pulp_container/app/tasks/sign.py#L81-L84.

pulp_container/tests/functional/conftest.py
gerrod3: I think I do want the specific patch backported. You can hold off on it till I finish fixing the older branches with the PT fix.

gerrod3:
@dralley Add changelogs for the issues to get the check-commits to pass. You can make them `.misc` since they aren't user facing.

pulp_container/tests/functional/api/test_sign_manifests.py
dralley: So, do you want the signing task changed in this PR, or file an issue for it, or just a followup PR, etc?

pulp_container/tests/functional/conftest.py
dralley: Ok, so the first commit should go to the previous release and main, and the 2nd commit to all branches?

Should I split it off to a separate PR then?  That would probably simplify things.

dralley:
@gerrod3 Changelogs even though none of the actual plugin code is touched?

pulp_container/tests/functional/api/test_sign_manifests.py
gerrod3: Let's just track it in an issue and make a followup.

pulp_container/tests/functional/conftest.py
gerrod3: Yes.

Up to you, whatever is easier.

gerrod3:
If you don't want to then don't, it's not a big deal. We'll just leave our check-commits sad :disappointed:

gerrod3 (approved):

gerrod3 (approved):

pulp_container/tests/functional/api/test_sign_manifests.py
dralley: Filed https://github.com/pulp/pulp_container/issues/2519

PR #2496, opened 2026-09-10
Fix pull-through distribution base paths

### LINKED ISSUE #2494: pulp_container SHOULD not disregard the base_path rules
**Version**
pulp_container: all of them
pulpcore: < 3.117

**Describe the bug**

**To Reproduce**
On pulpcore < 3.117 run:
```bash
pulp file repository create --name "test" --autopublish
pulp file distribution create --name breakme --repository test --base-path break/me
pulp file content upload --repository test --file LICENSE --relative-path LICENSE
curl http://localhost:8080/pulp/content/
curl http://localhost:8080/pulp/content/break/
curl http://localhost:8080/pulp/content/break/me/
curl http://localhost:8080/pulp/content/break/me/LICENSE

# --------

pulp container distribution create --name offender --base-path break
curl http://localhost:8080/pulp/content/break/me/LICENSE
```

gerrod3:
## What this does

- Restores pulpcore base-path overlap validation for container distributions.
- Stores UUID base paths for new pull-through distributions and marks them with a reserved pulp label.
- Looks up marked pull-through distributions by name, then falls back to the legacy base-path lookup.
- Adds a repair command for existing pull-through distributions and reports entries that need manual repair or still overlap another distribution.
- Adds Django warning pulp_container.W001 until operators run the repair command.

## Repair command

    pulpcore-manager container-repair-pull-through-distributions

The command repairs distributions where name matches base_path. It leaves mismatched distributions untouched and reports them for manual handling.

## Tests

- Full unit suite: 43 passed, 3 subtests passed.
- API persistence test for the UUID base path and marker label passed.
- Ruff and formatting checks passed.
- Django system checks passed.

Closes #2494


CHANGES/2494.bugfix
```diff
@@ -0,0 +1 @@
+Prevent container distributions from overlapping other distribution base paths. Pull-through distributions **NEED** to be fixed before upgrading pulpcore; run `pulpcore-manager container-repair-pull-through-distributions` to repair existing distributions.
```
mdellweg, line 1: Given that the constraint is a trigger only looking for new base_paths being inserted, you might get away with a few 500 if you run the command after updating...

pulp_container/app/serializers.py
```diff
@@ -499,17 +519,9 @@ def validate(self, data):
...
-            )
-
-        if not all(re.match(VALID_BASE_PATH_REGEX_COMPILED, p) for p in value.split("/")):
-            raise serializers.ValidationError(
-                _("The provided base path contains forbidden characters.")
-            )
-
-        return value
+        validate_registry_path(value)
+        # Ensure we follow pulpcore's cross-distribution overlap validation in addition to the OCI rules.
```
mdellweg, line 523: :+1:

pulp_container/app/serializers.py
```diff
@@ -556,18 +568,53 @@ class ContainerPullThroughDistributionSerializer(DistributionSerializer):
...
+        return validate_registry_path(value)
 
     def validate(self, data):
         validated_data = super().validate(data)
 
+        # This is the temporary fix for pull-through distros violating core's overlapping base-path rule
+        # The user must supply base-path and name as the same value, then after we validate they are good
+        # we set base-path to an UUID and use the name to construct the prefix of new distros' base-paths
+        # avoiding the overlapping problem; base-paths can share prefixes, but one must not be a subset of
+        # another. A proper fix should have us move off of core Distributions all together.
```
mdellweg, line 586: This get's me thinking:
Would it help to allow the core.base_path to be Null?
Or should we add a (hidden to the user) name_scope field so container appearing on "v2" could enforce the overlap rules completely independent of all the plugins appearing on "pulp/content"?

But then you said in the oci-registry there is no such restriction at all, right? We just need to make sure container keeps clear of the "pulp/content" paths. (That could be `name_scope=null`.)

How does pulp_container scope by Domains again?

mdellweg (commented):
Thank you for looking into this!

pulp_container/app/serializers.py
gerrod3: As it is currently, domains become the first part of the name when enabled. Normal upload/sync of nginx becomes `/v2/nginx`, with domains on it becomes `/v2/{domain_name}/nginx`, even within the default domain. 

Right now we use the content app solely for the final redirect to download the blob artifacts. We use the base-path uniqueness rules to enforce each repository/image has a unique name, but other than that we don't care about base-path. In fact using pulp-container with other plugins means we have fewer available names we can use for a image which is probably not what we want. Allowing base-path to be nullable could help, but I think we should just add a new field like `name_scope` that is required, and then derequire base-path and set it to a UUID, if we want to stick to using Distributions. If users see base-path can be null they will want to try to use it for other plugins and probably break those plugins.

CHANGES/2494.bugfix
gerrod3: Yeah, but this is meant to also be backported to older versions so I didn't want to include a forced migration.

mdellweg: I was trying to say the actual situation is even better than the changelog states (and that is OK, so this is not a request to change.).
Even after upgrading, it's still better late than never to run the repair.

gerrod3: Resolving: this clarifies that the repair remains useful after upgrading and explicitly requests no code change.

pulp_container/app/serializers.py
gerrod3: Resolving this acknowledgment; no action is requested.

pulp_container/app/serializers.py
gerrod3: Resolving: domain scoping was answered here, and the broader name-scope design is outside this repair PR.

pulp_container/tests/unit/test_serializers.py
```diff
@@ -103,3 +144,137 @@ def test_invalid_tag(self):
         self.assertFalse(serializer.is_valid())
         self.assertIn("tag", serializer.errors)
         self.assertIn("tag is not valid", str(serializer.errors["tag"][0]))
+
+
+class TestContainerPullThroughDistributionSerializer(TestCase):
```
mdellweg, line 149: We can write unittests in pytest fashion too. (Specifically with pytest-django...)

mdellweg (approved):

